PR #6, opened 2020-01-30
Generate code

### LINKED ISSUE #2: Use official SDK to build Objects
The official SDK contains JSON file describing every method/request/response

ie. https://github.com/aws/aws-sdk-php/blob/efdbdde0968355917470a255c0d2e509c49f7476/src/data/sqs/2012-11-05/api-2.json#L214-L229

To provide a library compatible with official Documentation, I suggest to use those files to generates our DTO/Objects.

The content of `simple-aws` is:
- low level logic to perform request, deal with response, authentication, pagination
- a developper tool to generate the Request/Response Functions based on SDK JSON files (similar to https://github.com/thecodingmachine/safe)

Optionnaly it will also provides:
- helper classes that provides a non-official but more friendly contract (`$s3->putFile($path)`) (ie like psr/cache vs psr/simple-cache)

What do you think?

Nyholm:
This is related to #2 

In the `build` folder we may run a command like: 
```
./build generate s3 PutObject
```

This will create a `S3Client::putObject()` and a class called `PutObjectOutput`. 

The output file is not 100% populated with the correct data. But properties, and getters are generated.


build/src/Generator/ClassFactory.php
```diff
@@ -0,0 +1,175 @@
...
+use Nette\PhpGenerator\PhpNamespace;
+use Nette\PhpGenerator\Property;
+use Nette\SmartObject;
+
+/**
+ * Generate Nette class from existing source.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class ClassFactory
```
Nyholm: This class is very similar to `Nette\PhpGenerator\Factory` with just a few changes. 

I wanted to include use statements and method bodies.

PR #13, opened 2020-01-31
Make sure we can configure code generator

Nyholm:


build/manifest.json
```diff
@@ -1,20 +1,22 @@
...
             "methods": {
                 "PutObject": {
-                    "generated": "2020-01-30T19:55:12+00:00"
+                    "generated": "2020-01-30T19:55:12+00:00",
+                    "generate-method": true
                 },
                 "CreateBucket": {
-                    "generated": "2020-01-30T19:55:15+00:00"
+                    "generated": "2020-01-31T09:32:53+00:00",
+                    "generate-method": true
```
Nyholm: S3 might be special. So don generate any methods.

PR #19, opened 2020-01-31
Adding "regenerate" command

Nyholm:


Nyholm:
This will also support optional return parameters

PR #20, opened 2020-01-31
Add Signer

jderusse:


src/S3/S3Client.php
```diff
@@ -21,6 +21,11 @@ protected function getServiceCode(): string
         return 's3';
     }
 
+    protected function getSignatureVersion(): string
+    {
+        return 's3';
```
Nyholm: v3?

jderusse: .. sadly.. no (This data is extracted from `API.json` file)

Nyholm: oh.

jderusse: For your Information, Every endpoints use V4, exept:
- `s3` thatis v4 + a custom header 
- `importexport` that use v2 (not implemented)

PR #22, opened 2020-02-01
Fix endpoint URL and Output wrapped

jderusse:
This PR fixes 2 things:

* [x] Endpoint URL. the API.Json files contains a "http.requestUri" field that define the pattern of the endpoint (ie. `/{Bucket}/{Key+}`
* [x] Wrapped output. Example SQS.GetQueueUrl

### DIFF
--- a/build/Generator/ApiGenerator.php
+++ b/build/Generator/ApiGenerator.php
@@ -66,7 +66,7 @@ class ApiGenerator
     /**
      * Generate classes for the output. Ie, the result of the API call.
      */
-    public function generateResultClass(array $shapes, $service, $baseNamespace, $className, $root = false)
+    public function generateResultClass(array $shapes, $service, $baseNamespace, $className, $wrapper = null, $root = false)
     {
         $namespace = new PhpNamespace($baseNamespace);
         $class = $namespace->addClass($className);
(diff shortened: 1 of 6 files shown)


build/Generator/ApiGenerator.php
```diff
@@ -81,8 +81,8 @@ public function generateResultClass(array $shapes, $service, $baseNamespace, $cl
             // Add trait only if file does not exists
             $traitFilename = \sprintf('%s/%s/Result/%s.php', $this->srcDirectory, $service, $traitName);
             if (!\file_exists($traitFilename)) {
-                $this->createOutputTrait($baseNamespace, $traitName, $members, $traitFilename);
             }
+                $this->createOutputTrait($baseNamespace, $traitName, $members, $wrapper, $traitFilename);
```
Nyholm: This is debug, right?

src/Core/AbstractApi.php
```diff
@@ -109,15 +109,17 @@ private function fillEndpoint(?string $endpoint): string
     /**
      * Fallback function for getting the endpoint. This could be overridden by any APIClient.
      *
-     * @param array $uri   parameters that should go in the URI
+     * @param string $uri   parameters that should go in the URI
      * @param array $query parameters that should go in the query string
      */
-    protected function getEndpoint(array $uri, array $query):?string
+    protected function getEndpoint(string $uri, array $query):?string
```
Nyholm: There is difference between `getEndpoint` and `fillEndpoint`. 

I wanted `getEndpoint` to be extended by special apis like s3. They have a custom URL..
But Im not super happy with my solution.

jderusse: I've the feeling that endpoints don't have to be customized (everything is available in API.json file).
I've tested my PR (which removes getEndpoint from Client) anbd both S3 and SQS works.

build/Generator/ApiGenerator.php
jderusse: yep, sorry

PR #21, opened 2020-02-01
S3 input objects

Nyholm:
Autogenerated =)

PR #23, opened 2020-02-01
Added generated classes for SQS::sendMessage()

Nyholm:
I took the improved generator from #21 and added `SqsClient::sendMessage()`

```
 ./generate new Sqs SendMessage
```

It is too late for me to verify if it is correct or how it looks. Let's see tomorrow

### DIFF
--- a/src/Sqs/Input/SendMessageRequest.php
+++ b/src/Sqs/Input/SendMessageRequest.php
@@ -1,0 +1,177 @@
+<?php
+
+namespace AsyncAws\Sqs\Input;
+
+class SendMessageRequest
+{
+    /**
+     * @required
+     *
+     * @var string|null
+     */
+    private $QueueUrl;
+
+    /**
+     * @required
+     *
+     * @var string|null
+     */
+    private $MessageBody;
+
+    /**
+     * @var int|null
+     */
+    private $DelaySeconds;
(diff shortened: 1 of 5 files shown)


src/Sqs/Input/SendMessageRequest.php
```diff
@@ -0,0 +1,177 @@
...
+
+        return $this;
+    }
+
+    public function requestHeaders(): array
+    {
+        $headers = [];
+
+        return $headers;
+    }
```
Nyholm: These request methods should not be empty

src/Sqs/SqsClient.php
```diff
@@ -1,23 +1,35 @@
...
+     *   MessageDeduplicationId?: string,
+     *   MessageGroupId?: string,
+     * }|SendMessageRequest $input
+     */
+    public function sendMessage($input): SendMessageResult
     {
-        return 'sqs';
+        $input = SendMessageRequest::create($input);
+
+        $response = $this->getResponse('POST', '', $input->requestHeaders(), $this->getEndpoint($input->requestUri(), $input->requestQuery()));
```
Nyholm: Body is missing

Nyholm:
Fixed in #28

PR #33, opened 2020-02-01
Allow you to configure trait regeneration

Nyholm:
This is just some refactoring and code cleanup


src/Sqs/Result/SendMessageResultTrait.php
```diff
@@ -10,8 +10,6 @@ protected function populateFromResponse(ResponseInterface $response): void
     {
         $data = new \SimpleXMLElement($response->getContent(false));
 
-        // TODO Verify correctness
-        $data = $data->SendMessageResult;
```
jderusse: this should have been generated by the `wrapper` thing

Nyholm: Thank you!

PR #34, opened 2020-02-01
Validate request input

### LINKED ISSUE #26: Throw exception if not all require parameters exists in a request object
This should not be valid: 
```php
$input = new \AsyncAws\S3\Input\GetObjectRequest();
$input->setBucket('bref-example');
$result = $s3->getObject($input);
```

Currently it will be a "listObject", which is way wrong..

Nyholm:
This will fix #26. 

There is an issue with this. 

If you do a `PutObjectAclRequest` we validate all required parameters (['Bucket', 'Key']). But you could pass an `AccessControlPolicy` which have a list of `Grant`. Each `Grant` has a `Grantee` which has one required parameter ('$Type'). We never validate `$Type` since `AccessControlPolicy` is not required. 

This is a minor thing but I think it should be documented.

### DIFF
--- a/build/Generator/ApiGenerator.php
+++ b/build/Generator/ApiGenerator.php
@@ -235,7 +235,7 @@ foreach ([$requiredArray] as \$name) {
     }
 }
 PHP;
-        $class->addMethod('validate')->setPublic()->setReturnType('void')->setBody(empty($requiredArray) ? '' : $validateBody);
+        $class->addMethod('validate')->setPublic()->setReturnType('void')->setBody(empty($requiredProperties) ? '// There are no required properties' : $validateBody);
 
         $this->fileWriter->write($namespace);
     }
--- a/build/Generator/ApiGenerator.php
+++ b/build/Generator/ApiGenerator.php
@@ -163,6 +163,7 @@ PHP
         $constructor->addComment('} $input');
         $constructor->addParameter('input')->setType('array')->setDefaultValue([]);
         $constructorBody = '';
+        $requiredProperties = [];
 
         foreach ($members as $name => $data) {
             $parameterType = $members[$name]['shape'];
@@ -185,6 +186,7 @@ PHP
 
             $property = $class->addProperty($name)->setPrivate();
             if (\in_array($name, $shapes[$className]['required'] ?? [])) {
+                $requiredProperties[] = $name;
                 $property->addComment('@required');
             }
             $property->addComment('@var ' . $parameterType . ($nullable ? '|null' : ''));
@@ -213,6 +215,7 @@ PHP
         }
 
         $constructor->setBody($constructorBody);
+        $class->addConstant('REQUIRED_PARAMETERS', $requiredProperties)->setPublic();
         if ($root) {
             $this->inputClassRequestGetters($inputShape, $class, $operationName);
         }
@@ -369,6 +372,7 @@ PHP;
     {
         $body = <<<PHP
 \$input = $inputClassName::create(\$input); 
+\$this->validate(\$input);
 
 PHP;
 
--- a/build/Generator/ApiGenerator.php
+++ b/build/Generator/ApiGenerator.php
@@ -222,20 +222,32 @@ PHP
 
         // Add validate()
         $namespace->addUse(InvalidArgument::class);
-        $requiredArray = '\'' . implode("', '", $requiredProperties) . '\'';
+        $validateBody = '';
 
-        $validateBody = <<<PHP
+        if (!empty($requiredProperties)) {
+            $requiredArray = '\'' . implode("', '", $requiredProperties) . '\'';
+
+            $validateBody = <<<PHP
 foreach ([$requiredArray] as \$name) {
-    if (null === \$value = \$this->\$name) {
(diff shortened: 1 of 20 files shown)


src/Core/AbstractApi.php
```diff
@@ -135,4 +135,28 @@ protected function getEndpoint(string $uri, array $query):?string
...
+            // There is nothing we can do
+            return;
+        }
+        $requirements = \constant($class . '::REQUIRED_PARAMETERS');
+        if (!\is_array($requirements)) {
+            throw new InvalidArgument(sprintf('Constant "" must be an array.', $class . '::REQUIRED_PARAMETERS'));
+        }
+
+        foreach ($requirements as $name) {
+            $method = 'get' . $name;
```
jderusse: this could be faster at runtime if we generate a method `isValid()`. 

By implementing the logic here we performs a lot of function call and logic that could have been done when generating the file.

WDYT?

Nyholm: I was first planning to have an `AbstractInput` with an `isValid()` function to do these checks. But I changed my mind since I didnt feel like creating a new class =).

Nyholm: I'll make sure to generate these.

Nyholm:
PR is updated

src/S3/Input/CopyObjectRequest.php
```diff
@@ -859,4 +861,17 @@ public function requestUri(): string
...
     }
+
+    public function validate(): void
+    {
+        foreach (['Bucket', 'CopySource', 'Key'] as $name) {
+            if (null === $value = $this->$name) {
+                throw new InvalidArgument(sprintf('Missing parameter "%s" when validating the "%s". The value cannot be null.', $name, __CLASS__));
+            }
+
+            if (\is_object($value) && method_exists($value, 'validate')) {
```
jderusse: that could be guessed at build time isn't it?

src/S3/Input/CopyObjectRequest.php
```diff
@@ -859,4 +861,17 @@ public function requestUri(): string
 
         return "/{$uri['Bucket']}/{$uri['Key']}";
     }
+
+    public function validate(): void
+    {
+        foreach (['Bucket', 'CopySource', 'Key'] as $name) {
+            if (null === $value = $this->$name) {
+                throw new InvalidArgument(sprintf('Missing parameter "%s" when validating the "%s". The value cannot be null.', $name, __CLASS__));
```
jderusse: Shouldn't we add the className? When validating a sub-object, the name of the parameter could be meaningless

src/S3/Input/CopyObjectRequest.php
Nyholm: Could.. but it is a bit tricky.. I rather leave that for a future improvement.

src/S3/Input/CopyObjectRequest.php
Nyholm: We are using `__CLASS__ `. Is there need for something more?

build/Generator/ApiGenerator.php
```diff
@@ -217,6 +220,23 @@ private function generateInputClass(string $service, string $operationName, stri
             $this->inputClassRequestGetters($inputShape, $class, $operationName);
         }
 
+        // Add validate()
+        $namespace->addUse(InvalidArgument::class);
+        $requiredArray = '\'' . implode("', '", $requiredProperties) . '\'';
+
+        $validateBody = <<<PHP
```
jderusse: whats about?
```php
foreach ($requiredProperties as $property) {
   $validateBody.= "if (null === \$this->$property) throw ...;";
   if ($defintions[$property]['type'] === 'struct') {
     $validateBody.=' \$this->$property->validate();";
   }
}

File: src/S3/Input/CopyObjectRequest.php (hunk 3 above)
Comment (jderusse) [reply src/S3/Input/CopyObjectRequest.php]: nvm.. sorry for the noise

File: build/Generator/ApiGenerator.php (hunk 4 above)
Comment (Nyholm) [reply build/Generator/ApiGenerator.php]: Oh, that is correct. I was looking to figure things out in the return type. 

I'll give it a try

Comment (Nyholm) [reply build/Generator/ApiGenerator.php]: Fixed it!

**Comment by Nyholm** [comment]
The issue I described in the initial comment is not fixed.

File: src/S3/Input/AccessControlPolicy.php [hunk 5]
```diff
@@ -56,4 +56,14 @@ public function setOwner(?Owner $value): self
...
+
+    public function validate(): void
+    {
+        foreach ($this->Grants as $item) {
+            $item->validate();
+        }
+        if ($this->Owner) {
+            $this->Owner->validate();
+        }
+    }
```
Nyholm: Here we validate `Grants` and `Owner` even though they are not required. 

So, if they exist, they have to be valid.

jderusse: good catch

jderusse:
thank you

Nyholm:
Thank you for the reviews and suggestions

PR #36, opened 2020-02-01
Throw better exception when response is resolved

Nyholm:
Or should we just return quietly?
That might be better..

What you you think?

### DIFF
--- a/src/Core/Result.php
+++ b/src/Core/Result.php
@@ -9,6 +9,7 @@ use AsyncAws\Core\Exception\Http\ClientException;
 use AsyncAws\Core\Exception\Http\NetworkException;
 use AsyncAws\Core\Exception\Http\RedirectionException;
 use AsyncAws\Core\Exception\Http\ServerException;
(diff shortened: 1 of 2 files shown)


Nyholm:
Alternative PR #37

src/Core/Result.php
```diff
@@ -74,6 +79,10 @@ protected function populateFromResponse(ResponseInterface $response): void
 
     final public function cancel(): void
     {
+        if (null === $this->response) {
+            throw new LogicException('The responses is already resolved.');
```
jderusse: I think this one should be silent, because the opposite is not true:
If the response is not null and you call cancel, you won't get any error. But it doesn't mean that the request will be trully canceled: there are possibility that the request is already finished (at least from on AWS servers) and you cancels nothing.

If canceling a request that is already resolved on AWS i silent, then canceling a request that is already resolve in tha Object should be silent too.

jderusse:
Closed by #37

PR #38, opened 2020-02-01
Merge initialize and resolve

jderusse:
`Resolve` and `Initialize` performs (almost) the same thing. (Resolve is the "ligth" version that does not hydrate the object).
I wonder if we should keet those 2 methods, or if we shouldn't always Hydrate  the object. 

Moreover one could perform a `->resolve()` then expect to get working object  and want to serialize/or var_dump it.

This PR merges both methods together. (and also fix issues in `resolve` and `cancel` to deal with null response)


Nyholm:
There is a difference here though. 
If you resolve, you make sure the request is sent. When you initialize, you download the response. 

If there is a 1mb response body, you will notice a difference. 

I’m all for merging, but I want a way to send a request without downloading response header or body.

jderusse:
Then I sugest to either revert this PR (which does not make sens if we introduce a new method afterward) or Introduce a new `info` method that return the current status of the request.

Nyholm:
I think in `info()` method makes sense. 

It should give the user "some" information about the request, maybe even the request object?

jderusse:
added method info.. (very very basic for now=

Nyholm:
Great. 
But the PR still has the issue that I cannot send a request without downloading the response body.

jderusse:
> But the PR still has the issue that I cannot send a request without downloading the response body.

how about?
```php
$info = $response->info();
if (isset($info['response'])) $info['response']->getStatusCode();
```

jderusse:
maybe too complicated... let's close it

src/Core/Result.php
```diff
@@ -74,6 +61,35 @@ protected function populateFromResponse(ResponseInterface $response): void
         if (300 <= $statusCode) {
             throw new RedirectionException($this->response);
         }
+
+        $this->populateFromResponse($this->response);
```
Nyholm: Here is the issue. As soon as we run resolve, we execute this function. 

This function starts to take stuff from the response. 

* `$response->getStatusCode()` - Download only the first line of the request. 
* `$response->getHeaders()` - Download the headers
* `$response->getContent()` - Download the body

So if you never call `$response->getContent()` you never download the body. But we call `$response->getHeaders()` and `$response->getContent()` as soon as we resolve the Result.

PR #39, opened 2020-02-01
Fix CI 

jderusse:


jderusse:
Remaining faillure will be fixed in dedicated PR

PR #40, opened 2020-02-01
Fix PSALM issues

jderusse:
Thanks PSALM... spoted several issues in `Core`

### DIFF
--- a/src/Core/AwsClient.php
+++ b/src/Core/AwsClient.php
@@ -22,6 +22,17 @@ class AwsClient extends AbstractApi
      */
     private $serviceCache;
 
+    protected function getServiceCode(): string
+    {
+        // This will never work on the base API. .
+        throw new RuntimeException(sprintf('The $endpoint parameter is required on "%s::request()".', __CLASS__));
+    }
+
+    protected function getSignatureVersion(): string
+    {
+        return 'v4';
+    }
+
     public function s3(): S3Client
     {
         if (!class_exists(S3Client::class)) {
@@ -59,11 +70,5 @@ class AwsClient extends AbstractApi
         }
 
         return $this->serviceCache[__METHOD__];
-    }
-
-    protected function getServiceCode(): string
-    {
-        // This will never work on the base API. .
-        throw new RuntimeException(sprintf('The $endpoint parameter is required on "%s::request()".', __CLASS__));
     }
 }
--- a/psalm.xml
+++ b/psalm.xml
@@ -10,6 +10,7 @@
         <directory name="src" />
         <ignoreFiles>
             <directory name="src/CodeGenerator" />
+            <directory name="src/*/Tests" />
             <directory name="vendor" />
         </ignoreFiles>
     </projectFiles>
--- a/src/CodeGenerator/Generator/ApiGenerator.php
+++ b/src/CodeGenerator/Generator/ApiGenerator.php
@@ -61,6 +61,27 @@ class ApiGenerator
         $namespace->addUse($inputClass);
         $classes = $namespace->getClasses();
         $class = $classes[\array_key_first($classes)];
+        if (null !== $prefix = $definition->getEndpointPrefix()) {
+            if (!$class->hasMethod('getServiceCode')) {
+                $class->addMethod('getServiceCode')
+                    ->setReturnType('string')
+                    ->setVisibility(ClassType::VISIBILITY_PROTECTED)
+                ;
+            }
+            $class->getMethod('getServiceCode')
+                ->setBody("return '$prefix';");
+        }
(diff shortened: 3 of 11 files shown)


build/Generator/ApiGenerator.php
```diff
@@ -61,6 +61,13 @@ public function generateOperation(ServiceDefinition $definition, string $service
         $namespace->addUse($inputClass);
         $classes = $namespace->getClasses();
         $class = $classes[\array_key_first($classes)];
+        if (null !== $version = $definition->getSignatureVersion()) {
```
Nyholm: We can also check if the `$class` already have this function.

build/Generator/ApiGenerator.php
```diff
@@ -61,6 +61,13 @@ public function generateOperation(ServiceDefinition $definition, string $service
...
+            $class->getMethod('getSignatureVersion')
+                ->setReturnType('string')
+                ->setVisibility(ClassType::VISIBILITY_PROTECTED)
+                ->setBody('return ' . \var_export($version, true) . ';');
+        }
```
Nyholm: You can do the very same thing with service name, right?

src/Core/AwsClient.php
```diff
@@ -4,23 +4,61 @@
...
 use AsyncAws\Sqs\SqsClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 /**
  * Base API client that instantiate other API classes if needed.
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class AwsClient extends AbstractApi
+class AwsClient
```
Nyholm: Could you elaborate here? I let it extend the AbstractApi because I didnt want to duplicate this code. 

I also want to support something like this: 

```php
use AsyncAws\Core\AwsClient;
use Symfony\Component\HttpClient\HttpClient;

$awsClient = new AwsClient(HttpClient::create(), [
    'region' => 'eu-central-1',
    'accessKeyId' => 'foo',
    'accessKeySecret' => 'bar',
]);

$result = $awsClient->request('POST', [
    'Action' => 'SendMessage',
    'MessageBody' => 'foobar'
], 
[
  /* headers */
], 
 'https://sqs.eu-central-1.amazonaws.com/5555555555/my_queue'
);
```

build/Generator/ApiGenerator.php
jderusse: The class must have this function (because method is abstract). But here, I want to always update, and make sure that we use the recommended version (in case AWS change it)

Nyholm: Are you sure `$class->hasMethod('getSignatureVersion')` always will return true? Im actually hesitant.

build/Generator/ApiGenerator.php
jderusse: exact

src/Core/AwsClient.php
jderusse: Hmm.. I didn't think about that.. revert..

I reverted that change, and defined an arbitrary SignatureVersion.

In fact, the example won't work, as the request need to be signed and for that purpose needs a serviceName...

Nyholm: Hm.. Yeah, I wrote the example before the code. 

We should make sure that works... somehow.. Or remove that class.

build/Generator/ApiGenerator.php
```diff
@@ -61,6 +61,19 @@ public function generateOperation(ServiceDefinition $definition, string $service
         $namespace->addUse($inputClass);
         $classes = $namespace->getClasses();
         $class = $classes[\array_key_first($classes)];
+        if (null !== $prefix = $definition->getEndpointPrefix()) {
+            $class->getMethod('getServiceCode')
+                ->setReturnType('string')
+                ->setVisibility(ClassType::VISIBILITY_PROTECTED)
+                ->setBody('return ' . \var_export($prefix, true) . ';');
```
Nyholm: Why `var_export`?  This will always be a string, right?

Btw, This is the first time I see `var_export` =)

build/Generator/ApiGenerator.php
jderusse: if not, the class is not loadable : 
`PHP Fatal error:  Class AsyncAws\Sqs\SqsClient contains 1 abstract method and must therefore be declared abstract or implement the remaining methods`

But you're right. I'll add a test

build/Generator/ApiGenerator.php
jderusse: this is a very safe way to escape a variable.

symfony uses var_export a lot when it generates code.

build/Generator/ApiGenerator.php
Nyholm: Thank you

PR #42, opened 2020-02-01
Added meta files and readmes

Nyholm:


src/CodeGenerator/.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
jderusse: If the project is a sub-tree split, what's the purpose of this file (especially the `pull_request` option

Nyholm: `pull_request`  has no meaning. But we will get nice badges in the readme =)

PR #46, opened 2020-02-01
Add Info

jderusse:
provides informations on current responses

### DIFF
--- a/src/Core/Result.php
+++ b/src/Core/Result.php
@@ -23,8 +23,6 @@ class Result
      */
     private $response;
 
-    private $initialized = false;
-
     public function __construct(ResponseInterface $response)
     {
         $this->response = $response;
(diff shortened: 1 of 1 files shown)


src/Core/Result.php
```diff
@@ -76,6 +73,32 @@ protected function populateFromResponse(ResponseInterface $response): void
...
+     */
+    final public function info(): array
+    {
+        if (!isset($this->response)) {
+            return [
+                'resolved' => true,
+            ];
+        }
+
+        $info = $this->response->getInfo();
```
Nyholm:
```suggestion
        $httpStatus = $this->response->getInfo('http_code');
```

src/Core/Result.php
```diff
@@ -76,6 +73,32 @@ protected function populateFromResponse(ResponseInterface $response): void
...
+
+        return [
+            'resolved' => false,
+            'response' => $this->response,
+            'status' => $info['http_code'],
```
Nyholm:
```suggestion
            'status' => $httpStatus,
```

src/Core/Result.php
jderusse: Cool. I didn't know that shortcut.

Don't be mad at me, I didn't pick your suggestio because didn't used the intermediate variables.

Nyholm: Oh. I’m not mad at all. 
I’m happy you are delivering so much excellence. :)

PR #49, opened 2020-02-01
CI updates on subtreesplit repos

Nyholm:


Nyholm:
I want to prevent this: https://github.com/async-aws/sqs/runs/421223429?check_suite_focus=true#step:4:12

PR #50, opened 2020-02-01
Speed up build by running fewer checks

Nyholm:


Nyholm:
Finally down to 7 checks, grouped output and clear error messages.

jderusse:
great, it cleaner.!

I wonder if it will be faster?. previous job matrix were running in parallel, now they are running in sequence.

what's about skipping job if files in the component is  not modified?

Nyholm:
Oh. Checking if modified is not always preferred. What if we only modify core but tests in SQS fail?

I was planning to use GNU parallel but I could not install in :/

It is faster since sometimes jobs didn’t start because others were running. We’ve also remove a lot of “setup” and initialization time

jderusse:
I was thinking about
```
on:
  push:
    branches:
      - master
  pull_request:
    paths:
    - 'src/Core/**'
    - 'src/${{ matrix.component }}/**'
```

I don't know if it works...

Nyholm:
That will still explode the number of jobs. 

We could ( symfony does) run all php versions in the same job. They are installed on the job. But that might be overkill. 

We still need to add a job for “lowest deps” and BC check.

PR #53, opened 2020-02-02
Fixed Psalm issues

Nyholm:
There is one todo, I'll write an issue


src/Core/Credentials/IniFileProvider.php
```diff
@@ -91,6 +91,7 @@ private function loadProfiles(array $filepaths): array
                 continue;
             }
             if ('~' === $filepath[0]) {
+                // FIXME is $homeDir ever defined?
                 $homeDir = $homeDir ?? $this->getHomeDir();
```
Nyholm: This should be fixed

jderusse: what's the issue ?

PR #57, opened 2020-02-02
Add Sts Client

jderusse:
This client is required to create Credentials and Assuming Role.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -48,7 +48,17 @@
             "source": "https:\/\/raw.githubusercontent.com\/aws\/aws-sdk-php\/3.133.6\/src\/data\/sqs\/2012-11-05\/api-2.json",
             "methods": {
                 "SendMessage": {
-                    "generated": "2020-02-01T16:27:01+00:00",
+                    "generated": "2020-02-02T15:37:23+01:00",
+                    "generate-result-trait": true
+                }
+            }
+        },
+        "Sts": {
+            "source": "https:\/\/raw.githubusercontent.com\/aws\/aws-sdk-php\/3.133.6\/src\/data\/sts\/2011-06-15\/api-2.json",
+            "namespace": "AsyncAws\\Core",
+            "methods": {
+                "GetCallerIdentity": {
+                    "generated": "2020-02-02T16:45:58+01:00",
                     "generate-result-trait": true
                 }
             }
--- a/src/CodeGenerator/Generator/ApiGenerator.php
+++ b/src/CodeGenerator/Generator/ApiGenerator.php
@@ -21,11 +21,6 @@ use Symfony\Contracts\HttpClient\ResponseInterface;
 class ApiGenerator
 {
     /**
-     * @var string
-     */
-    private $srcDirectory;
-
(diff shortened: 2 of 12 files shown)


manifest.json
```diff
@@ -48,7 +48,17 @@
...
                 "SendMessage": {
-                    "generated": "2020-02-01T16:27:01+00:00",
+                    "generated": "2020-02-02T15:37:23+01:00",
+                    "generate-result-trait": true
+                }
+            }
+        },
+        "Sts": {
+            "source": "https:\/\/raw.githubusercontent.com\/aws\/aws-sdk-php\/3.133.6\/src\/data\/sts\/2011-06-15\/api-2.json",
+            "namespace": "AsyncAws\\Core",
```
jderusse: Use the `\Core` namespace because will be required For authentication. Moving it to a dedicated endpoint will ended in a Circular Dependency

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -91,7 +84,7 @@ public function generateOperation(ServiceDefinition $definition, string $service
         $method->addComment('@param array{');
         $this->addMethodComment($method, $inputShape, $baseNamespace . '\\Input');
         $method->addComment('}|' . $inputClassName . ' $input');
-        $method->addParameter('input');
+        $method->addParameter('input')->setDefaultValue([]);
```
jderusse: avoid calling `->method([])` when method does not need arguments

src/Core/Configuration.php
```diff
@@ -44,6 +44,7 @@ class Configuration
     ];
 
     private const DEFAULT_OPTIONS = [
+        self::OPTION_REGION => 'us-east-1',
```
jderusse: Some services are global and don't need region (ie. STS).

I hesitate with:
- fallback to `us-east-1` when the Service is global
- trigger an exception when the service requires a region.

WDYT?

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -273,12 +266,12 @@ private function generateInputClass(string $service, string $operationName, stri
...
     }
 
-    private function inputClassRequestGetters(array $inputShape, ClassType $class, string $operation): void
+    private function inputClassRequestGetters(array $inputShape, ClassType $class, string $operation, string $apiVersion): void
     {
         foreach (['header' => '$headers', 'querystring' => '$query', 'payload' => '$payload'] as $requestPart => $varName) {
             $body[$requestPart] = $varName . ' = [];' . "\n";
             if ('payload' === $requestPart) {
-                $body[$requestPart] = $varName . " = ['Action' => '$operation'];\n";
+                $body[$requestPart] = $varName . " = ['Action' => '$operation', 'Version' => '$apiVersion'];\n";
```
jderusse: Services STS requires to have this version in the payload :/

src/Core/Configuration.php
Nyholm: This is fine =)

PR #59, opened 2020-02-02
Add InstanceProfile authentication

jderusse:


src/Core/Credentials/InstanceProvider.php
```diff
@@ -0,0 +1,82 @@
...
+use AsyncAws\Core\Configuration;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\Exception\HttpExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * Provides Credentials from the running EC2 metadata server using the IMDS version 1
```
Nyholm: Excellent!

jderusse: Version 2 is little more complicated (not implemented by golang sdk, but is implemented by php sdk)

PR #62, opened 2020-02-02
Fix prefer-lowest test

jderusse:


Nyholm:
I agree with this, but it is just temporary, right? This should be removed when we tagged a version of each package

PR #68, opened 2020-02-02
Make body Streamable

### LINKED ISSUE #60: Find a solution for steaming responses
Currently, the only way to stream a response is to:

```php
$result = ...;
$info = $result->getInfo();
foreach($httpClient->stream($info['response']) as $chunk) {
  // ...
}
```

Do we want to provide a better way?

Nyholm:
I know I said we should wait with #60. But I couldn't stop thinking about it. 

This will fix #60

### DIFF
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -350,7 +350,7 @@ class S3Client extends AbstractApi
      *   PartNumber?: int,
      * }|GetObjectRequest $input
      */
-    public function getObject($input = []): GetObjectOutput
+    public function getObject($input): GetObjectOutput
     {
         $input = GetObjectRequest::create($input);
         $input->validate();
--- a/src/S3/S3Client.php
+++ b/src/S3/S3Client.php
@@ -238,46 +238,6 @@ class S3Client extends AbstractApi
     }
 
     /**
-     * @see http://docs.amazonwebservices.com/AmazonS3/latest/API/RESTObjectGET.html
-     *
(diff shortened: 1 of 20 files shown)


src/S3/S3Client.php
```diff
@@ -237,46 +237,6 @@ public function copyObject($input = []): CopyObjectOutput
...
-     *   ResponseExpires?: \DateTimeImmutable|string,
-     *   VersionId?: string,
-     *   SSECustomerAlgorithm?: string,
-     *   SSECustomerKey?: string,
-     *   SSECustomerKeyMD5?: string,
-     *   RequestPayer?: string,
-     *   PartNumber?: int,
-     * }|GetObjectRequest $input
-     */
-    public function getObject($input = []): GetObjectOutput
```
jderusse: Is their a way to tell avoid such diff?
- tell to `nette/php-generator` to sort methods?
- add a php-cs-fixer on `/*/*Client` to sort methods?

Nyholm:
This PR is wip

src/S3/S3Client.php
Nyholm: > tell to nette/php-generator to sort methods?

No, and there is a massive "copy code" to achieve that. 

> add a php-cs-fixer on /*/*Client to sort methods?

I dont know. Can php-cs-fixer sort? Why not sort everything then?

Nyholm:
This PR is ready for review

PR #71, opened 2020-02-02
Adds a test that asserts output has been regenerated

jderusse:


.github/workflows/checks.yml
```diff
@@ -46,6 +46,40 @@ jobs:
...
+
+    - name: Fix cs
+      uses: docker://oskarstark/php-cs-fixer-ga
+
+    - name: Assert up-to-date
+      run: |
+        if [ -z "$(git status src --porcelain)" ]; then exit 0; fi
+
+        git diff src
+        exit 1
```
Nyholm: This will always fail

jderusse: yes. That's the purpose. It exist with 0 at line 79 when every thing is OK.

I'll inverse the IF block

jderusse:
I don't know how to deal with PR that update the generated file...

From one side: The Diff is unreadable. I would like to forbid human touching the generated code, and setup a job/bot that automatically generate output when a PR is merged and create a dedicated PR....

BUT on another hand, the output is sometime very help full to understand the change...

.github/workflows/checks.yml
Nyholm: Ah, okey. I see.

jderusse:
WIP. Test should failed... I've to diff tomorrow

PR #81, opened 2020-02-03
Add documentation

jderusse:
This PR adds Documentation on top of Client's methods and Objects properties:

2 commit in the PR: 
- First for businesslogic in CodeGenerate
- Second to regenerate files

- Only the first paragraph is displayed (some endpoint are VERY verbose)
- simple algo to convert html to markdown converter
- link (from the first paragraphe) are reported as `@see` comment

I wonder if the property is the right place for the comment. Should'it be the setter? or Constructor (how is this compatible with psalm doc bloc?)


src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -229,6 +274,10 @@ private function generateInputClass(string $service, string $apiVersion, string
             }
 
             $property = $class->addProperty($name)->setPrivate();
+            if (null !== $propertyDocumentation = $this->definition->getParameterDocumentation($className, $name, $data['shape'])) {
```
Nyholm: I think it makes sense to have it on the property for the input class. 

:+1:

src/Core/Sts/Result/AssumeRoleResponse.php
```diff
@@ -8,8 +8,23 @@
 
 class AssumeRoleResponse extends Result
 {
+    /**
+     * The temporary security credentials, which include an access key ID, a secret access key, and a security (or session)
+     * token.
+     */
     private $Credentials;
+    /**
```
Nyholm: Oh no, I dont like this CS anymore =)

Can you remove this line from .php.cs: `'class_attributes_separation' => ['elements'=>['method']],`?

That will add an extra space between properties =)

.php_cs
```diff
@@ -70,7 +70,7 @@ return PhpCsFixer\Config::create()
         'single_quote' => true,
         'phpdoc_to_comment' => false,
         'ordered_class_elements' => ['sortAlgorithm'=>'none'],
-        'class_attributes_separation' => ['elements'=>['method']],
+        'class_attributes_separation' => ['elements'=>['property']],
```
Nyholm: This line could be removed. Then we fallback on symfony standard which is the same as: 

```
'class_attributes_separation' => ['elements'=>['property', 'method', 'const']],
```

jderusse: Hmm..  when I removed it, nothing changed

PR #82, opened 2020-02-03
Add support for Streamable input

### LINKED ISSUE #73: Support for streamable input
It is just a minor thing. We should remove return type hint on parameters like `PutObjectRequest::getBody()`. 

We should also update the constructor doc block. 

The correct signature is `array|string|resource|\Traversable|\Closure`

Nyholm:
This will fix #73

I've not generated output yet. Waiting for #81 to be merged.


Nyholm:
I changed my mind. This diff is small. I added the generated code.

Nyholm:
Thank you

PR #83, opened 2020-02-03
Fix object generation with sub-object

jderusse:


jderusse:
I'wasn't confident either. replace recursion by updating the prefix

PR #97, opened 2020-02-05
Handle pagination

jderusse:
This is the very first layer of pagination.

It provide an iterrator that iterate over the items in the current page. Calling API to fetch next page will be done in another PR.

For the user, when the operation returns a paginated response, the they receive an \IteratorAggregate result
usage:

```php
foreach ($sqsClient->ListQueues() as $queueUrl) {
  echo $queueUrl .PHP_EOL;
}

$list = $s3Client->ListObjects();
foreach ($list as $object) {
  echo $list->getName() . $object->getPrefix().$object->getKey() .PHP_EOL;
}
```

nb: Some operation returns the "merge" of 2 lists (see `S3::ListObjects`).
ListObjectsOutput contains:
- Contents: Objects[]
- CommonPrefixes: CommonPrefixes[]

When iterating, my implementation returns a new class `ObjectCommonPrefixes[]` which decorates the 2 Class..

WDYT?

### DIFF
--- a/src/CodeGenerator/Generator/ApiGenerator.php
+++ b/src/CodeGenerator/Generator/ApiGenerator.php
@@ -11,6 +11,7 @@ use AsyncAws\Core\StreamableBody;
 use AsyncAws\Core\XmlBuilder;
 use Nette\PhpGenerator\ClassType;
 use Nette\PhpGenerator\Method;
+use Nette\PhpGenerator\Parameter;
 use Nette\PhpGenerator\PhpNamespace;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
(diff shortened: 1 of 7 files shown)


src/S3/Result/ListObjectsOutput.php
```diff
@@ -105,6 +105,22 @@ public function getIsTruncated(): ?bool
         return $this->IsTruncated;
     }
 
+    /**
+     * Iterates over Contents and CommonPrefixes.
+     *
+     * @return ObjectAndCommonPrefix[]
```
jderusse: phpstan is not happy with this. Can you help me?

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -200,10 +200,135 @@ private function doGenerateResultClass(string $baseNamespace, string $className,
...
+                if ('structure' !== $listShape['type']) {
+                    throw new \RuntimeException('This is not implemented yet');
+                }
+                $type = $this->safeClassName($resultShape['member']['shape']);
+                $mergedTypes[] = $resultShape['member']['shape'];
+            }
+
+            $mergedClass = $this->generateMergeClass($mergedTypes, $baseNamespace);
+            $namespace = new PhpNamespace($baseNamespace);
+            $namespace->addUse($baseNamespace . $mergedClass);
```
Nyholm: What are we doing with this namespace?

src/S3/Result/ListObjectsOutput.php
```diff
@@ -105,6 +105,22 @@ public function getIsTruncated(): ?bool
...
+     */
+    public function getIterator(): \Traversable
+    {
+        $this->initialize();
+
+        return (function () {
+            foreach ($this->Contents as $index => $item) {
+                yield new ObjectAndCommonPrefix($item, $this->CommonPrefixes[$index] ?? null);
```
Nyholm: Is  the `$this->Contents` array always longer than `$this->CommonPrefixes`?

src/S3/Result/ListObjectsOutput.php
Nyholm: How about something like `@return \Traversable<int, ObjectAndCommonPrefix >`?

Nyholm: Using `\Traversable<ObjectAndCommonPrefix>` is good enough. Both will work

src/CodeGenerator/Generator/ApiGenerator.php
jderusse: this namespace contains the class of the Output object and will use the Merged Class.
It's not useful for now given both classes are in the same namespace but it will be an issue if Merged class is moved somewhere else

Nyholm: Hm.. Sure, but a `PhpNamespace` is something we eventually should print. As far as I can see, these two lines does not have an effect on the raw output. Ie, it does not matter if the MergedClass is moved somewhere else, these lines will not effect anything.

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -200,10 +200,135 @@ private function doGenerateResultClass(string $baseNamespace, string $className,
...
+
+            $returnType = "{$mergedClass}";
```
Nyholm:
```suggestion
            $returnType = $mergedClass;
```

Isn't this the same?

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -679,7 +810,7 @@ private function resultClassAddProperties(string $baseNamespace, bool $root, ?ar
                     $this->doGenerateResultClass($baseNamespace, $listItemShapeName);
                     $returnType = $this->safeClassName($listItemShapeName);
                 } else {
-                    $returnType = $type . '[]';
+                    $returnType = $type;
```
Nyholm: Are you sure? 

Isn't it an array of `$type`?

src/S3/Result/ListObjectsOutput.php
jderusse: Hmmmm... I totaly missunderstood how it works. After investigating: . In fact when `result_key` contains several keys, they are not related..
For instance, when you call `s3::ListObject`, you have 2 list: `contents` that contains the objects and `commonPrefixes` that contains list of "folders". Mixing theme does not make sens. 

I suggest to either iterrating over Content then CommonPrefixes and the user have to deal with `if($item instanceOf AwsObject) ...`
Or to provide 2 methods: getContentAsTraversable and getCommonPrefixesAsTraversable


WDYT?

src/CodeGenerator/Generator/ApiGenerator.php
jderusse: I applied your sugestion, a `\Traversable<int, {$returnType}>`. Maybe the variable needs a better name

src/CodeGenerator/Generator/ApiGenerator.php
jderusse: indeed... removed

src/S3/Result/ListObjectsOutput.php
Nyholm: Let’s iterate over them both. That is how flysystem does. 

If you really want to have only one object type, you still can... (sure, but without pagination)

src/CodeGenerator/Generator/ApiGenerator.php
Nyholm: Oh. That is why this line was changed. 

Good good. It is just me that was confused.

src/S3/Result/GetObjectAclOutput.php
```diff
@@ -56,7 +56,7 @@ protected function populateResult(ResponseInterface $response, ?HttpClientInterf
             'ID' => $this->xmlValueOrNull($data->Owner->ID, 'string'),
         ]);
         $this->Grants = [];
-        foreach ($data->Grants as $item) {
+        foreach ($data->Grant as $item) {
```
Nyholm: I can’t understand why this is changed. Does $data->Grant even exist?

It is probably correct, it is just my ignorance.

jderusse:
I think I found the good the right balance. The `IteratorAggregate` yield over each list (the user have to check the type of the item).

AND the getter of each list expose a parameter that let the choice to the user if they want to fetch only the current page (and let them handle the pagination by their own) OR lazy load next pages.

```php
$list = $s3Client->listObjects(['Delimiter' => '/']);
foreach ($list as $item) {
 if ($item instanceof AwsObject) echo $item->getKey();
 if ($item instanceof CommonPrefixe) echo $item->getPrefix();
}

// iterate only over the current page
foreach ($list->getContents(true) as $item) {
 if ($item instanceof AwsObject) echo $item->getKey();
}

// iterate over the current page then load nexts page 
foreach ($list->getContents(false) as $item) {
 if ($item instanceof AwsObject) echo $item->getKey();
}
```

```

src/S3/Result/GetObjectAclOutput.php
jderusse: this is because of https://github.com/async-aws/aws/pull/97/files#diff-bcf76e97ba120d21d9167a6e0778079dR497

AWS returns `<xml><Grant></Grant</xml>` while the property is named `Grants`. This information is defined in the `locationName` property of the shape

Nyholm: Thank you

PR #99, opened 2020-02-05
Make sure Metadata is useful

### LINKED ISSUE #91: Handle "Metadata"
On the S3 HeadObjectOutput there is a parameter "Metadata" that is never set. 

I think, that all parameters on the object is the metadata..

### LINKED ISSUE #92: HeadObjectOutput::LastModified should be a \DateTimeImmutable
Like everything else marked as a "timestamp"

Nyholm:
This will fix #91 and fix #92 

~~It is basically an array of all properties. Have I understood this correctly?~~


Nyholm:
> It is basically an array of all properties. Have I understood this correctly?

No. Metadata is extra properties you can set on an S3 object. They must begin with `x-amz-meta-`. 
https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -719,15 +728,35 @@ private function resultClassPopulateResult(string $operationName, string $classN
...
             if (($member['location'] ?? null) !== 'header') {
                 $nonHeaders[$name] = $member;
 
                 continue;
             }
+            $headers[$name] = $member;
+
             $locationName = $member['locationName'] ?? $name;
-            $body .= "\$this->$name = \$headers['{$locationName}'];\n";
+            $body .= "\$this->$name = \$headers['{$locationName}'] ?? null;\n";
```
Nyholm: Header may not be defined, better use `??`

src/S3/Result/GetObjectOutput.php
```diff
@@ -395,35 +395,37 @@ protected function populateResult(ResponseInterface $response, ?HttpClientInterf
...
+        $this->StorageClass = $headers['x-amz-storage-class'] ?? null;
+        $this->RequestCharged = $headers['x-amz-request-charged'] ?? null;
+        $this->ReplicationStatus = $headers['x-amz-replication-status'] ?? null;
+        $this->PartsCount = $headers['x-amz-mp-parts-count'] ?? null;
+        $this->TagCount = $headers['x-amz-tagging-count'] ?? null;
+        $this->ObjectLockMode = $headers['x-amz-object-lock-mode'] ?? null;
+        $this->ObjectLockRetainUntilDate = $headers['x-amz-object-lock-retain-until-date'] ?? null;
+        $this->ObjectLockLegalHoldStatus = $headers['x-amz-object-lock-legal-hold'] ?? null;
+        $this->Metadata = [
+        ];
```
jderusse: is it expected to get an empty array?

Nyholm: No. I've updated the PR.

Nyholm:
Finally done. 

- All headers are lower key and multi dimension array. 
- I use `filter_var` on header result and body result
- I use `?? null` on headers to make sure they are null if not defined. 
- Support for S3 metadata. 

This will also fix #92

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -667,6 +667,14 @@ private function resultClassAddProperties(string $baseNamespace, bool $root, ?ar
             if ('structure' === $memberShape['type']) {
                 $this->doGenerateResultClass($baseNamespace, $parameterType);
                 $parameterType = $baseNamespace . '\\' . $this->safeClassName($parameterType);
+            } elseif ('map' === $memberShape['type']) {
```
jderusse: The ApiGenerator contains several occurence of `if type == 'structure` or `if type == 'list'`. Shouldn't we consider `map` also?

Nyholm: We maybe should. I write an issue.

Nyholm: See #100

Nyholm:
Thank you

PR #105, opened 2020-02-06
Add Basics SQS operations

jderusse:
Most of those operation are needed for a client like `symfony/messenger`
* SendMessage
* DeleteMessage
* ChangeMessageVisibility
* GetQueueUrl
* GetQueueAttributes
* CreateQueue

I also added those methods that could be usefull for testing or managing the broker
* DeleteQueue
* ListQueues
* PurgeQueue

Missing operation (ApiGeneartor fails to generate it... PR in progress)
* ReceiveMessage


Nyholm:
Ah, lovely to read generated code =)

PR #106, opened 2020-02-06
Refactor Result Generator

jderusse:
In order to generate the `Sqs::ReceiveMessage` operation, I had refactor the `parseXmlResponse` method..  a lot..

before the methods generated both computing result **and** affect result to the variable.
now it just compute the result, this is the responsability of the caller to affect the result. It allows to easilly use recursive and remove duplicate code.

This PR adds support for List, Map, and composition of each type (map of structure, list in structures);

To simplify the code, List (and map) generation are always wrapped in a closure, because sometime we can not use temporary variable;

```php
// this works
function __construct()
{
  $this->key = [];
  foreach ($data->Key as $item) {
    $this->key[] = $item;
  }
}

// this don't works
function __construct()
{
  $this->sub = new Foo([
    'key' => [], // ??... I can't iterate here...
  ])
}

// closures always works
function __construct()
{
  $this->sub = new Foo([
    'key' => (function(xml){
      $key = [];
      foreach ($data->Key as $item) {
        $key[] = $item;
      }
      return $key;
    })($this->sub)
  ])
}

```


Nyholm:
Could you rebase please?

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -487,59 +487,124 @@ private function inputClassRequestGetters(array $inputShape, ClassType $class, s
...
+                return $this->parseXmlResponseMap($shapeName, $input);
+            case 'string':
+            case 'boolean':
+            case 'integer':
+            case 'long':
+            case 'timestamp':
+            case 'blob':
+                return $this->parseXmlResponseScalar($shapeName, $input);
+            default:
+                throw new \RuntimeException(sprintf('Type %s is not yet implemented', $shape['type']));
```
Nyholm: Good.

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -487,59 +487,124 @@ private function inputClassRequestGetters(array $inputShape, ClassType $class, s
...
+        foreach ($shape['members'] as $memberName => $memberData) {
+            if (\in_array(($memberData['location'] ?? null), ['header', 'headers'])) {
+                continue;
             }
 
-            return "\$this->$rootObjectName = \$this->xmlValueOrNull(\$data->$rootObjectName, '$type');\n";
+            $properties[] = strtr('$this->PROPERTY_NAME = PROPERTY_ACCESSOR;', [
+                'PROPERTY_NAME' => $memberName,
+                'PROPERTY_ACCESSOR' => $this->parseXmlResponse('$data', $memberName, $memberData),
+            ]);
```
Nyholm: Nice =)

src/CodeGenerator/Generator/ApiGenerator.php
```diff
@@ -834,7 +911,7 @@ private function resultClassPopulateResult(string $operationName, string $classN
     {
         $shape = $this->definition->getShape($className);
 
-        // Parse headers
+        // Parse headers'
```
Nyholm: oops

src/Core/Result.php
```diff
@@ -129,6 +129,7 @@ protected function populateResult(ResponseInterface $response, ?HttpClientInterf
             case '\DateTimeInterface':
```
Nyholm: Is `\\DateTimeInterface` needed here too?

Nyholm: I dont think so

Nyholm:
I rebased it and merged it. 

Thank you

PR #109, opened 2020-02-06
Create SES::sendEmail

Nyholm:


Nyholm:
Hm, those missing classes was not generated..

src/Ses/Input/Body.php
```diff
@@ -0,0 +1,73 @@
...
+     * An object that represents the version of the message that is displayed in email clients that support HTML. HTML
+     * messages can include formatted text, hyperlinks, images, and more.
+     *
+     * @var Content|null
+     */
+    private $Html;
+
+    /**
+     * @param array{
+     *   Text?: \AsyncAws\Ses\Input\Text|array,
```
jderusse: The docblock on property says it's `Content`

Nyholm: Yeah. But the class was not generated.

jderusse: maybe the class does not exist and the issue is the doc lock here ?

Nyholm: Yeah, the doc block is wrong.

Nyholm:
Wow, you are quick. Thank you

PR #110, opened 2020-02-06
Refactor ApiGenerator

Nyholm:
This is a massive rewrite of the ApiGenerator. I've split the ApiGenerator up to three files: 
- OperationGenerator
- ResultGenerator
- GeneratorHelper

I also created some objects from our Shape and Operation. I based it on @jderusse idea:

```php
$operation = $definition->getOperation();
$shape = $operation->getOutput();
if ($shape instanceOf StructureShape) {
  echo count($shape->getMembers());
} elseif ($shape instanceOf ListShape) {
  echo get_class($shape->getMember());
} else {
  echo $shape->getType();
}
```

Only files in src/CodeGenerator was changed. So we know the result of my refactoring did not effect the output.


Nyholm:
Future work: One could of course use the `Shape` and `Operation` classes more. I've cheated by letting them implement `\ArrayAccess`

Nyholm:
Im done with the PR now. 

The final push extracted XmlParser form ResultGenerator

src/CodeGenerator/Generator/GeneratorHelper.php
```diff
@@ -12,6 +12,7 @@
  * Small methods that might be useful when generating code.
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ * @author Jérémy Derussé <jeremy@derusse.com>
```
jderusse: thanks :hugs:

Nyholm: We should remove all or slap your name all over the place.

PR #112, opened 2020-02-07
Add Makefile to initialise tests to make sure docker is running.

Nyholm:
After this PR, nothing stops me from actually writing a test =)

### DIFF
--- a/src/Core/StreamableBody.php
+++ b/src/Core/StreamableBody.php
@@ -52,14 +52,20 @@ class StreamableBody
 
     public function getContentAsResource()
     {
-        $resource = \fopen('php://temp', 'rw+');
-        foreach ($this->responseStream as $chunk) {
-            fwrite($resource, $chunk->getContent());
+        try {
+            $resource = \fopen('php://temp', 'rw+');
+            foreach ($this->responseStream as $chunk) {
+                fwrite($resource, $chunk->getContent());
+            }
+
+            // Rewind
+            \fseek($resource, 0, \SEEK_SET);
+
+            return $resource;
+        } catch (\Throwable $e) {
+            \fclose($resource);
+
+            throw $e;
         }
-
-        // Rewind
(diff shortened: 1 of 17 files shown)


src/S3/Makefile
```diff
@@ -0,0 +1,10 @@
+.EXPORT_ALL_VARIABLES:
+
+# Build the PHP runtimes
+initialize:
+	docker pull lphoward/fake-s3
+	docker start async_aws_s3 || docker run -d -p 4569:4569 --name async_aws_s3 lphoward/fake-s3
```
jderusse: `docker rm -fv` to clear previous test ?

Nyholm: I should probably add a "stop" method. But I want to "start all" so I can run my test one by one..

Nyholm:
The test fails because a test in S3 shows a bug in Core. 

That bug is fixed in this PR too.

src/Core/StreamableBody.php
```diff
@@ -43,24 +43,23 @@ public function getContentAsString(): string
     {
         $resource = $this->getContentAsResource();
 
-        return \stream_get_contents($resource);
+        try {
+            return \stream_get_contents($resource);
+        } finally {
+            \fclose($resource);
```
jderusse: good catch

Nyholm: Yeah. I guess it is a good idea to have tests after all =)

src/Core/StreamableBody.php
```diff
@@ -43,24 +43,23 @@ public function getContentAsString(): string
...
+        }
     }
 
     public function getContentAsResource()
     {
         $resource = \fopen('php://temp', 'rw+');
```
jderusse: shouldn't it be closed if method failed?

Nyholm: Yeah, Good

Issue #121, opened 2020-02-08
Sign request when body is resource


Nyholm:
Related PRs that might be helpful: #125 and #126

Nyholm:
This is fixed by #159

PR #125, opened 2020-02-08
Adding failing tests for S3Client::DeleteObjects

Nyholm:
This is related to #121 (I think). It is an authentication issue.


jderusse:
fixed by #157 

Thank you @Nyholm

PR #129, opened 2020-02-08
Adding tests for SES

Nyholm:

PR #130, opened 2020-02-08
Use localstack/localstack for testing

Nyholm:
The Localstack docker image includes most services. 
https://hub.docker.com/r/localstack/localstack

This includes #129.


Nyholm:
There are two test failures here that I cannot explain. One for S3 and one for SQS. 

I'll leave this for now.

Nyholm:
Using localstack is not a good idea. It does not always give the proper responses. 
I tried with SQS and saw that our existing integration tests fails

PR #138, opened 2020-02-11
Iterate results over pages

jderusse:
Full pagination supported

### DIFF
--- a/src/S3/Result/ListObjectsV2Output.php
+++ b/src/S3/Result/ListObjectsV2Output.php
@@ -3,6 +3,8 @@
 namespace AsyncAws\S3\Result;
 
 use AsyncAws\Core\Result;
+use AsyncAws\S3\Input\ListObjectsV2Request;
+use AsyncAws\S3\S3Client;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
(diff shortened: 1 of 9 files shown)


src/S3/Result/ListObjectsV2Output.php
```diff
@@ -82,16 +84,31 @@ class ListObjectsV2Output extends Result implements \IteratorAggregate
...
+        }
+        $input = $this->lastRequest;
+        if (!$input instanceof ListObjectsV2Request) {
+            throw new \InvalidArgumentException('missing last request injected in paginated result');
+        }
+        $page = $this;
         while (true) {
-            yield from $this->CommonPrefixes;
+            $page->initialize();
+            yield from $page->CommonPrefixes;
```
Nyholm: Im not sure if it is better or worse.. But one could do 
```php
yield from $page->getCommonPrefixes(true);
```

That would remove the need for `$page->initialize();`

src/S3/S3Client.php
```diff
@@ -324,7 +324,7 @@ public function listObjectsV2($input): ListObjectsV2Output
             $this->getEndpoint($input->requestUri(), $input->requestQuery())
         );
 
-        return new ListObjectsV2Output($response);
+        return new ListObjectsV2Output($response, $this->httpClient, $this, $input);
```
Nyholm: HttpClient is not really needed, is it?

Nyholm: Oh, yes. it is of course. It is used for the Steamable response. 

Sorry for the noise.

jderusse:
SQS's ListQueues has a configuration for pagination, but everything is returned in the first page (there is no "nextMarker" or thing like that)

src/S3/Result/ListObjectsV2Output.php
jderusse: good catch, I'll fix that.

jderusse:
Don't merge it, I'll had an integration test

jderusse:
RFR

pagination is not fully tested (currentPage vs LazyLoad) because not implemented in FakeS3 client, and we don't have (yet) endpoint that use paginated results...

PR #136, opened 2020-02-11
Use static closure

jderusse:

PR #143, opened 2020-02-13
Remove call to xmlValueOrNull

jderusse:
alternative to #136 

Improve performances by avoiding call to a method and most logic could be computed at build time.

### DIFF
--- a/src/CodeGenerator/Generator/GeneratorHelper.php
+++ b/src/CodeGenerator/Generator/GeneratorHelper.php
@@ -24,16 +24,19 @@ class GeneratorHelper
             case 'structure':
             case 'map':
                 return 'array';
-            case 'boolean':
-                return 'bool';
-            case 'integer':
-                return 'int';
-            case 'timestamp':
-                return '\DateTimeImmutable';
-            case 'string':
             case 'long':
+            case 'string':
             case 'blob':
                 return 'string';
+            case 'integer':
+                return 'int';
+            case 'float':
+            case 'double':
+                return 'float';
+            case 'boolean':
+                return 'bool';
+            case 'timestamp':
+                return '\DateTimeImmutable';
             default:
                 throw new \RuntimeException(sprintf('Type %s is not yet implemented', $parameterType));
         }
@@ -65,7 +68,7 @@ class GeneratorHelper
         return $output;
     }
 
-    public static function addMethodComment(ServiceDefinition $definition, Shape $inputShape, string $baseNamespace, bool $allNullable = false): array
+    public static function addMethodComment(ServiceDefinition $definition, Shape $inputShape, string $baseNamespace, bool $allNullable = false, bool $inputSafe = false): array
     {
         $body = [];
         foreach ($inputShape['members'] as $name => $data) {
@@ -87,12 +90,20 @@ class GeneratorHelper
             } elseif ($data['streaming'] ?? false) {
                 $param = 'string|resource|\Closure';
             } elseif ('timestamp' === $param) {
-                $param = '\DateTimeInterface|string';
+                $param = $inputSafe ? '\DateTimeInterface' : '\DateTimeInterface|string';
             } else {
                 $param = self::toPhpType($param);
             }
 
-            $body[] = sprintf('  %s%s: %s,', $name, $allNullable || $nullable ? '?' : '', $param);
+            if ($allNullable || $nullable) {
+                if ($inputSafe) {
+                    $body[] = sprintf('  %s: %s,', $name, \strpos($param, '|') ? 'null|' . $param : '?' . $param);
+                } else {
+                    $body[] = sprintf('  %s?: %s,', $name, $param);
+                }
+            } else {
+                $body[] = sprintf('  %s: %s,', $name, $param);
+            }
         }
 
         return $body;
--- a/src/CodeGenerator/Generator/XmlParser.php
+++ b/src/CodeGenerator/Generator/XmlParser.php
@@ -73,12 +73,19 @@ class XmlParser
             case 'map':
                 return $this->parseXmlResponseMap($shape, $input);
             case 'string':
-            case 'boolean':
-            case 'integer':
             case 'long':
-            case 'timestamp':
+                return $this->parseXmlResponseString($shape, $input);
+            case 'integer':
+                return $this->parseXmlResponseInteger($shape, $input);
+            case 'float':
+            case 'double':
+                return $this->parseXmlResponseFloat($shape, $input);
+            case 'boolean':
+                return $this->parseXmlResponseBool($shape, $input);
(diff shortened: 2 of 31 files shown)


src/Core/Result.php
```diff
@@ -115,31 +115,4 @@ public function __construct(ResponseInterface $response, HttpClientInterface $ht
...
     }
-
-    final protected function xmlValueOrNull(?\SimpleXMLElement $xml, string $type)
-    {
-        if (null === $xml) {
-            return null;
-        }
-
-        $value = $xml->__toString();
-        if (empty($value) && 0 === $xml->count()) {
```
jderusse: I removed the `if (count($xlk) === 0) return null`, because I'm not sure this is true.
In official php sdk, this is not the case. for instance, in S3::ListObjects => the response contains `<Prefix></Prefix>` and php returns an empty string `""` see https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-s3-2006-03-01.html#listobjectsv2-example-1

Nyholm: Note that `$xml->foobar` will always return an SimpleXmlElement. No matter how the XML looks like. That is why we check if "foobar" exists. We do that with `$xml->foobar->count() === 0`

src/S3/Result/AwsObject.php
```diff
@@ -37,21 +37,21 @@ class AwsObject
...
      * @param array{
-     *   Key?: string,
-     *   LastModified?: \DateTimeInterface|string,
-     *   ETag?: string,
-     *   Size?: string,
-     *   StorageClass?: string,
-     *   Owner?: \AsyncAws\S3\Result\Owner|array,
+     *   Key: ?string,
+     *   LastModified: ?\DateTimeInterface,
+     *   ETag: ?string,
```
Nyholm: Im not sure this is true. We are not sure that all keys always exists.

Nyholm:
I like this approach. I did a quick review. I left a few comments.

src/Core/Result.php
jderusse: yes but casted to book it returns false.

```
$xml->exist ? 'true' : 'false'; // true
$xml->doesNot ? 'teue' : 'false'; // false
```

I'm AFK, but I'll provide a 3v4l + open an issue on phpstan for that

src/S3/Result/AwsObject.php
jderusse: this class is used for output/result. this class is called from the a code generated all keys are hard coded and should exist (value can be null, but all keys exists). If it's not the case, psalm will yield.

Nyholm: Okey. 

As long as we know that the response from AWS not always contain all keys.

jderusse: whether or not the AWS reponse contains all keys, this class is instanciated in with all key
in https://github.com/async-aws/aws/blob/4c7eb2d7a772c6f0070d88f357f793fab28edc68/src/S3/Result/ListObjectsV2Output.php#L214-L222

src/Core/Result.php
jderusse: Added a test to cover that case

jderusse: Bug opened on phpstan (to remove the rule) see https://github.com/phpstan/phpstan/issues/2997

src/Core/Tests/ResultTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+        self::assertTrue($data->Int ? true : false);
+        self::assertSame(1337, (int) (string) $data->Int);
+        self::assertTrue($data->EmptyInt ? true : false);
+        self::assertSame(0, (int) (string) $data->EmptyInt);
+
+        self::assertTrue($data->BoolTrue ? true : false);
+        self::assertSame(true, 'true' === (string) $data->BoolTrue);
+        self::assertTrue($data->BoolFalse ? true : false);
+        self::assertSame(false, 'true' === (string) $data->BoolFalse);
```
Nyholm: assertFalse

src/Core/Tests/ResultTest.php
```diff
@@ -0,0 +1,47 @@
...
+        self::assertTrue($data->EmptyString ? true : false);
+        self::assertSame('', (string) $data->EmptyString);
+
+        self::assertTrue($data->Int ? true : false);
+        self::assertSame(1337, (int) (string) $data->Int);
+        self::assertTrue($data->EmptyInt ? true : false);
+        self::assertSame(0, (int) (string) $data->EmptyInt);
+
+        self::assertTrue($data->BoolTrue ? true : false);
+        self::assertSame(true, 'true' === (string) $data->BoolTrue);
```
Nyholm: assertTrue

src/Core/Tests/ResultTest.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Core\Tests;
+
+use PHPUnit\Framework\TestCase;
+
+class ResultTest extends TestCase
```
Nyholm: It seams like this test is testing SimpleXmlElement. 

We could make a comment why we need this test here

src/S3/Result/AwsObject.php
Nyholm: I saw that now when doing a second review. 

So the `*Output` classes will do all the checking for null. 
:+1:

src/CodeGenerator/Generator/GeneratorHelper.php
```diff
@@ -27,11 +27,14 @@ public static function toPhpType(string $parameterType): string
             case 'boolean':
                 return 'bool';
             case 'integer':
+            case 'long':
```
Nyholm: Are you sure you can do this? What if a number larger than INT_MAX is passed?

src/CodeGenerator/Generator/XmlParser.php
```diff
@@ -73,12 +73,21 @@ private function parseXmlResponse(string $currentInput, ?string $memberName, arr
             case 'map':
                 return $this->parseXmlResponseMap($shape, $input);
             case 'string':
-            case 'boolean':
+            case 'character':
+            case 'byte':
+                return $this->parseXmlResponseString($shape, $input);
             case 'integer':
             case 'long':
```
Nyholm: Same here

src/CodeGenerator/Generator/GeneratorHelper.php
jderusse: You're right. The official SDK use "int" for documentation, but cast to string.

I reverted that part

src/Core/Tests/ResultTest.php
jderusse: added a comment

PR #148, opened 2020-02-16
Use a timeout when calling resolve

jderusse:
This PR leverage full async of reponse, to periodically check if the response is resolved.

This would be a requirement to use async-aws in symfony/messenger

### DIFF
--- a/src/Core/Result.php
+++ b/src/Core/Result.php
@@ -35,11 +35,11 @@ class Result
     private $response;
 
     /**
-     * @var HttpClientInterface|null
+     * @var HttpClientInterface
      */
     private $httpClient;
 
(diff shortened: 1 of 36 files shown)


src/Core/Result.php
```diff
@@ -38,15 +38,22 @@ public function __construct(ResponseInterface $response, HttpClientInterface $ht
     /**
      * Make sure the actual request is executed.
      *
+     * @param float|null timeout Duration in seconds before abording. When null wait until the end of execution.
```
Nyholm:
```suggestion
     * @param float|null timeout Duration in seconds before aborting. When null wait until the end of execution.
```

Issue #145, opened 2020-02-16
Create a Symfony Bundle


jderusse:
provides:
- semantic configuration (covers client's Configuration options)
- autowiring

Nyholm:
A symfony bridge like a bundle?

jderusse:
yes indeed, title updated

Nyholm:
Suggestion: 

The configuration could look like this: 

```yaml
async_aws:
  default: # Will be merged with other configuration
    config:
      region: eu-central-1
  sqs:
    config:
        region: us-west-1
  sns: ~ # This will complain if we dont have that package installed
```

I can get a service called `AsyncAws\S3\S3Client`. I've never configured it, but it used the default config.

----------

I dont think we should support multiple types of `S3Clients`. You have to register your own services if you want more than one.

PR #150, opened 2020-02-16
Added a small Symfony Bundle

Nyholm:
This is related to #145 

I've have not tested it (at all). I will have to add some tests too. 

Im happy with merging now but wait with creating a package and repo

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,9 +17,14 @@
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.16",
+        "matthiasnoback/symfony-config-test": "^4.1",
         "nette/php-generator": "^3.3",
         "nette/utils": "^3.0",
+        "nyholm/symfony-bundle-test": "^1.6",
+        "symfony/config": "^4.4 || ^5.0",
         "symfony/console": "^5.0",
+        "symfony/dependency-injection": "^4.4 || ^5.0",
+        "symfony/http-kernel": "^4.4 || ^5.0",
         "symfony/phpunit-bridge": "^5.0"
     },
     "autoload": {
--- a/src/Symfony/DependencyInjection/AsyncAwsExtension.php
+++ b/src/Symfony/DependencyInjection/AsyncAwsExtension.php
@@ -1,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Symfony\DependencyInjection;
+
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+
+class AsyncAwsExtension extends Extension
+{
+    public function load(array $configs, ContainerBuilder $container)
+    {
+        $configuration = new Configuration();
+        $config = $this->processConfiguration($configuration, $configs);
+
+        $usedServices = $this->registerConfiguredServices($container, $config);
+        $usedServices = $this->registerInstalledServices($container, $config, $usedServices);
+        $this->autowireServices($container, $usedServices);
+    }
+
+    private function registerConfiguredServices(ContainerBuilder $container, array $config): array
+    {
+        $availableServices = AwsPackagesProvider::getAllServices();
+        $usedServices = [];
+        $defaultConfig = $config;
+        unset($defaultConfig['services']);
+
+        foreach ($config['services'] as $name => $data) {
+            $client = $availableServices[$data['type']]['client'];
+            if (!class_exists($client)) {
+                throw new InvalidConfigurationException(sprintf('You have configured "async_aws.%s" but the "%s" package is not installed. Try running "composer require %s"', $name, $name, $availableServices[$name]['package']));
+            }
+
+            $config = array_merge($defaultConfig, $data);
+            if ($config['register_service']) {
+                $usedServices[$name] = $client;
+                $this->addServiceDefinition($container, $name, $config, $client);
+            } else {
+                $usedServices[$name] = null;
+            }
+        }
+
+        return $usedServices;
+    }
+
+    private function registerInstalledServices(ContainerBuilder $container, array $config, array $usedServices): array
+    {
+        if (!$config['register_service']) {
+            return $usedServices;
+        }
+
+        unset($config['services']);
+        $availableServices = AwsPackagesProvider::getAllServices();
+        foreach ($availableServices as $name => $data) {
+            if (\array_key_exists($name, $usedServices)) {
+                continue;
+            }
+
+            $client = $data['client'];
+            if (!class_exists($client)) {
+                continue;
+            }
+
+            $usedServices[$name] = $client;
+            $this->addServiceDefinition($container, $name, $config, $client);
+        }
+
+        return $usedServices;
+    }
+
+    private function addServiceDefinition(ContainerBuilder $container, string $name, array $config, string $clientClass): void
+    {
+        if (\array_key_exists('logger', $config)) {
+            $logger = $config['logger'] ? new Reference($config['logger']) : null;
+        } else {
+            // Use default Symfony logger unless explicitly set to null.
+            $logger = new Reference('logger', ContainerInterface::NULL_ON_INVALID_REFERENCE);
+        }
+
+        $definition = new Definition($clientClass);
+        $definition->addArgument($config['config']);
+        $definition->addArgument(isset($config['credential_provider']) ? new Reference($config['credential_provider']) : null);
+        $definition->addArgument(isset($config['http_client']) ? new Reference($config['http_client']) : null);
+        $definition->addArgument($logger);
+        $container->setDefinition(sprintf('async_aws.service.%s', $name), $definition);
+    }
+
+    private function autowireServices(ContainerBuilder $container, array $usedServices): void
+    {
+        $awsServices = AwsPackagesProvider::getAllServices();
+        foreach ($usedServices as $name => $client) {
+            if (null === $client) {
+                // This client is disabled.
+                continue;
+            }
+
+            $serviceId = sprintf('async_aws.service.%s', $name);
+            if (isset($awsServices[$name])) {
+                $container->setAlias($client, $serviceId);
+
+                continue;
+            }
+
+            // Assert: Custom name
+            $container->registerAliasForArgument($serviceId, $client, $name);
+        }
+    }
+}
--- a/src/Symfony/DependencyInjection/AwsPackagesProvider.php
+++ b/src/Symfony/DependencyInjection/AwsPackagesProvider.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Symfony\DependencyInjection;
+
+class AwsPackagesProvider
+{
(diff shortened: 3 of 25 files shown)


composer.json
```diff
@@ -10,8 +10,11 @@
         "ext-hash": "*",
         "ext-json": "*",
         "psr/log": "~1.0",
+        "symfony/config": "^4.4 || ^5.0",
```
jderusse: should be in require-dev on root package

Nyholm: Thank you

src/Symfony/README.md
```diff
@@ -0,0 +1,25 @@
...
+## Configure
+
+```yaml
+async_aws:
+  default: # Will be merged with other configuration
+    config:
+      region: eu-central-1
+  sqs:
+    config:
+        region: us-west-1
```
jderusse: indentation

src/Symfony/composer.json
```diff
@@ -0,0 +1,31 @@
...
+    "keywords": [
+        "aws",
+        "symfony",
+        "bundle"
+    ],
+    "license": "MIT",
+    "require": {
+        "php": "^7.2.5",
+        "symfony/config": "^4.4 || ^5.0",
+        "symfony/dependency-injection": "^4.4 || ^5.0",
```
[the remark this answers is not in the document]
jderusse: If we use a class defined in core, we'll add an entry in "require" section.
If we don't, then I don't get why we have to define a supported version.

src/Symfony/README.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+## Configure
+
+```yaml
+async_aws:
+  default: # Will be merged with other configuration
+    config:
+      region: eu-central-1
+  sqs:
```
jderusse: There an issue with this config structure. I can't declare 2 clients with differents configurations. Imagine, I want to copy files from a bucket to another that requires differents credentials.

What's about:
```yaml
async_aws:
  default: 
    region: eu-central-1
  foo:
    region: us-east-1
```

That generate the following Binds (check if class Exists or not)
```phpp
\AsyncAws\S3\S3Client // region: eu-central-1
\AsyncAws\S3\S3Client $fooClient // region: us-east-1
```

Drawback would be, the user won't have nice exception telling "please install composer req async-aws/sqs", that could be fixed with (but I wonder if it whroth it):

```yaml
async_aws:
  default: 
    region: eu-central-1
    services: [s3, sqs]
```

src/Symfony/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,50 @@
...
+
+        $node = $rootNode->children();
+
+        if ('default' !== $name) {
+            $node->booleanNode('register_service')->info('If set to false, no service will be created.')->defaultTrue()->end();
+        }
+
+        $node
+            ->arrayNode('config')->arrayPrototype()->end()
+            ->scalarNode('credential_provider')->info('A service name for AsyncAws\Core\Credentials\CredentialProvider.')->defaultNull()->end()
```
jderusse: I'm not sure we need such configuration. Couldn't they decorates the bundle's internal services instead?

src/Symfony/DependencyInjection/AsyncAwsExtension.php
```diff
@@ -0,0 +1,67 @@
...
+        unset($config['default']);
+
+        foreach ($config as $name => $data) {
+            $client = $availableServices[$name]['client'];
+            if (!class_exists($client)) {
+                throw new InvalidConfigurationException(sprintf('You have configured "async_aws.%s" but the "%s" package is not installed. Try running "composer require %s"', $name, $name, $availableServices[$name]['package']));
+            }
+            unset($availableServices[$name]);
+
+            $config = array_merge($defaultConfig, $data);
```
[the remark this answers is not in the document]
jderusse: Looks OK to me

src/Symfony/README.md
```diff
@@ -0,0 +1,25 @@
+# Async AWS Symfony Bundle
+
+![](https://github.com/async-aws/symfony-bundle/workflows/Tests/badge.svg?branch=master)
+![](https://github.com/async-aws/symfony-bundle/workflows/BC%20Check/badge.svg?branch=master)
+
+An small SymfonyBundle that helps with configuration and autowiring.
```
jderusse: `A small`

src/Symfony/composer.json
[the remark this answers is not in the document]
Nyholm: The bundle does not require anything. But if the AbstractClient changes the constructor param, the bundle will break.

src/Symfony/README.md
Nyholm: But you cannot autowire for those two S3 clients...

But sure. May I suggest to use the “foo” key as service. Ie. If you write async_aws.foo you need async_aws.foo.type: s3. 

You could do just:

async_aws.s3: ~

jderusse: > But you cannot autowire for those two S3 clients

What's about "Autowiring by type and name" https://symfony.com/blog/new-in-symfony-4-2-autowiring-by-type-and-name

When using:
```php
function __construct(S3Client $foo) // $foo because I defined a key `foo:` in config
{
}
```
Like Symfony FW does for Cache adapters and Lock.


> If you write async_aws.foo you need async_aws.foo.type: s3.

Thats could be optionnal: When type is not provided, we can provide a binding for all existing clients

src/Symfony/composer.json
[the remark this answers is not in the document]
Nyholm: I required core since we are using the constuctor

src/Symfony/README.md
Nyholm: This is fixed. See Readme

src/Symfony/DependencyInjection/Configuration.php
Nyholm: Na. 
Currently the library decides what should happen if nothing is configured.

Nyholm:
Wohoo. This PR is done. I've added a bunch of tests. =)

Nyholm:
40 min later: Im really done now. 

Enjoy reviewing

src/Symfony/README.md
```diff
@@ -0,0 +1,43 @@
...
+    config: # Will be merged with other configuration
+        region: eu-central-1
+
+    services:
+        ses: ~ # This will complain if we dont have that package installed
+        sqs:
+            config:
+                region: us-west-1
+        foobar:
+            type: sqs
```
Nyholm: Are we happy with this config and autowiring?

Nyholm:
Btw, Do you have any suggestion/comments on the Bundle name, package name or the repository name?

src/Symfony/DependencyInjection/AwsPackagesProvider.php
```diff
@@ -0,0 +1,37 @@
...
+                'package' => 'async-aws/ses',
+            ],
+            'sqs' => [
+                'client' => \AsyncAws\Sqs\SqsClient::class,
+                'package' => 'async-aws/sqs',
+            ],
+            'sns' => [
+                'client' => \AsyncAws\Sns\SnsClient::class,
+                'package' => 'async-aws/sns',
+            ],
```
jderusse: What's about STS?

src/Symfony/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,67 @@
...
+{
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder('async_aws');
+        /** @var ArrayNodeDefinition $rootNode */
+        $rootNode = $treeBuilder->getRootNode();
+        $rootNode
+            ->fixXmlConfig('service')
+            ->children()
+                ->booleanNode('register_service')->info('If set to false, no service will be created.')->defaultTrue()->end()
```
jderusse: What's the purpose fot this? 

If people don't wan't services for a client, they just have to not Inject it? Autowire will remove not-used services at the end

src/Symfony/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,67 @@
...
+        return $treeBuilder;
+    }
+
+    private static function validateType(array $services)
+    {
+        $awsServices = AwsPackagesProvider::getServiceNames();
+        foreach ($services as $name => $config) {
+            if (\in_array($name, $awsServices)) {
+                if (isset($config['type']) && $name !== $config['type']) {
+                    throw new InvalidConfigurationException(sprintf('You cannot define a service named "%s" with type "%s". That is super confusing.', $name, $config['type']));
```
jderusse: good catch :)

src/Symfony/DependencyInjection/AsyncAwsExtension.php
```diff
@@ -0,0 +1,113 @@
...
+
+            $serviceId = sprintf('async_aws.service.%s', $name);
+            if (isset($awsServices[$name])) {
+                $container->setAlias($client, $serviceId);
+
+                continue;
+            }
+
+            // Assert: Custom name
+            $container->registerAliasForArgument($serviceId, $client, $name);
```
jderusse: What's about also adding `$container->registerAliasForArgument($serviceId, $client)` for the default service (`sqs`, `s3`, ..).

It would allow out of the box
```shell
$ composer req async-aws/sqs
```
```yaml
async_aws:
    config:
        region: eu-central-1
```

```php
public function __construct(SqsClient $myLittlerClient) {
}
```

src/Symfony/composer.json
```diff
@@ -0,0 +1,40 @@
+{
+    "name": "async-aws/symfony-bundle",
```
jderusse: what's about `async-aws/aws-bundle` or `async-aws/async-aws-bundle`
For consistency with most of bundles https://packagist.org/search/?q=bundle

`doctrine-bundle`, `twig-bundle`, `flysystem-bundle`, `redis-bundle`, `cords-bundle`, ...

src/Symfony/DependencyInjection/AsyncAwsExtension.php
Nyholm: Isn’t that already the case? I’m sure it is. That is at least the intention.

src/Symfony/DependencyInjection/AwsPackagesProvider.php
Nyholm: I forgot about it :)

src/Symfony/DependencyInjection/Configuration.php
Nyholm: If someone wants to register a service without help of the bundle. They can disable the bundle's service to avoid conflicts with autowiring.

src/Symfony/composer.json
Nyholm: Lets do async-aws/async-aws-bundle

src/Symfony/DependencyInjection/AsyncAwsExtension.php
jderusse: Oh nice !!.. I thougth that the last argument of `registerAliasForArgument` was the name of the property to bind..

src/Symfony/README.md
jderusse: Nice !

src/Symfony/composer.json
```diff
@@ -0,0 +1,40 @@
...
+    },
+    "autoload": {
+        "psr-4": {
+            "AsyncAws\\Symfony\\": ""
+        },
+        "exclude-from-classmap": [
+            "/Tests/"
+        ]
+    },
+    "minimum-stability": "dev"
```
jderusse: not needed. Is it?

Nyholm: Currently it is since we don’t have any releases.

PR #152, opened 2020-02-17
Preload next page

jderusse:
Small optimisation on pagination to preload next page yielding items from current page

### DIFF
--- a/src/CodeGenerator/Generator/ResultGenerator.php
+++ b/src/CodeGenerator/Generator/ResultGenerator.php
@@ -185,6 +185,16 @@ class ResultGenerator
             ->addComment("@return \Traversable<$iteratorType>")
             ->setBody($this->generateOutputPaginationLoader($iteratorBody, $pagination, $namespace, $baseNamespace, $service, $operation))
         ;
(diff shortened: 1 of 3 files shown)


Nyholm:
To be clear, as I can see it is only an optimization if:
- you look at the first element in the list
- Then make a second request with that same HttpClient (any other http request)
- you wait for the result
- now it is likely that we’ve fetched part of the next pagination page. 

If I don’t make a second request, this will don’t give any benefit nor performance penalty.

jderusse:
hmm. maybe I misunderstood how http client...

my idea was : to start prefetching next page when you start reading in the current page.
while the caller is processing items of page N, http is loading page N+1

The drawback is: if you stop reading in the middle, we preload z page for nothing.

Nyholm:
> The drawback is: if you stop reading in the middle, we preload z page for nothing.

That is true only if you waited on a second request. 

The HttpClient only fetches data from the server when we call resolve(). When the client fetches data it fetches data on all responses at the same time. 

So, if the user fetches the first item in the list and nothing more, we just create one response object that we didn’t use. The API server will never know about it.

PR #153, opened 2020-02-18
Implement waiters - sync

jderusse:
Regarding the User Experience. I hesitate between

Current Impl:
```php
$sqsClient->queueExists(['QueueName' => 'fooBar']); // bool
$sqsClient->waitForQueueExists(['QueueName' => 'fooBar'], $delay, $maxAttempts); // void
```

and Idea
```php
$checker = $sqsClient->queueExists(['QueueName' => 'fooBar']); // Checker
$checker->isSuccessfull(); // bool
$checker->isFailure(); // bool
$checker->wait($delay, $maxAttempts); // void: wait  untill either success, failure, or timeout
$checker->resolve(); // call to internal operation is async itself, which is not possible with my current impl
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -91,6 +91,20 @@ foreach($result as $file) {
 }
 ```
 
+## Waiter
+
+Similar to Official AWS PHP SDK, Async-Aws provides waiters to let you wait
+until an long operation finished.
+
+```php
+// create a queue Async and don't wait for the response.
+$sqsClient->createQueue(['QueueName' => 'fooBar']);
+
+$sqsClient->waitForQueueExists(['QueueName' => 'fooBar']);
+```
+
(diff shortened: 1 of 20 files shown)


Nyholm:
I like your idea. It seams cleaner and friendlier.

But I don’t understand the comment about calling resolve()

jderusse:
waiters internally called a method. For instance the waiter `QueueExists` called `GetQueueUrl` and check is the response is 200 or 404.
With my impl, the call to `QueueExists` is synchronous. If we want to be full Async, I should replace my `return bool` by an Object that provide a `resolve` method...

Nyholm:
Fully async FTW!

jderusse:
closed in favor of #156

PR #156, opened 2020-02-19
Implement waiters - async

jderusse:
Aleternative to #153 that return a full async object instead of blocking resolution.

This PR adds an `WaiterResult extends Result` that contains most of business logic:
PRO:
- user still get a `Result` when calling the client

CON:
- had to change set `$response` to protected
- if user calls `->resolve()` they could have an exception (for instance, 404 not found) which is not a error for the waiter (the method `WaiterResult::wait())` catch such exception and return weither or not it's success or a failure.

Tell me which PR you prefer.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -91,6 +91,23 @@ foreach($result as $file) {
 }
 ```
 
+## Waiter
+
+Similar to Official AWS PHP SDK, Async-Aws provides waiters to let you wait
+until an long operation finished.
+
+```php
+// create a queue Async and don't wait for the response.
+$sqsClient->createQueue(['QueueName' => 'fooBar']);
+
+$waiter = $sqsClient->queueExists(['QueueName' => 'fooBar']);
+echo $waiter->isSuccess(); // false
+$waiter->wait();
+echo $waiter->isSuccess(); // true
+```
+
+[more information about waiters and hasers...](./docs/waiter.md)
+
 ## Organization
 
 | Repository | Namespace | Package name |
--- a/docs/waiter.md
+++ b/docs/waiter.md
@@ -1,0 +1,43 @@
+# Waiter and Haser
+
+Similar to Official AWS PHP SDK, Async-Aws provides waiters to let you wait
+until an long operation finished.
+
+To ease the development, the above code is wrapped in a helper called "waiter".
+
+```php
+// create a queue Async and don't wait for the response.
+$sqsClient->createQueue(['QueueName' => 'fooBar']);
+
+$waiter = $sqsClient->queueExists(['QueueName' => 'fooBar']);
+
+echo $waiter->isSuccess(); // false
+
+$waiter->wait();
+
+echo $waiter->isSuccess(); // true
+```
+
+A waiter provides methods that let you check the status of the operation.
+* `isSuccess()` returns true when operation is successful
+* `isFailure()` returns true when operation failed
+* `isPending()` returns true when the state of the operation is not yet determinate
+* `getState()` returns the state of the operation either `sucess`, `failure` or `pending`
+* `wait($timeout, $delay)` waits until the state of the operation is determinate.
+
+As usual, Async-Aws is async and not blocking by default:
+
+```php
+$waiter = $sqsClient->queueExists(['QueueName' => 'fooBar']);
+while(true) {
+    if ($waiter->wait(0)) {
+        // When method `wait` returns true, the state is resolved.
+        break;
+    }
+
+    // perform business logic
+    sleep(1);
+}
+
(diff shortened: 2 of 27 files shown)


Readme.md
```diff
@@ -91,6 +91,23 @@ foreach($result as $file) {
...
+until an long operation finished.
+
+```php
+// create a queue Async and don't wait for the response.
+$sqsClient->createQueue(['QueueName' => 'fooBar']);
+
+$waiter = $sqsClient->queueExists(['QueueName' => 'fooBar']);
+echo $waiter->isSucess(); // false
+$waiter->wait();
+echo $waiter->isSucess(); // true
```
Nyholm: Typo. isSuccess

Readme.md
```diff
@@ -91,6 +91,23 @@ foreach($result as $file) {
...
+
+Similar to Official AWS PHP SDK, Async-Aws provides waiters to let you wait
+until an long operation finished.
+
+```php
+// create a queue Async and don't wait for the response.
+$sqsClient->createQueue(['QueueName' => 'fooBar']);
+
+$waiter = $sqsClient->queueExists(['QueueName' => 'fooBar']);
+echo $waiter->isSucess(); // false
```
Nyholm: Same

docs/waiter.md
```diff
@@ -0,0 +1,43 @@
...
+
+To ease the development, the above code is wrapped in a helper called "waiter".
+
+```php
+// create a queue Async and don't wait for the response.
+$sqsClient->createQueue(['QueueName' => 'fooBar']);
+
+$waiter = $sqsClient->queueExists(['QueueName' => 'fooBar']);
+
+echo $waiter->isSucess(); // false
```
Nyholm: Same

docs/waiter.md
```diff
@@ -0,0 +1,43 @@
...
+
+$waiter->wait();
+
+echo $waiter->isSucess(); // true
```
Nyholm: Same

Nyholm:
Do we really need to extend Result? My thought is not complete but I still want to share

jderusse:
no, we do 't need to extends Result, it's the opposite: it adds complexity.

But I thought that "everything returns a Result" was a good thing...

I.suggedt to use a. Interface instead. what do you think?

Nyholm:
I think you can remove the “cons” by not extending Result. 

If we introduce a Result interface, what would be the shared methods between a Result and a Waiter?

Readme.md
jderusse: thanks

jderusse:
Removed `extends Result`.

Indeed, I thougth about adding an interface that provides `resolve`, `cancel` and `info` methods. But I'm not sure it whorth it. 

Let's keep theme separate, we'll add such interface if needed.

PR RFR

src/Core/Waiter.php
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * @var AbstractApi|null
+     */
+    protected $awsClient;
+
+    /**
+     * @var array|object|null
+     */
+    protected $request;
```
Nyholm: Only a minor thing. Rename this to “input” and add some description. 

Or maybe I’m just confused, but I thought this was a request object...

Nyholm: It will never be an array BTW.

src/Core/Waiter.php
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        if (!isset($this->response)) {
+            $this->stealResponse($this->refreshState());
+        }
+
+        $exception = null;
+        $this->resolve();
+
+        \assert(null !== $this->response);
```
Nyholm: We never unset the response in Resolve(). $this->response will always have a value, right?

jderusse: yes. your right.

But psalm yield because `$this->response` can be null. I hesitate between ignoring the psalm error and adding this `assert`.

Nyholm: I think psalm complained because the property is nullable.

src/Core/Waiter.php
jderusse: Fixed. Renamed Result's request too for consistency

src/Core/Waiter.php
jderusse: yes exactly. The property is nullable, and psalm yield. 
But as you sayed, in the current workflow, nothing unset the property and it can't be null.

Should I ignore the error reported by psalm? Or add a useless `assert`?

jderusse: hmm. removed and nobody complains anymore... weird.

PR #157, opened 2020-02-19
Fix DeleteObjects

jderusse:
while fixing #125 I discovered many issues:
- the generated XML were wrong (I don't know if the bug has been introduced during my refactor)
- the signature was wrong (because of wrong content type)
- fake S3 just don't impement DeleteObjects 

This PR:
- refactor the way XML are generated. XML code is now generated at built time  (in the same way with parse XML from output)
- fix signature
- mark test "DeleteObjects" skiped (tested localy with my private Key)


src/S3/Input/PutObjectAclRequest.php
```diff
@@ -196,14 +196,59 @@ public function getVersionId(): ?string
...
+        $root->setAttribute('xmlns', 'http://s3.amazonaws.com/doc/2006-03-01/');
+
+        if (null === $input = $this->AccessControlPolicy) {
+            return $document->saveXML();
+        }
+
+        if (null !== $input_Grants = $input->getGrants()) {
+            foreach ($input_Grants as $input_Grants_item) {
+                $root->appendChild($root_AccessControlList = $document->createElement('AccessControlList'));
+                $input_Grants_item_AccessControlList = $input_Grants_item;
```
jderusse: mhh.. Something is missing. It should generates:
```xml
<AccessControlPolicy xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <AccessControlList>
      <Grant>
         <Grantee>
            <DisplayName>string</DisplayName>
            <EmailAddress>string</EmailAddress>
            <ID>string</ID>
            <xsi:type>string</xsi:type>
            <URI>string</URI>
         </Grantee>
         <Permission>string</Permission>
      </Grant>
   </AccessControlList>
   <Owner>
      <DisplayName>string</DisplayName>
      <ID>string</ID>
   </Owner>
</AccessControlPolicy>
```

 (cf https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectAcl.html)

jderusse: fixed

Nyholm: Cool I did know what was wrong =)

PR #159, opened 2020-02-20
Upload resource

jderusse:
This PR fix signature issue when the input contains a resource.

Basicaly it convert the resource into a string. Which is not ideal, but it looks like that's what the official SDK does...


Nyholm:
Oh, sorry. I merged then in the wrong order.

PR #163, opened 2020-02-20
Fix deep input

jderusse:
When a request contains sub-objects the return requestBody was wrong.

Added a test: src/Sqs/Tests/Unit/Input/SendMessageRequestTest.php to assert output

### DIFF
--- a/src/Ses/Input/SendEmailRequest.php
+++ b/src/Ses/Input/SendEmailRequest.php
@@ -127,26 +127,104 @@ class SendEmailRequest
     public function requestBody(): array
     {
         $payload = ['Action' => 'SendEmail', 'Version' => '2019-09-27'];
-        if (null !== $this->FromEmailAddress) {
-            $payload['FromEmailAddress'] = $this->FromEmailAddress;
+        $indices = new \stdClass();
+        if (null !== $v = $this->FromEmailAddress) {
+            $payload['FromEmailAddress'] = $v;
         }
-        if (null !== $this->Destination) {
-            $payload['Destination'] = $this->Destination;
+
+        (static function ($input) use (&$payload, $indices) {
+            (static function ($input) use (&$payload, $indices) {
+                $indices->kb596653 = 0;
+                foreach ($input as $value) {
+                    ++$indices->kb596653;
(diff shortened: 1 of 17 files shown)


src/Ses/Input/SendEmailRequest.php
```diff
@@ -127,26 +127,104 @@ public function getReplyToAddresses(): array
...
+            foreach ($input as $value) {
+                ++$indices->k7c3bed1;
+                (static function ($input) use (&$payload, $indices) {
+                    $payload["EmailTags.{$indices->k7c3bed1}.Name"] = $input->getName();
+                    $payload["EmailTags.{$indices->k7c3bed1}.Value"] = $input->getValue();
+                })($value);
+            }
+        })($this->EmailTags);
+
+        if (null !== $v = $this->ConfigurationSetName) {
```
Nyholm: Oh, This code is not super pretty.

src/Sqs/Input/CreateQueueRequest.php
```diff
@@ -69,15 +69,24 @@ public function gettags(): array
...
-        if (null !== $this->QueueName) {
-            $payload['QueueName'] = $this->QueueName;
-        }
-        if (null !== $this->Attributes) {
-            $payload['Attribute'] = $this->Attributes;
-        }
-        if (null !== $this->tags) {
-            $payload['Tag'] = $this->tags;
-        }
+        $indices = (object) [];
```
Nyholm: What about? 
`$indices = new \stdObject()`

Isnt this the same?

src/Ses/Input/SendEmailRequest.php
jderusse: Agreed :/

I'm open to any suggestion

src/Sqs/Input/CreateQueueRequest.php
jderusse: fixed

src/Ses/Input/SendEmailRequest.php
Nyholm: I don’t have any...
Let’s merge and improve this later.

PR #170, opened 2020-02-21
Adding Lambda and SNS

Nyholm:
These are needed for some of my upstream projects.


src/CodeGenerator/Generator/XmlDumper.php
```diff
@@ -94,6 +94,7 @@ private function dumpXmlShape(Member $member, Shape $shape, string $output, stri
         }
 
         switch ($shape->getType()) {
+            case 'blob':
```
Nyholm: This the only change I made to the code generator

PR #173, opened 2020-02-21
Make sure we dont throw exception twice

### LINKED ISSUE #171: Dont throw exception twice
If an exception is thrown on `->resolve()`, it can be caught. But then on `__destruct()`, we will call resolve again and a new message is thrown.

Nyholm:
This will fix #171

### DIFF
--- a/src/Core/Result.php
+++ b/src/Core/Result.php
@@ -121,12 +121,12 @@ class Result
     {
         if (!isset($this->response)) {
             return [
-                'resolved' => true,
+                'resolved' => $this->isResolved,
             ];
         }
 
         return [
-            'resolved' => false,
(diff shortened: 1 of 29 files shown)


src/Core/Result.php
```diff
@@ -41,6 +41,14 @@ class Result
      */
     private $httpClient;
 
+    /**
+     * A Result can be resolved many times. This boolean is true if the result
+     * has been resolved at least once.
+     *
+     * @var bool
+     */
+    private $isResolved = false;
```
jderusse: could be returned in `info()` too

Nyholm: Good idea!

PR #175, opened 2020-02-22
Added test for getting an error message from RestJsonError

Nyholm:


composer.json
```diff
@@ -20,7 +20,7 @@
         "matthiasnoback/symfony-config-test": "^4.1",
         "nette/php-generator": "^3.3",
         "nette/utils": "^3.0",
-        "nyholm/symfony-bundle-test": "^1.6",
+        "nyholm/symfony-bundle-test": "^1.6.1",
```
Nyholm: This will make sure dep-lowest does not fail

src/Core/Test/Http/SimpleMockedResponse.php
```diff
@@ -48,14 +51,29 @@ public function cancel(): void
 
     public function getInfo(string $type = null)
     {
-        if (null === $type || 'response_headers' === $type) {
-            return [];
+        $info = [
+            'response_headers' => $this->getFlatHeaders(),
```
jderusse: if info is `http_code`, we compute that for nothing :-/

Nyholm: It is just tests ¯\_(ツ)_/¯

Nyholm: I also expect it to be more things to this array later.

Nyholm:
FYI: Im doing all these updates because "Lowest dependencies" should not fail anymore.

Issue #177, opened 2020-02-22
[S3] PutObjectAcl should not have a body when using canned ACL


Nyholm:
Looking at these examples: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectAcl.html#API_PutObjectAcl_Examples

If one is using the `x-amz-acl` header, then the request body should be empty. Currently we get an error because the body contains: 

```xml
<?xml version="1.0" encoding="UTF-8"?>
```

Nyholm:
Im not sure how we (from the definition)  are supposed to know this.. 

We may need to add some custom (non-generated) logic to some adapters. Ie, we do need the traits again.

PR #178, opened 2020-02-22
Flysystem v2 s3

Nyholm:
This adapter is working. But I cannot make sure it is working with Docker. (I have not spent too much time on it). 

The only bug I found was related to S3. Reported here: #177


Nyholm:
Im not super happy with the namespace.. could that be improved?

jderusse:
do we need to support v1?

If v3 is released would we create a new component or that component will support both version?

IMHO Flysystem\S3 works.

Nyholm:
Hm.. Im not sure. 

I would like to be able to support both versions.

Oh, of course. We can support all verisons of flysystem with two different classes.

jderusse:
> Oh, of course. We can support all verisons of flysystem with two different classes.

yes, that's what I mean in my second sentence

Nyholm:
So, everything is working except #177, but that is not blocker for merging. 

The only other thing that is broken are the tests =) I dont know why docker is not helping me at all.

src/Flysystem/S3/Tests/Integration/S3FilesystemV2Test.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    private $s3Client;
+
+    public static function setUpBeforeClass(): void
+    {
+        static::$adapterPrefix = 'travis-ci/' . bin2hex(random_bytes(10));
+
+        if (!getenv('FLYSYSTEM_AWS_S3_KEY')) {
+            // Fixme, Docker is not working
+            self::markTestSkipped('Docker image is not working.');
```
Nyholm: We dont need integration tests. They exist (or should exist) in the async-aws/s3 package.

Nyholm:
When I test with "real" s3 it all passes. I'll merge this PR and create a package

PR #179, opened 2020-02-22
Flysystem v1 s3

Nyholm:
I'll leave this here for now.


Nyholm:
I will merge this when the CI is green. I'll work on integration tests in the next PR. I will also make sure we test both versions.

PR #182, opened 2020-02-22
Adding CloudFormation

Nyholm:
This is also needed for an upstream project.

### DIFF
--- a/src/CloudFormation/Result/DescribeStacksOutput.php
+++ b/src/CloudFormation/Result/DescribeStacksOutput.php
@@ -131,7 +131,7 @@ class DescribeStacksOutput extends Result implements \IteratorAggregate
             foreach ($xml as $item) {
                 $items[] = new Stack([
                     'StackId' => ($v = $item->StackId) ? (string) $v : null,
-                    'StackName' => ($v = $item->StackName) ? (string) $v : null,
+                    'StackName' => (string) $item->StackName,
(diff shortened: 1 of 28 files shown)


src/CloudFormation/Result/DescribeStacksOutput.php
```diff
@@ -0,0 +1,232 @@
...
+                    'CreationTime' => ($v = $item->CreationTime) ? new \DateTimeImmutable((string) $v) : null,
+                    'DeletionTime' => ($v = $item->DeletionTime) ? new \DateTimeImmutable((string) $v) : null,
+                    'LastUpdatedTime' => ($v = $item->LastUpdatedTime) ? new \DateTimeImmutable((string) $v) : null,
+                    'RollbackConfiguration' => new RollbackConfiguration([
+                        'RollbackTriggers' => (function (\SimpleXMLElement $xml): array {
+                            $items = [];
+                            foreach ($xml as $item) {
+                                $items[] = new RollbackTrigger([
+                                    'Arn' => ($v = $item->Arn) ? (string) $v : null,
+                                    'Type' => ($v = $item->Type) ? (string) $v : null,
```
Nyholm: This is an issue with the code generator. RollbackTrigger must have two parameters (Arn and Type) and they must be string. Using them like this here allows them to be null. 

This is the error psalm is complaining about.

Nyholm: Fixed in #185

PR #184, opened 2020-02-22
Make sure AwsClient is useful

Nyholm:


src/Core/AbstractApi.php
```diff
@@ -167,7 +167,7 @@ private function stringifyBody($body): string
         return (string) $body;
     }
 
-    private function getSigner()
+    private function getSigner(?string $scope = null)
     {
         if (null === $this->signer) {
```
jderusse: the signer should vary with method's argument.

using the same AwsClient to perform 2 requests in 2 services will failed.

Nyholm:
Hm.. maybe `AwsClient` should not extend `AbstractClient`. That will remove the "feature" of being allowed to make requests with it. 

The `AwsClient` will then only work as a service storage.

src/Core/AwsClient.php
```diff
@@ -4,33 +4,93 @@
...
+        if ($configuration->isDefault(Configuration::OPTION_ENDPOINT)) {
+            $this->credentialProvider = $credentialProvider ?? new CacheProvider(new ChainProvider([
+                new ConfigurationProvider(),
+                new IniFileProvider($this->logger),
+                new InstanceProvider($this->httpClient, $this->logger),
+            ]));
+        } else {
+            $this->credentialProvider = $credentialProvider ?? new CacheProvider(new ChainProvider([
+                new ConfigurationProvider(),
+                new IniFileProvider($this->logger),
```
Nyholm: Could we refactor this constructor code somehow?

jderusse: fixed by #190

jderusse: remove the `else` part and just keep the `if`

Nyholm: Like this?

jderusse: :+1:

PR #185, opened 2020-02-22
Dont use null for required parameters

Nyholm:
This will fix (or try to fix) the issue spotted in #182

### DIFF
--- a/src/Sqs/Result/ReceiveMessageResult.php
+++ b/src/Sqs/Result/ReceiveMessageResult.php
@@ -76,7 +76,7 @@ class ReceiveMessageResult extends Result
 
                                     return $items;
                                 })($item->Value->BinaryListValue),
-                                'DataType' => ($v = $item->Value->DataType) ? (string) $v : null,
+                                'DataType' => (string) $item->Value->DataType,
                             ]);
                         }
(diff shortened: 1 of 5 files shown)


src/Sqs/Result/ReceiveMessageResult.php
```diff
@@ -76,7 +76,7 @@ protected function populateResult(ResponseInterface $response, HttpClientInterfa
 
                                     return $items;
                                 })($item->Value->BinaryListValue),
-                                'DataType' => ($v = $item->Value->DataType) ? (string) $v : null,
+                                'DataType' => (string) $item->Value->DataType,
```
Nyholm: This belongs to `MessageAttributeValue`. It says that the DataType property is required. But it can be null. 

Will this work?

Nyholm: No.. this will never be null

Nyholm: Sorry, I was wrong. 

The `Result\MessageAttributeValue::$dataType` is not nullable. I was looking at the input.

Nyholm:
Thank you

PR #188, opened 2020-02-22
Fixed nullable nested objects

Nyholm:
I found this issue in CloudFormation. 

We have to make sure nested objects work properly.. Im not sure if I broke something in my previous PRs... I probably did.. anyhow. Here is a test so we dont break things in the future.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -13,3 +13,4 @@ parameters:
 
     ignoreErrors:
         - '|Ternary operator condition is always true\.|'
+        - '|Negated boolean expression is always false\.|'
--- a/psalm.baseline
+++ b/psalm.baseline
@@ -17,6 +17,9 @@
       <code>$input-&gt;getBody()</code>
     </PossiblyNullArgument>
   </file>
+  <file src="src/Sqs/Result/ReceiveMessageResult.php">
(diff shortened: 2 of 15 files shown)


Nyholm:
Have a look at https://github.com/async-aws/aws/pull/188/commits/92518290dc62af327ee0f617b47b975da03091ec.
We missed that lists are nested like: 

```xml
<DescribeStacksResponse xmlns="http://cloudformation.amazonaws.com/doc/2010-05-15/">
  <DescribeStacksResult>
    <Stacks>
      <member>
        <Outputs>

`

Nyholm:
> We missed that lists are nested like...

This is not always true. If the ListShape does not have a location name, then we should add `->member`

jderusse:
I think we have the same behavior is see send email for CC addresses.

Nyholm:
Thank you for the review. 

Hm. The SES for SendEmail is not the same. In this PR we are dealing with responses. 

Or did I misunderstand you?

jderusse:
https://docs.aws.amazon.com/fr_fr/ses/latest/APIReference/API_SendEmail.html 
looks for `ReplyToAddresses.member.N`

I didn't take time to dig and investigate in `member` thing. But I think it's related.

PR #189, opened 2020-02-22
Update test configuration

Nyholm:
Make sure we tests all components in the "root" test.


Nyholm:
I also exclude the CodeGenerator from the code coverage. Other tests are testing the result of the code generator. So it is covered indirectly.

PR #190, opened 2020-02-22
Simplify Crendential generation when not provided

jderusse:


Nyholm:
Hm.. We should figure this out.. This PR introduced NullProvider in Core. That means that next feature version (minor version or if before 1.0 -> patch version) of Core will include that class. 

Instead of merging this as is, I think we must update branch alias in Core and update requirements for other packages. Does that make sense?

PR #191, opened 2020-02-22
Use filter_var and isset

Nyholm:
Doing the same fixes on RestJsonParser.


Nyholm:
This PR fixes Lambda Invoke. =)

PR #192, opened 2020-02-22
[Symfony] Adding new packages

Nyholm:


src/Symfony/Bundle/DependencyInjection/AwsPackagesProvider.php
```diff
@@ -9,6 +9,14 @@ class AwsPackagesProvider
     public static function getAllServices(): array
     {
         return [
+            'cloudformation' => [
```
Nyholm: - cloudFormation?
- cloud_formation?

jderusse: `cloud-formation` ?
or `CloudFormation` like the official bundle https://github.com/aws/aws-sdk-php-symfony

Nyholm: This goes into the yaml config. That is why Im asking. Having dash in yaml isnt really nice, is it?

Nyholm: We should use snake case to be consistent with other config: 
https://symfony.com/doc/current/reference/configuration/framework.html

jderusse: don't Symfony use snakeCase to separate words? like `max_redirects` but use all lowercase for the same word? like `passphrase`?
just asking.

Nyholm: Yes. Symfony use snake_case to separate words. I think `Cloud Formation` is two words.

PR #194, opened 2020-02-23
Adding tests for S3 Result

Nyholm:


src/Flysystem/S3/Tests/Integration/S3FilesystemV2Test.php
```diff
@@ -108,6 +108,6 @@ private function s3Client(): S3Client
         $this->shouldCleanUp = true;
         $options = ['accessKeyId' => $key, 'accessKeySecret' => $secret, 'region' => $region];
 
-        return $this->s3Client = new S3Client($options, new NullProvider());
```
jderusse: if you remove this,  CI tests will call the AWS endpoint to fetch instance Role, that slows tests.
You should either :
- check if class NullProvider exists
-  use random credentials in options parameters

Nyholm: That is fine for this one. This is only used when some environment variables are being set. Ie, it is not used on CI only when you really want to test with aws servers.

jderusse: are you sure?
by default it:
- look for credentials in options
- look for credentials in HOME/.AWS/credentials
- call AWS endpoint

given in CI we provides neither option nor config file, it always call AWS endpoint.

Nyholm: Yes, have a look here: 
https://github.com/async-aws/aws/blob/master/src/Flysystem/S3/Tests/Integration/S3FilesystemV2Test.php#L96-L111

jderusse: ahhhh. got it.   sorry

PR #195, opened 2020-02-23
Update branch alias and fixed build lowest

Nyholm:
I hope this works =)

### DIFF
--- a/src/Core/composer.json
+++ b/src/Core/composer.json
@@ -16,7 +16,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "0.1-dev"
+            "dev-master": "0.2-dev"
         }
     },
     "autoload": {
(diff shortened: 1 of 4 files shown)


Nyholm:
Wohoo. It is working. 
I think I've figured it out. @jderusse could you please confirm. 

So, related to https://github.com/async-aws/aws/pull/190#issuecomment-590005114
What #190 should have done was to update the branch aliases like in this PR. Because S3 expected the new features in Core. 

I think #190's dep-lowest test would still fail, but the next PR would be green.

jderusse:
good catch..

I think, that an alternative would have been to handle both case (with and without NullProvider) in #190...
Shouldn't we move the lowest test to required?

Nyholm:
> Shouldn't we move the lowest test to required?
Sure, that makes sense now when we have tagged our first version

PR #209, opened 2020-02-23
Add TestCase that simplify assertion on http form

jderusse:


phpstan.neon.dist
```diff
@@ -10,6 +10,7 @@ parameters:
         - src/*/Tests/*
         - src/**/Tests/*
         - src/CodeGenerator/Generator/PhpGenerator/ClassFactory.php
+        - src/Core/Test/TestCase.php
```
jderusse: this class extends TestCase, and we don't requires phpunit

jderusse: replace by a trait would fix that issue

src/CloudFormation/Tests/Unit/Input/DescribeStackEventsInputTest.php
```diff
@@ -3,7 +3,7 @@
 namespace AsyncAws\CloudFormation\Tests\Unit\Input;
 
 use AsyncAws\CloudFormation\Input\DescribeStackEventsInput;
-use PHPUnit\Framework\TestCase;
+use AsyncAws\Core\Test\TestCase;
```
jderusse: not compatible with core 0.2 . provide a migration path.

Nyholm: Core 0.2 is not released yet. So this is fine

Nyholm:
I think it is a.good idea to add a test case. I would be happy to merge after to make PHPStan happy. 

It is okey that Lowest Deps fails. The packages that used the new TestCase required core 0.2. 

If you really dont want to merge this with a red CI, you could split it up to two PRs. First PR where you only make changes to the core package. Once that is merged, you make the second PR with the rest of the changes.

PR #210, opened 2020-02-23
Test ses

jderusse:


src/Ses/Input/SendEmailRequest.php
```diff
@@ -222,12 +231,15 @@ public function requestBody(): string
...
             foreach ($input as $value) {
+                ++$indices->k4a82301;
+
                 if (null !== $value) {
-                    (static function (MessageTag $input) use (&$payload) {
-                        $payload['EmailTags']['member'][]['Name'] = $input->getName();
-                        $payload['EmailTags']['member'][]['Value'] = $input->getValue();
+                    (static function (MessageTag $input) use (&$payload, $indices) {
+                        $payload['EmailTags'][$indices->k4a82301]['Name'] = $input->getName();
+                        $payload['EmailTags'][$indices->k4a82301]['Value'] = $input->getValue();
```
jderusse: This is why I introduced `indices` in that serializer.

The previous code geneates 1 item per property (because of `[]`) 
```
$payload['foo'][]['Key1'] = '1';
$payload['foo'][]['Key2'] = '2';
```

src/Ses/Input/SendEmailRequest.php
```diff
@@ -127,35 +127,44 @@ public function getReplyToAddresses(): array
...
         }
 
         if (null !== $this->Destination) {
-            (static function (Destination $input) use (&$payload) {
-                (static function (array $input) use (&$payload) {
+            (static function (Destination $input) use (&$payload, $indices) {
+                (static function (array $input) use (&$payload, $indices) {
+                    $indices->kbe21867 = -1;
                     foreach ($input as $value) {
-                        $payload['Destination']['ToAddresses']['member'][] = $value;
```
jderusse: I was wrong. The official sdk does not use `member` :-(.

But I don't have only Curl sample to assert it :-(

PR #212, opened 2020-02-24
Use container's http_client when available

jderusse:
That PR leverage the Framework's http client service (with Debug and WebProfile information)


Nyholm:
You need to update the ConfigurationTest.php

PR #214, opened 2020-02-24
Adds a VarDumper ResultCaster

### LINKED ISSUE #211: provide a VarExporter Caster
provide a VarExporter Caster to created readable dumps. (without request, client, ...)

jderusse:
Also move `prefetch` results in the `result` class to remove duplicate code

Fixes #211

### DIFF
--- a/src/Core/Result.php
+++ b/src/Core/Result.php
@@ -37,6 +37,11 @@ class Result
     private $response;
 
     /**
+     * @var self[]
+     */
+    private $prefetchResults = [];
+
+    /**
      * @var HttpClientInterface
      */
     private $httpClient;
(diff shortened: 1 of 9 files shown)


src/Core/Result.php
```diff
@@ -139,6 +148,17 @@ public function __destruct()
         }
 
         $this->response->cancel();
+        unset($this->response);
+    }
+
+    protected function registerPrefetch(self $result): void
```
Nyholm: Make these final

src/Core/Result.php
```diff
@@ -139,6 +148,17 @@ public function __destruct()
         }
 
         $this->response->cancel();
+        unset($this->response);
```
Nyholm: Why do we need to unset the response? The `Result` object is being destructed.

jderusse: if we don't, the Result's destructor calls `$this->resolve()` which triggers a NetworkException because the `$response` is already canceled

src/Core/Result.php
jderusse: fixed, thank

src/Core/Result.php
Nyholm: Oh, yes. that is true. 

Thank you.

PR #216, opened 2020-02-24
Tweaks

nicolas-grekas:
Nice, it's fun to see how the interfaces are used in unextected ways (and it works :) )

### DIFF
--- a/src/Core/AbstractApi.php
+++ b/src/Core/AbstractApi.php
@@ -97,7 +97,7 @@ abstract class AbstractApi
      */
     final protected function getResponse(string $method, $body, $headers = [], ?string $endpoint = null): ResponseInterface
     {
-        $body = $this->stringifyBody($body);
+        $body = $this->stringifyBody($body); // can't we keep the streaming upload behavior?
         if (!isset($headers['content-type'])) {
             $headers['content-type'] = 'text/plain';
         }
--- a/src/Core/AbstractApi.php
+++ b/src/Core/AbstractApi.php
@@ -97,7 +97,7 @@ abstract class AbstractApi
      */
     final protected function getResponse(string $method, $body, $headers = [], ?string $endpoint = null): ResponseInterface
     {
-        $body = $this->stringifyBody($body); // can't we keep the streaming upload behavior?
+        $body = $this->stringifyBody($body);
         if (!isset($headers['content-type'])) {
             $headers['content-type'] = 'text/plain';
         }
--- a/src/Core/Credentials/InstanceProvider.php
+++ b/src/Core/Credentials/InstanceProvider.php
@@ -8,6 +8,7 @@ use AsyncAws\Core\Configuration;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\Exception\DecodingExceptionInterface;
 use Symfony\Contracts\HttpClient\Exception\HttpExceptionInterface;
 use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
@@ -37,7 +38,7 @@ class InstanceProvider implements CredentialProvider
     {
         // fetch current Profile
         try {
-            $response = $this->httpClient->request('GET', self::ENDPOINT, ['timeout' => 1.0]);
+            $response = $this->httpClient->request('GET', self::ENDPOINT, ['timeout' => 1.0]); // why hardcode 1.0 (same below)?
             $profile = $response->getContent();
         } catch (TransportExceptionInterface $e) {
             $this->logger->info('Failed to fetch Profile from Instance Metadata.', ['exception' => $e]);
@@ -52,17 +53,16 @@ class InstanceProvider implements CredentialProvider
         // fetch credentials from profile
         try {
             $response = $this->httpClient->request('GET', self::ENDPOINT . '/' . $profile, ['timeout' => 1.0]);
-            $result = \json_decode($response->getContent(), true);
-            if (\json_last_error() > 0) {
-                $this->logger->info('Failed to decode Credentials.', ['error' => \json_last_error_msg()]);
-
-                return null;
-            }
+            $result = $response->getContent()->toArray()
             if ('Success' !== $result['Code']) {
                 $this->logger->info('Unexpected instance profile.', ['response_code' => $result['Code']]);
 
                 return null;
             }
+        } catch (DecodingExceptionInterface $e) {
+            $this->logger->info('Failed to decode Credentials.', ['exception' => $e]);
+
+            return null;
         } catch (TransportExceptionInterface $e) {
             $this->logger->info('Failed to fetch Profile from Instance Metadata.', ['exception' => $e]);
 
--- a/src/Core/Credentials/InstanceProvider.php
+++ b/src/Core/Credentials/InstanceProvider.php
@@ -28,17 +28,20 @@ class InstanceProvider implements CredentialProvider
 
     private $httpClient;
 
-    public function __construct(?HttpClientInterface $httpClient = null, ?LoggerInterface $logger = null)
+    private $timeout;
+
+    public function __construct(?HttpClientInterface $httpClient = null, ?LoggerInterface $logger = null, float $timeout = 1.0)
     {
         $this->logger = $logger ?? new NullLogger();
         $this->httpClient = $httpClient ?? HttpClient::create();
+        $this->timeout = $timeout;
     }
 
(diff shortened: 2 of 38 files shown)


src/Core/Result.php
```diff
@@ -85,20 +85,22 @@ public function __destruct()
...
+                    $statusCode = $this->response->getStatusCode();
+                    break;
+                }
+            } catch (TransportExceptionInterface $e) {
+                throw new NetworkException('Could not contact remote server.', 0, $e);
             }
-            $statusCode = $this->response->getStatusCode();
-        } catch (TransportExceptionInterface $e) {
-            // When a network error occurs
-            $this->isResolved = true;
```
jderusse: revert

src/Core/AbstractApi.php
```diff
@@ -97,7 +97,7 @@ public function __construct($configuration = [], ?CredentialProvider $credential
      */
     final protected function getResponse(string $method, $body, $headers = [], ?string $endpoint = null): ResponseInterface
     {
-        $body = $this->stringifyBody($body);
+        $body = $this->stringifyBody($body); // can't we keep the streaming upload behavior?
```
Nyholm: No.. and it is super weird. We need the body for signing the request. This is how the official SDK is doing to...

src/Core/Result.php
```diff
@@ -85,20 +85,22 @@ public function __destruct()
...
 
-        try {
-            if (null !== $timeout && $this->httpClient->stream($this->response, $timeout)->current()->isTimeout()) {
-                return false;
+        foreach ($this->httpClient->stream($this->response, $timeout) as $chunk) {
+            try {
+                if ($chunk->isTimeout()) {
+                    return false;
+                }
+                if ($chunk->isFirst()) {
```
jderusse: Not sure to understand what this change solves.

If we receive a chunk it have to be the first right?

src/Core/Result.php
```diff
@@ -143,35 +145,33 @@ public function __destruct()
...
 
         $this->response->cancel();
-        unset($this->response);
+        $this->response = null;
     }
 
     final protected function registerPrefetch(self $result): void
     {
-        $this->prefetchResults[spl_object_hash($result)] = $result;
+        $this->prefetchResults[spl_object_id($result)] = $result;
```
jderusse: :+1:

src/Core/Waiter.php
```diff
@@ -114,14 +114,15 @@ public function __destruct()
...
+     * > When null wait until the end of execution => not what the code does, isn't it? See proposal in Result.php
      *
      * @return bool whether the request is executed or not
      *
      * @throws NetworkException
      */
     final public function resolve(?float $timeout = null): bool
     {
-        if (!isset($this->response)) {
+        if (null === $this->response) {// calling twice resolve() in a row will yield strange results I think
```
Nyholm: Hm.. We currently allow that. It will throw exception twice.

src/Core/Waiter.php
```diff
@@ -114,14 +114,15 @@ public function __destruct()
      * Make sure the actual request is executed.
      *
      * @param float|null $timeout Duration in seconds before aborting. When null wait until the end of execution.
+     * > When null wait until the end of execution => not what the code does, isn't it? See proposal in Result.php
```
jderusse: Should be `wait until the arrival of the first bits`?

src/Core/Waiter.php
```diff
@@ -197,7 +198,7 @@ public function __destruct()
             if ($delay > $timeout - (\microtime(true) - $start)) {
                 break;
             }
-            \usleep((int) ceil($delay * 1000000));
+            \usleep((int) ceil($delay * 1000000)); // This looks suspicious, there should be never be a need to sleep to me
```
jderusse: In AWS, a waiter implements a polling pattern:
- sqs->createQueue()
- sqs->queueExist()->wait()
- - sqs->getQueueUrl() // 404
- - sleep 5sec
- - sqs->getQueueUrl() // 404
- - sleep 5sec
- - sqs->getQueueUrl() // 200

$delay, is the time to wait between 2 calls

src/Lambda/Result/AddLayerVersionPermissionResponse.php
```diff
@@ -34,7 +34,7 @@ public function getStatement(): ?string
 
     protected function populateResult(ResponseInterface $response, HttpClientInterface $httpClient): void
     {
-        $data = json_decode($response->getContent(false), true);
+        $data = $response->getContent(false)->toArray();
```
jderusse: you means `$response->toArray(false)` right?

src/Core/Waiter.php
nicolas-grekas: but what if there is no exception?

src/Core/AbstractApi.php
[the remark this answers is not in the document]
nicolas-grekas: maybe then you should accept only a string body?

src/Core/Credentials/InstanceProvider.php
```diff
@@ -37,7 +38,7 @@ public function getCredentials(Configuration $configuration): ?Credentials
     {
         // fetch current Profile
         try {
-            $response = $this->httpClient->request('GET', self::ENDPOINT, ['timeout' => 1.0]);
+            $response = $this->httpClient->request('GET', self::ENDPOINT, ['timeout' => 1.0]); // why hardcode 1.0 (same below)?
```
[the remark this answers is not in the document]
nicolas-grekas: or left to the defaults?

src/Core/Result.php
nicolas-grekas: this means the method becomes order-sentitive

src/Core/Result.php
nicolas-grekas: in theory, there can be informative chunks before the first one

src/Core/Waiter.php
nicolas-grekas: the client is not used at all right if the timeout is null - I fail to understand how this can doanything useful...

src/Core/Waiter.php
nicolas-grekas: there is a timeout argument already - it's enough to "wait" for me.

src/Lambda/Result/AddLayerVersionPermissionResponse.php
nicolas-grekas: I suppose, depending on how you want to deal with 300-599 here :)

src/Core/Waiter.php
jderusse: I think you missed a piece here. This method is polling an endpoint.

`sqs->queueExist()->wait(300, 5)` means, call the endpoint `getQueueUrl` every 5 seconds until the response is not 404 (the methods `getQueueUrl` responds in few milliseconds). Abort after 5 minutes.

Not sleeping between each call, will triggers a RateLimit error from AWS.

nicolas-grekas: OK, you must be right :)

src/Core/Waiter.php
jderusse: ohh.. good catch

src/Core/Result.php
jderusse: thanks for the hint

src/Core/Result.php
jderusse: refactored

src/Core/Result.php
```diff
@@ -85,38 +84,50 @@ public function __destruct()
...
-        if ($this->isResolved) {
+        if (null !== $this->resolveResult) {
+            if ($this->resolveResult instanceof \Exception) {
+                throw $this->resolveResult;
+            }
+
+            if (\is_bool($this->resolveResult)) {
+                return $this->resolveResult;
+            }
+        }
```
Nyholm: Maybe throw exception if it is unexpected value.

src/Core/Waiter.php
```diff
@@ -121,20 +128,39 @@ public function __destruct()
...
+        }
+
+        if (null === $this->response) {
             return true;
         }
 
         try {
-            if (null !== $timeout && $this->httpClient->stream($this->response, $timeout)->current()->isTimeout()) {
-                return false;
+            if (null !== $timeout) {
```
Nyholm: We don’t need this If statement, right?

The behavior will be the same without it.

jderusse: We don't need it. But test won't pass because MockClient::stream does accept SimpleMockResponse

src/Core/Result.php
jderusse: done

src/Core/Waiter.php
jderusse: removed (and fixed issue with MockStream)

jderusse:
Thank you @nicolas-grekas

PR #223, opened 2020-02-25
Stream http request

jderusse:
this allows : 
```php
$s3 = new S3Client(['region' => 'eu-west-1']);
$s3->PutObject(['Bucket' => 'Foo', 'Key' => 'string', 'Body' => 'content']);
```
or
```php
$s3 = new S3Client(['region' => 'eu-west-1']);
$s3->PutObject(['Bucket' => 'Foo', 'Key' => 'resource', 'Body' => \fopen(__FILE__, 'r')]);
```
or
```php
$s3 = new S3Client(['region' => 'eu-west-1']);
$cpt = 0;
$s3->PutObject(['Bucket' => 'Foo', 'Key' => 'callable', 'Body' => function() use (&$cpt) {
return $cpt++>2 ? '' : __METHOD__;
}]);
```

### DIFF
--- a/src/Core/Signers/Request.php
+++ b/src/Core/Signers/Request.php
@@ -20,13 +20,16 @@ final class Request
     private $body;
 
     /**
-     * @param string[]|string[][] $headers
+     * @param string[]|string[][]      $headers
+     * @param string|resource|callable $body
      */
-    public function __construct(string $method, string $url, array $headers, string $body)
+    public function __construct(string $method, string $url, array $headers, $body)
     {
         $this->method = $method;
         $this->url = $url;
-        $this->headers = $headers;
+        foreach ($headers as $key => $value) {
+            $this->headers[\strtolower($key)] = $value;
+        }
         $this->body = $body;
     }
 
@@ -42,12 +45,12 @@ final class Request
 
     public function hasHeader($name): bool
     {
-        return \array_key_exists($name, $this->headers);
+        return \array_key_exists(strtolower($name), $this->headers);
     }
 
     public function setHeader($name, $value): void
     {
-        $this->headers[$name] = $value;
+        $this->headers[strtolower($name)] = $value;
     }
 
     public function getHeaders(): array
@@ -55,8 +58,27 @@ final class Request
         return $this->headers;
     }
 
-    public function getBody(): string
+    /**
+     * @return string|string[]
+     */
+    public function getHeader(string $name)
+    {
+        return $this->headers[strtolower($name)] ?? null;
+    }
+
+    /**
+     * @return string|resource|callable
+     */
+    public function getBody()
     {
         return $this->body;
+    }
+
+    /**
+     * @param string|resource|callable $body
+     */
+    public function setBody($body)
+    {
+        $this->body = $body;
     }
 }
--- a/src/Core/Signers/SignerV4.php
+++ b/src/Core/Signers/SignerV4.php
@@ -4,6 +4,7 @@ namespace AsyncAws\Core\Signers;
 
 use AsyncAws\Core\Credentials\Credentials;
 use AsyncAws\Core\Exception\InvalidArgument;
+use AsyncAws\Core\Exception\RuntimeException;
 
 /**
  * Version4 of signer.
@@ -12,6 +13,10 @@ use AsyncAws\Core\Exception\InvalidArgument;
  */
 class SignerV4 implements Signer
 {
+    private const ALGORITHM_REQUEST = 'AWS4-HMAC-SHA256';
+    private const ALGORITHM_CHUNK = self::ALGORITHM_REQUEST . '-PAYLOAD';
+    private const CHUNK_SIZE = 64 * 1024;
+
     private const BLACKLIST_HEADERS = [
         'cache-control' => true,
         'content-type' => true,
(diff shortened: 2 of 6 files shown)


src/S3/Tests/Integration/S3ClientTest.php
```diff
@@ -486,9 +486,11 @@ public function testPutObjectAcl(): void
 
     public function testUploadFromClosure()
     {
-        $content = 'some content';
-        $closure = \Closure::fromCallable(function () use ($content) {
-            return $content;
```
jderusse: This is a change is the way I deal with callable.

Previously the callable returns the entiere body. Which is useless because everything will be set in memory.

I changed the behavior to use the same pattern that Curl and the HttpClient call the method until it returns an empty string.

An other solution could be to accept `\Generator` instead of callable.

src/Core/Signers/SignerV4.php
```diff
@@ -73,49 +94,148 @@ public function sign(Request $request, ?Credentials $credentials): void
...
+
+        if (\is_string($body)) {
+            $request->setHeader('x-amz-content-sha256', hash('sha256', $body));
+            $request->setHeader('content-length', $contentLength);
+
+            return;
+        }
+
+        $streamReader = null;
+        if (\is_callable($body)) {
```
Nyholm: This will only happen if body is a callable and the content-length header was defined, right?

src/S3/Tests/Integration/S3ClientTest.php
Nyholm: Could you document a proper callable in the S3 readme? (or somewhere you see fit)

Nyholm: We should also document the argument to the callable

jderusse: Good Idea. 

Added a comment in the readme

src/Core/Signers/Request.php
```diff
@@ -20,13 +20,16 @@ final class Request
...
+     * @param string|resource|callable $body
      */
-    public function __construct(string $method, string $url, array $headers, string $body)
+    public function __construct(string $method, string $url, array $headers, $body)
     {
         $this->method = $method;
         $this->url = $url;
-        $this->headers = $headers;
+        foreach ($headers as $key => $value) {
+            $this->headers[\strtolower($key)] = $value;
```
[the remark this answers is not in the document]
jderusse: Does it whorth it? The class is `@internal` and not exposed to userland.

Maintaining a mapping requires to solve issues like `what if I call `setHeader('Foo', 'x')` then `setHeader('FOO', 'x')` should we keep the first or the last.

src/Core/Signers/SignerV4.php
jderusse: yes, if the size is unknown (not sent via header, and is not a stream), then the method tries first to convert the stream/callable into a string. And the follow the old way

src/Core/Signers/Request.php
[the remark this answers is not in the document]
Nyholm: Let's skip this for now and see if AWS's headers are case sensitive.

src/S3/README.md
```diff
@@ -10,3 +10,30 @@ An API client for S3.
...
+
+## Upload File via Stream
+
+```php
+$s3Client = new S3Client(['region' => 'eu-west-1']);
+
+$resource = \fopen('/path/to/big/file', 'r');
+$s3Client->PutObject([
+    'Bucket' => 'ah-brightcove',
+    'Key' => 'resource',
```
Nyholm: name this something else... Like `file.jpg`
```suggestion
    'Key' => 'file.jpg',
```

src/S3/README.md
```diff
@@ -10,3 +10,30 @@ An API client for S3.
...
+    'Bucket' => 'ah-brightcove',
+    'Key' => 'callable',
+    'ContentLength' => filesize('/path/to/big/file'), // This is important
+    'Body' => static function(int $length) use ($fp) {
+        return fread($fp, $length);
+    },
+]);
+```
+
+When using a `Closure`,it's important to provide the property ContentLength.
```
Nyholm:
```suggestion
When using a `Closure`, it's important to provide the property `ContentLength`.
```

src/S3/README.md
```diff
@@ -10,3 +10,30 @@ An API client for S3.
...
+    'Bucket' => 'ah-brightcove',
+    'Key' => 'resource',
+    'Body' => $resource,
+]);
+
+// or via Closure
+$fp = \fopen('/path/to/big/file', 'r');
+$s3Client->PutObject([
+    'Bucket' => 'ah-brightcove',
+    'Key' => 'callable',
```
Nyholm:
```suggestion
    'Key' => 'file.jpg',
```

src/S3/README.md
```diff
@@ -10,3 +10,30 @@ An API client for S3.
...
+    'ContentLength' => filesize('/path/to/big/file'), // This is important
+    'Body' => static function(int $length) use ($fp) {
+        return fread($fp, $length);
+    },
+]);
+```
+
+When using a `Closure`,it's important to provide the property ContentLength.
+This information is required by AWS, and cannot be guessed by AsyncAws.
```
Nyholm:
```suggestion
This information is required by AWS, and cannot be guessed by AsyncAws.
If `ContentLength` is absent, AsyncAws will read the output into memory before 
sending the request.
```

jderusse: Oh yes, you're right it's important

Nyholm:
Thank you very much for this

PR #226, opened 2020-02-26
Stream

jderusse:
The Stream provides usefull methods (for Signer) to split a body in chunks

### DIFF
--- a/src/Core/AbstractApi.php
+++ b/src/Core/AbstractApi.php
@@ -14,6 +14,7 @@ use AsyncAws\Core\Exception\InvalidArgument;
 use AsyncAws\Core\Signer\Request;
 use AsyncAws\Core\Signer\Signer;
 use AsyncAws\Core\Signer\SignerV4;
+use AsyncAws\Core\Stream\StreamFactory;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Symfony\Component\HttpClient\HttpClient;
@@ -74,8 +75,8 @@ abstract class AbstractApi
     }
 
     /**
-     * @param string[]|string[][]            $headers headers names provided as keys or as part of values
-     * @param array|string|resource|callable $body
+     * @param string[]|string[][]                    $headers headers names provided as keys or as part of values
+     * @param string|resource|callable|iterable|null $body
      */
     final public function request(string $method, $body = '', $headers = [], ?string $endpoint = null): Result
     {
@@ -91,8 +92,8 @@ abstract class AbstractApi
     abstract protected function getSignatureScopeName(): string;
 
     /**
-     * @param string[]|string[][]      $headers headers names provided as keys or as part of values
-     * @param string|resource|callable $body
+     * @param string[]|string[][]                    $headers headers names provided as keys or as part of values
+     * @param string|resource|callable|iterable|null $body
      */
     final protected function getResponse(string $method, $body, $headers = [], ?string $endpoint = null): ResponseInterface
     {
@@ -100,8 +101,14 @@ abstract class AbstractApi
             $headers['content-type'] = 'text/plain';
         }
 
-        $request = new Request($method, $this->fillEndpoint($endpoint), $headers, $body);
+        $stream = StreamFactory::create($body);
+
+        $request = new Request($method, $this->fillEndpoint($endpoint), $headers, $stream);
         $this->getSigner()->sign($request, $this->credentialProvider->getCredentials($this->configuration));
+
+        if (!$request->hasHeader('content-length') && null !== $length = $request->getBody()->length()) {
+            $request->setHeader('content-length', $length);
+        }
 
         return $this->httpClient->request($request->getMethod(), $request->getUrl(), ['headers' => $request->getHeaders(), 'body' => $request->getBody()]);
     }
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -8,12 +8,14 @@
 - Methods `AwsClient::cloudFormation()`, `AwsClient::lambda()`, `AwsClient::sns()`
 - Protected methods `Result::registerPrefetch()` and `Result::unregisterPrefetch()`
 - Timeout parameter to `InstanceProvider::__construct()`
+- Requests can now be streamed
+- Strealable request accepts iterable alongside string, callable, resource
 
 ### Changed
 
 - Removed `AwsClient` and replaced it with `AwsClientFactory`
-- Class `AsuncAws\Core\Signer\Request` is marked as internal
-- Make sure behavior of calling `Result::resolve()` is consistent.
+- Class `AsyncAws\Core\Signer\Request` is marked as internal
+- Make sure behavior of calling `Result::resolve()` is consistent
 - Rename namespace `Signers` into `Signer`.
 
 ## 0.1.0
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Protected methods `Result::registerPrefetch()` and `Result::unregisterPrefetch()`
 - Timeout parameter to `InstanceProvider::__construct()`
 - Requests can now be streamed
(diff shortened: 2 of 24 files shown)


src/Core/Signers/SignerV4.php
```diff
@@ -54,13 +57,10 @@ public function __construct(string $scopeName, string $region)
...
     {
-        $body = $request->getBody() ?? '';
-        if (\is_resource($body) && -1 === fseek($body, 0)) {
-            throw new RuntimeException('Unable to seek the resource');
-        }
-
         if (null === $credentials) {
-            $request->setHeader('content-length', $this->getContentLength($request));
+            if (!$request->hasHeader('content-length') && null !== $length = $request->getBody()->length()) {
+                $request->setHeader('content-length', $length);
```
jderusse: This is not really the role of the Signer, it could be done in the AbstractApi

src/Core/AbstractApi.php
```diff
@@ -91,16 +95,30 @@ public function __construct($configuration = [], ?CredentialProvider $credential
     abstract protected function getSignatureScopeName(): string;
 
     /**
-     * @param string[]|string[][]      $headers headers names provided as keys or as part of values
-     * @param string|resource|callable $body
+     * @param string[]|string[][]                    $headers headers names provided as keys or as part of values
+     * @param string|resource|callable|iterable|null $body
      */
     final protected function getResponse(string $method, $body, $headers = [], ?string $endpoint = null): ResponseInterface
```
jderusse: I wonder if it shouldn't be `Stream $body`. This thing will be more efficient if performed by the called

Nyholm: That require that we update all `requestBody` functions.

src/Core/Signers/SignerV4.php
Nyholm: Yes, lets!

src/Core/AbstractApi.php
Nyholm: Hm.. Currently, we can mark all Streams as `@internal`. Is that a good thing?

src/Core/AbstractApi.php
```diff
@@ -91,16 +95,30 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
+        $stream = null;
+        if (null === $body || \is_string($body)) {
+            $stream = new StringStream($body ?? '');
+        } elseif (\is_resource($body)) {
+            $stream = new ResourceStream($body);
+        } elseif (\is_callable($body)) {
+            $stream = new CallableStream($body);
+        } elseif (\is_iterable($body)) {
+            $stream = new IterableStream($body);
+        }
```
Nyholm: If we decide to not make streams internal, we should move this to a `StreamFactory`

src/Core/Streams/CallableStream.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace AsyncAws\Core\Streams;
```
Nyholm: Can't we use `Stream` as namespace? 

Using singular is a good practice.

src/Core/Streams/FixedSizeStream.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return $this->decorated->length();
+    }
+
+    public function stringify(): string
+    {
+        return $this->decorated->stringify();
+    }
+
+    public function getIterator(): \Traversable
```
Nyholm: Nice. Did you write this yourself?  

It looks like one of these method you find as a 1000+ votes on StackOverflow  =)

src/Core/Signers/SignerV4.php
```diff
@@ -214,31 +159,38 @@ private function prepareBody(Request $request, string $amzDate, string $credenti
...
-
-        $last = false;
-        $streamBody = static function () use ($streamReader, $amzDate, $credentialScope, &$signature, &$last, $signingKey): string {
-            if ($last) {
-                return '';
+        $request->setHeader('content-length', $contentLength + $fullChunkCount * ($metaLength + \strlen((string) dechex(self::CHUNK_SIZE))) + ($chunkCount - $fullChunkCount) * ($metaLength + \strlen((string) dechex($contentLength % self::CHUNK_SIZE))) + $metaLength + 1);
+
+        if (!$body instanceof FixedSizeStream) {
+            $body = new FixedSizeStream($body, self::CHUNK_SIZE);
+        }
```
Nyholm: These 3 lines could be removed if we were using a named factory. Either: 

```php
$body = StreamFactory::fixedSizeStream($body);
```
or
```php
$body = FixedSizeStream::create($body);
```

src/Core/Signers/SignerV4.php
```diff
@@ -123,89 +123,34 @@ public function sign(Request $request, ?Credentials $credentials): void
...
+            $contentLength = $body->length();
         }
 
-        if (\is_string($body)) {
-            $request->setHeader('x-amz-content-sha256', hash('sha256', $body));
+        // no need to stream small body
+        if ($contentLength < self::CHUNK_SIZE) {
+            if (!$body instanceof StringStream) {
+                $request->setBody($body = new StringStream($body->stringify()));
+            }
```
Nyholm: These 3 lines could be removed if we were using a named factory. Either: 

```php
$body = StreamFactory::stringStream($body);
```
or
```php
$body = StringStream::create($body);
```

src/Core/Streams/ResourceStream.php
```diff
@@ -0,0 +1,57 @@
...
+        return fstat($this->content)['size'] ?? null;
+    }
+
+    public function stringify(): string
+    {
+        if (-1 === fseek($this->content, 0)) {
+            throw new InvalidArgument('Unable to seek the content.');
+        }
+
+        return \stream_get_contents($this->content);
```
Nyholm: What if we got a "read once" stream? Should we care? Will it work?

src/Core/Streams/CallableStream.php
Nyholm: Hm.. I see that we use `Signers` too.

src/Core/AbstractApi.php
jderusse: Let's move `Stream` to `@internal`, and we can remove it later if we want.
I also set classes final

src/Core/Streams/CallableStream.php
jderusse: Fixed (Signers will follow in a dedicated PR)

src/Core/Streams/FixedSizeStream.php
jderusse: thanks for the kind words.

I take inspiration from https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpClient/CurlHttpClient.php#L438-L453

But my version take care of memory

src/Core/Streams/ResourceStream.php
jderusse: The stream must be seekable (checked in constructor). Can a stream be "read once" and "Seakable"?

src/Core/Signers/SignerV4.php
jderusse: good idea.

src/Core/AbstractApi.php
jderusse: Added a streamFactory. It's simpler

src/Core/Signers/SignerV4.php
jderusse: good idea

src/Core/CHANGELOG.md
```diff
@@ -8,12 +8,14 @@
 - Methods `AwsClient::cloudFormation()`, `AwsClient::lambda()`, `AwsClient::sns()`
 - Protected methods `Result::registerPrefetch()` and `Result::unregisterPrefetch()`
 - Timeout parameter to `InstanceProvider::__construct()`
+- Requests can now be streamed
+- Strealable request accepts iterable alongside string, callable, resource
```
Nyholm:
```suggestion
- Streamable request accepts iterable alongside string, callable, resource
```

PR #229, opened 2020-02-27
Use better autoloading for test files

Nyholm:


composer.json
```diff
@@ -31,7 +31,8 @@
     "autoload": {
         "psr-4": {
             "AsyncAws\\": "src"
-        }
+        },
+        "files": ["src/Core/Test/PHPUnitTestCase.php"]
```
jderusse: Hmmm.. I'm not sure it should be loaded for non-dev

Nyholm: In already is.. And it is just a small file.

Nyholm: It needs to be here, or it will not be auto loaded from ie the s3 client.

Nyholm: I could maybe improve

src/Core/composer.json
```diff
@@ -23,6 +23,7 @@
         "psr-4": {
             "AsyncAws\\Core\\": ""
         },
+        "files": ["Test/PHPUnitTestCase.php"],
```
jderusse: I think it's only needed in `Generator` (+ `Aws` because of subtreesplit)  when using Reflection to change the source code of a Test.
When running test, there is no issue, the Class `PHPUnit\TestCase` is well loaded.

People using `Core` + our testCase should not be impacted

composer.json
jderusse: S3 client don't need it's unless they use it in tests, and in that case phpunit should be autoloaded

src/Core/composer.json
Nyholm: I've updated the solution now. I pretty much only moved DummyTest to a new file.

Nyholm:
The branch aliases are unrelated. I made them so the build is green

src/Core/composer.json
jderusse: do not missunderstand my comment. I liked using composer to load the alias. 

I was just worried about loading it for everybody with "production" settings.

Nyholm: Yeah. It was. 

I changed that so it is only loaded when the class is used.

PR #231, opened 2020-02-27
Fix missing dep in composer.json

### LINKED ISSUE #206: Update PHP extensions in composer.json
Each client is different. Some need `ext-SimpleXML` and some need `ext-json`. The composer.json should be updated accordingly.

jderusse:
Fixes #206


src/Flysystem/S3/composer.json
```diff
@@ -16,6 +16,8 @@
         }
     ],
     "require": {
+        "php": "^7.2.5",
+        "async-aws/core": "^0.1 || ^0.2",
```
Nyholm: The core package is not needed, right? We never use that one directly.

jderusse: we use it for exception and return typehint.
composer-require-checker tell me

PR #233, opened 2020-02-28
Move all files to src

Nyholm:
This will give us proper packages with src and tests folder.

The only downside is that we need to write one extra line in the `autoload` and `autoload-dev` section of our root composer.json. 

Related to https://github.com/Roave/BetterReflection/issues/551


Nyholm:
Thank you

PR #241, opened 2020-02-29
Move files to a new structure

Nyholm:
This will fix #241


Nyholm:
Hm.. We lost all history on the clients.. 
It doesnt matter too much, but just an FYI. 

Lucky we will not move files anymore =)

Nyholm:
I released new version to make sure the release diff makes sense.

PR #248, opened 2020-02-29
Improve organization overview. 

Nyholm:
I added this as an example of namespace, package name and location. I dont think it fill any purpose atm.


Nyholm:
I changed my mind. I'll make it useful instead of removing it

PR #254, opened 2020-03-01
Add tests for S3

jderusse:


src/Service/S3/src/Result/CopyObjectOutput.php
```diff
@@ -144,9 +144,9 @@ protected function populateResult(ResponseInterface $response, HttpClientInterfa
         $this->RequestCharged = $headers['x-amz-request-charged'][0] ?? null;
 
         $data = new \SimpleXMLElement($response->getContent(false));
-        $this->CopyObjectResult = !$data->CopyObjectResult ? null : new CopyObjectResult([
```
jderusse: spotted by test :-) base on S3 documentation.
When response contains a Payload, the there is no "intermediate" object.

PR #257, opened 2020-03-01
Leverage Enum properties provided by AWS

### LINKED ISSUE #236: Handle Enum
The api2-json file provides hints on enum. (ie. the Runtime Type in in Lambda)

We should:
- provide const for the values
- take those value in "validators"

```php
class ListLayerVersionsRequest
{
    /**
     * A runtime identifier. For example, `go1.x`.
     *
     * @var string|null
     */
    private $CompatibleRuntime;

    public function validate(): void
    {
        foreach ($this->CompatibleRuntime as $value) {
            if (!isset(Runtime::AVAILABLE_RUNTIMES[$value])) {
                throw new InvalidArgument(sprintf('The CompatibleRuntime "%s" in "%s" is not a Valide "Runtime". Available values are "%"', $value, __CLASS__, array_keys(Runtime::AVAILABLE_RUNTIMES)));
            }
        }
    }
}

// Class is named "Runtime" because of the name of the shape
class Runtime 
{
    public const NODEJS = 'nodejs';
    public const NODEJS_4_3 = 'nodejs4.3';
    public const NODEJS_6_10 = 'nodejs6.10';
    public const NODEJS_4_3_EDGE = 'nodejs4.3-edge';
    // ..

    // = "AVAILABLE_" + className.upper()
    public const AVAILABLE_RUNTIME = [
        self::NODEJS = true,
        self::NODEJS_4_3 = true,
        self::NODEJS_6_10 = true,
        self::NODEJS_4_3_EDGE = true,
        // ..
    ];
}
```

jderusse:
fixes #236

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -10,3 +10,9 @@ parameters:
 			count: 1
 			path: src/Integration/Symfony/Bundle/src/DependencyInjection/Configuration.php
 
+		-   "#^PHPDoc tag @param has invalid value#"
+		-   "#^PHPDoc tag @return with type AsyncAws\\\\[^\\\\]+\\\\Enum\\\\[^ ]+ is incompatible with native type string(\\|null)?\\.$#"
+		-   "#^Method AsyncAws\\\\[^ ]+ should return string\\|null but returns AsyncAws\\\\[^\\\\]+\\\\Enum\\\\[^\\.]+\\.$#"
+		-   "#^Property AsyncAws\\\\[^ ]+ \\(AsyncAws\\\\[^\\\\]+\\\\Enum\\\\[^\\)]+\\) does not accept string\\|null\\.$#"
+		-   "#^Invalid array key type AsyncAws\\\\[^\\\\]+\\\\Enum\\\\[^\\\\]+\\.$#"
+		-   "#^Parameter \\#3 ...\\$args of function sprintf expects bool\\|float\\|int\\|string\\|null, AsyncAws\\\\[^\\\\]+\\\\Enum\\\\[^ ]+ given\\.$#"
--- a/src/Core/src/Sts/Input/AssumeRoleRequest.php
+++ b/src/Core/src/Sts/Input/AssumeRoleRequest.php
@@ -155,6 +155,9 @@ class AssumeRoleRequest
         return $this->Policy;
     }
 
+    /**
+     * @return PolicyDescriptorType[]
+     */
     public function getPolicyArns(): array
     {
         return $this->PolicyArns;
@@ -175,6 +178,9 @@ class AssumeRoleRequest
         return $this->SerialNumber;
     }
 
+    /**
+     * @return Tag[]
+     */
     public function getTags(): array
     {
         return $this->Tags;
@@ -185,6 +191,9 @@ class AssumeRoleRequest
         return $this->TokenCode;
     }
 
+    /**
+     * @return string[]
+     */
     public function getTransitiveTagKeys(): array
     {
         return $this->TransitiveTagKeys;
@@ -292,6 +301,9 @@ class AssumeRoleRequest
         return $this;
     }
 
+    /**
+     * @param PolicyDescriptorType[] $value
+     */
     public function setPolicyArns(array $value): self
     {
         $this->PolicyArns = $value;
@@ -320,6 +332,9 @@ class AssumeRoleRequest
         return $this;
     }
 
+    /**
+     * @param Tag[] $value
+     */
     public function setTags(array $value): self
     {
         $this->Tags = $value;
@@ -334,6 +349,9 @@ class AssumeRoleRequest
         return $this;
     }
 
+    /**
+     * @param string[] $value
+     */
     public function setTransitiveTagKeys(array $value): self
     {
         $this->TransitiveTagKeys = $value;
@@ -343,14 +361,18 @@ class AssumeRoleRequest
 
     public function validate(): void
     {
-        foreach (['RoleArn', 'RoleSessionName'] as $name) {
-            if (null === $this->$name) {
-                throw new InvalidArgument(sprintf('Missing parameter "%s" when validating the "%s". The value cannot be null.', $name, __CLASS__));
-            }
+        if (null === $this->RoleArn) {
+            throw new InvalidArgument(sprintf('Missing parameter "RoleArn" when validating the "%s". The value cannot be null.', __CLASS__));
         }
+
+        if (null === $this->RoleSessionName) {
+            throw new InvalidArgument(sprintf('Missing parameter "RoleSessionName" when validating the "%s". The value cannot be null.', __CLASS__));
+        }
+
         foreach ($this->PolicyArns as $item) {
             $item->validate();
         }
+
         foreach ($this->Tags as $item) {
             $item->validate();
         }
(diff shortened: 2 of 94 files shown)


src/Service/CloudFormation/src/Result/StackEvent.php
```diff
@@ -43,6 +45,8 @@ class StackEvent
 
     /**
      * Current status of the resource.
+     *
+     * @see ResourceStatus
```
jderusse: does somebody know how to define enum in dockblock? something like `@var string{'foo'|'bar'}`.

src/Service/Lambda/src/Input/PublishLayerVersionRequest.php
```diff
@@ -196,13 +197,19 @@ public function setLicenseInfo(?string $value): self
...
-                throw new InvalidArgument(sprintf('Missing parameter "%s" when validating the "%s". The value cannot be null.', $name, __CLASS__));
-            }
+        if (null === $this->LayerName) {
+            throw new InvalidArgument(sprintf('Missing parameter "LayerName" when validating the "%s". The value cannot be null.', __CLASS__));
+        }
+
+        if (null === $this->Content) {
+            throw new InvalidArgument(sprintf('Missing parameter "Content" when validating the "%s". The value cannot be null.', __CLASS__));
         }
-        if ($this->Content) {
```
jderusse: I also refactor how requires properties are handled in validation: I replace `foreach (['foo', 'bar'` by `if (foo... if (bar...`.

That allows me to group similar validation on same properties and fix useless check like this one.

src/Service/S3/tests/Integration/S3ClientTest.php
```diff
@@ -300,9 +302,9 @@ public function testPutObjectAcl(): void
                 'Grants' => [new Grant([
                     'Grantee' => new Grantee([
                         'DisplayName' => 'me',
-                        'Type' => 'canonical',
+                        'Type' => Type::CANONICALUSER,
```
jderusse: Bug spotted :)

src/Service/S3/src/Enum/Type.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace AsyncAws\S3\Enum;
+
+class Type
+{
+    public const AMAZONCUSTOMERBYEMAIL = 'AmazonCustomerByEmail';
+    public const CANONICALUSER = 'CanonicalUser';
```
jderusse: should be `CANONICAL_USER`

src/Service/S3/tests/Integration/S3ClientTest.php
Nyholm: Are you sure?

src/Core/src/Sts/Input/AssumeRoleWithWebIdentityRequest.php
```diff
@@ -240,11 +240,18 @@ public function setWebIdentityToken(?string $value): self
 
     public function validate(): void
     {
-        foreach (['RoleArn', 'RoleSessionName', 'WebIdentityToken'] as $name) {
-            if (null === $this->$name) {
```
Nyholm: Did we remove “name”?

src/Service/S3/tests/Integration/S3ClientTest.php
jderusse: The const name is generated. So, yes, we ca choose whatever we want :)
Note: the value of the constant remain the same and is the value expected by AWS.

Maybe it's not obvious for `CANONICALUSER`, I found that `AMAZON_CUSTOMER_BY_EMAIL` is most readable than `AMAZONCUSTOMERBYEMAIL`

I fixed it.

jderusse: yes. The right value is `CanonicalUser`

src/Service/S3/src/Enum/Type.php
jderusse: fixe

src/Core/src/Sts/Input/AssumeRoleWithWebIdentityRequest.php
jderusse: `"name"`?.. you mean `$name` which is a variable containing either `RoleArn`, `RoleSessionName` or `WebIdentityToken`. The check is still performed bellow

src/Service/Lambda/src/Input/PublishLayerVersionRequest.php
Nyholm: Cool. I see this now.

src/Service/S3/src/Input/PutObjectRequest.php
```diff
@@ -799,9 +815,47 @@ public function setWebsiteRedirectLocation(?string $value): self
...
+        }
+
+        if (null !== $this->StorageClass) {
+            if (!isset(StorageClass::AVAILABLE_STORAGECLASS[$this->StorageClass])) {
+                throw new InvalidArgument(sprintf('Invalid parameter "StorageClass" when validating the "%s". The value "%s" is not a valid "StorageClass". Available values are %s.', __CLASS__, $this->StorageClass, implode(', ', array_keys(StorageClass::AVAILABLE_STORAGECLASS))));
+            }
+        }
+
+        if (null !== $this->RequestPayer) {
+            if (!isset(RequestPayer::AVAILABLE_REQUESTPAYER[$this->RequestPayer])) {
```
Nyholm: Just making an observation: if aws adds something to this enum value, we need to release a new version before someone can use the new aws feature.

src/Service/CloudFormation/src/Result/StackEvent.php
Nyholm: Look at the bottom of this page: https://psalm.dev/docs/annotating_code/typing_in_psalm/

src/Service/S3/src/Input/PutObjectRequest.php
jderusse: Good catch, but I think this is similar for every change in the Api contract (new property, new method, new(or remove) requirement...

Nyholm: See #258

src/Core/src/Sts/Input/AssumeRoleRequest.php
```diff
@@ -175,6 +178,9 @@ public function getSerialNumber(): ?string
         return $this->SerialNumber;
     }
 
+    /**
+     * @return Tag[]
```
jderusse: Not really related to that PR, but definining such dockbloc helps users and static tools (phpstan/psalm) following variable assignments.

Nyholm: Yes. We should do more of these

phpstan-baseline.neon
```diff
@@ -10,3 +10,9 @@ parameters:
 			count: 1
 			path: src/Integration/Symfony/Bundle/src/DependencyInjection/Configuration.php
 
+		-   "#^PHPDoc tag @param has invalid value#"
```
jderusse: phpstan don't like the Enum param docblock.. I had to ignore all those rules :/

Nyholm:
Is this ready to merge?

jderusse:
Yes, it's RFR => If your're OK with new PHPStan ignore rules  (cf https://github.com/async-aws/aws/pull/257#discussion_r386133675)

phpstan-baseline.neon
Nyholm: Im okey with this. Pslam is happy.

Issue #258, opened 2020-03-01
Periodically check the AWS Api contract


jderusse:
Add a GithubAction that dry-run the generation based on the latest available version, and trigger an alert/or create a PR when something changes.

That should fixes https://github.com/async-aws/aws/pull/257#discussion_r386121947

Nyholm:
Fixed in #262

PR #260, opened 2020-03-02
Download last metadata version

jderusse:
Fetch the last version from packagist, and download the last AWS Sdk stable version

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,22 +1,25 @@
 {
+    "variables": {
+        "${LATEST}": "3.133.6"
+    },
     "services": {
         "CloudFormation": {
-            "source": "https:\/\/raw.githubusercontent.com\/aws\/aws-sdk-php\/3.133.6\/src\/data\/cloudformation\/2010-05-15\/api-2.json",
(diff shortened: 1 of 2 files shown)


Nyholm:
How is the cache cleared? Is it only manual?

What if I:
- clone
- generate
- aws release
- make a PR
- PR will fail because my output does not match generated output

jderusse:
- the call to packagist (to fetch last version) is cached for 5 minutes
- the call to github (to fetch each .json) vary with full URL (which contains the version)

But you're right. If a new version is released in the meantime, the PR could be invalid.

I wonder if the the build should be reproducible (hard coded version) or accurate (dynamically fetch last version).

If you prefer staying reproductible, I'll have a look to the BOT (we talked here https://github.com/async-aws/aws/issues/258) to fetch last version and create a new PR to update the version WDYT?

Nyholm:
> the call to packagist (to fetch last version) is cached for 5 minutes

Good. 

Reproducible is fine. I just wanted to make sure that recurring contributors dont get error because AWS released new versions. 

Im happy to merge

jderusse:
Use hard coded version

PR #262, opened 2020-03-02
Periodically check for new updates

jderusse:
see example of build:
- https://github.com/jderusse/test-create-pr/runs/479244546?check_suite_focus=true

and generated PR
- https://github.com/jderusse/test-create-pr/pull/1/files

reflecting : 
https://github.com/aws/aws-sdk-php/blame/6f41bb144c415e89726eaad2ccdf4d0a54d6e7cf/src/data/lambda/2015-03-31/api-2.json#L2616

### DIFF
--- a/.github/workflows/watch.yml
+++ b/.github/workflows/watch.yml
@@ -1,0 +1,50 @@
+name: Watch for changes
+
+on:
+  schedule:
+  - cron: '0 6 * * *'
+
+jobs:
+  generated-code:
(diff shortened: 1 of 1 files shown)


jderusse:
It creates a PR only if there is a change, and update the PR when needed.
(I don't know if it's able to rebase, detect conflict, close when change manually pushed to master, ...) Let's figure it out when the case happen

Nyholm:
Okey. Awesome

PR #271, opened 2020-03-04
Use SSH key to PUSH bot PR

### LINKED ISSUE #270: PR created nightly don't triggers check
more information: https://github.com/peter-evans/create-pull-request/issues/48#issuecomment-536184102

jderusse:
Fixes #270 by using a deploy SSH key that will trigger JOB listening on "PUSH" events.

An other solution is to use a personal access token but is not ideal (security concerns)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,10 +1,7 @@
 name: Checks
 
-on:
-  push:
-    branches:
-      - master
-  pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/checks.yml
```diff
@@ -1,10 +1,7 @@
 name: Checks
 
-on:
-  push:
-    branches:
-      - master
```
Nyholm: Oh, that is right, We only run on pushes to master.

PR #274, opened 2020-03-05
Fix configuration not fallback to default

jderusse:
```php
Configuration::create(['endpoint' => null]);
```
should fallback to default value.


Nyholm:
:+1:

I'll add a test for this to show that this is the intended behavior.

PR #275, opened 2020-03-06
Added small tests for default values

Nyholm:
Related to #274

### DIFF
--- a/src/Core/tests/Unit/ConfigurationTest.php
+++ b/src/Core/tests/Unit/ConfigurationTest.php
@@ -12,27 +12,27 @@ class ConfigurationTest extends TestCase
     public function testDefaultValues()
     {
         $config = Configuration::create(['endpoint' => 'foo']);
-        $this->assertEquals('foo', $config->get('endpoint'));
+        self::assertEquals('foo', $config->get('endpoint'));
 
         $config = Configuration::create(['endpoint' => '']);
-        $this->assertEquals('', $config->get('endpoint'));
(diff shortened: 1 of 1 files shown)


src/Core/tests/Unit/ConfigurationTest.php
```diff
@@ -0,0 +1,25 @@
...
+    public function testDefaultValues()
+    {
+        $config = Configuration::create(['endpoint' => 'foo']);
+        $this->assertEquals('foo', $config->get('endpoint'));
+        $config = Configuration::create(['endpoint' => '']);
+        $this->assertEquals('', $config->get('endpoint'));
+
+        $config = Configuration::create(['endpoint' => null]);
+        $this->assertEquals('https://%service%.%region%.amazonaws.com', $config->get('endpoint'));
+    }
```
jderusse: Can we add a test about fallback from env variable then default?

```
$config = Configuration::create([]);
$this->assertEquals('https://%service%.%region%.amazonaws.com', $config->get('endpoint'));

$config = Configuration::create([]);
$this->assertEquals('default', $config->get('profile'));

putenv('AWS_PROFILE=foo');
$config = Configuration::create([]);
$this->assertEquals('foo', $config->get('profile'));

putenv('AWS_PROFILE');
$config = Configuration::create([]);
$this->assertEquals('default', $config->get('profile'));
```

Nyholm: Sure.

PR #286, opened 2020-03-09
Adding ResultMockFactory

Nyholm:
Im not sure this is a good idea. But it might be helpful when testing. 

```php
        /** @var AssumedRoleUser $result */
        $result = ResultMockFactory::create(AssumedRoleUser::class, [
            'Arn'=>'arn123',
            'AssumedRoleId'=>'foo123'
        ]);
```

You can also skip some (or all) mandatory parameters: 

```php
        /** @var AssumedRoleUser $result */
        $result = ResultMockFactory::create(AssumedRoleUser::class);
```

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6,14 +6,17 @@ parameters:
 			path: src/CodeGenerator/src/Command/GenerateCommand.php
 
 		-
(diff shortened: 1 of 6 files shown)


Nyholm:
Is this helpful?
Is this a good idea?

Nyholm:
Thank you. It is all yours

PR #287, opened 2020-03-09
Added interfaces for AbstractApi and Result

### LINKED ISSUE #288: Remove AbstractApi::request()
This function is only used if you want to request an endpoint not implemented. Is this needed? 

On the other hand... it is only 2 lines of code..

Nyholm:
This will make it easier to mock.

I took the final public function of Result and AbstractApi and added those to the interface. 

This will also fix #288

### DIFF
--- a/src/CodeGenerator/src/Generator/OperationGenerator.php
+++ b/src/CodeGenerator/src/Generator/OperationGenerator.php
@@ -11,6 +11,7 @@ use AsyncAws\CodeGenerator\Generator\Naming\ClassName;
 use AsyncAws\CodeGenerator\Generator\Naming\NamespaceRegistry;
 use AsyncAws\CodeGenerator\Generator\PhpGenerator\ClassFactory;
 use AsyncAws\Core\Result;
+use AsyncAws\Core\ResultInterface;
 use Nette\PhpGenerator\Method;
 
 /**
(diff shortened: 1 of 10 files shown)


Nyholm:
I noticed that dep-lowest was failing. I updated branch alias on core and sqs require section (since SQS was failing)

src/CodeGenerator/src/Generator/OperationGenerator.php
```diff
@@ -100,21 +101,21 @@ public function generate(Operation $operation): void
         }
 
         if (null !== $operation->getOutput()) {
-            $resutClass = $this->resultGenerator->generate($operation);
```
jderusse: good catch

Nyholm: We dont need to `git blame` this. I know it was me who did this typo. I make typos everywhere =)

src/Core/src/AbstractApi.php
```diff
@@ -76,16 +77,18 @@ public function __construct($configuration = [], ?CredentialProvider $credential
         ]));
     }
 
+    /**
+     * {@inheritdoc}
```
jderusse: I always wondered what's the purpose of this. I always thougth that IDE will fallback to parent docblock.

src/CodeGenerator/src/Generator/OperationGenerator.php
jderusse: sure about that? ;-) https://github.com/async-aws/aws/commit/2c3a63a7dccead9b4fe6921227006dab210b159c#diff-a69fae827839ebd310686b9067fc4518R97

Nyholm: oh =)

jderusse:
At first I thought adding an interface would help to create mock/proxy for testing (because methods are final).
But I was wrong: mocking a `ClientInterface` brings nothing to the tested Service. It will probably expect a FooClient to performs `$fooClient->MethodBar()` which is not covered by the interface.
Mocking the `ResultInterface` does not helps neither, because `MethodBar` of `FooClient` will returns a `BazResult`. Mocking thea method to returns a mock of `ResultInterface` will not be allowed.

Nyholm:
Thank you for giving this some thought and testing. I'll close this PR.

PR #290, opened 2020-03-09
Fix Host and bad Method call

jderusse:
Fix Port missing in host header
Fix transfert always chunked under alpine while it's not supported on fakesqs docker image

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -16,6 +16,7 @@ use AsyncAws\Core\Signer\Request;
 use AsyncAws\Core\Signer\Signer;
 use AsyncAws\Core\Signer\SignerV4;
 use AsyncAws\Core\Stream\StreamFactory;
+use AsyncAws\Core\Stream\StringStream;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Symfony\Component\HttpClient\HttpClient;
@@ -113,11 +114,18 @@ abstract class AbstractApi
         $request = new Request($method, $this->fillEndpoint($endpoint), $headers, $stream);
         $this->getSigner()->sign($request, $this->credentialProvider->getCredentials($this->configuration));
 
-        if (!$request->hasHeader('content-length') && null !== $length = $request->getBody()->length()) {
+        $length = $request->getBody()->length();
+        if (null !== $length && !$request->hasHeader('content-length')) {
             $request->setHeader('content-length', $length);
         }
 
-        return $this->httpClient->request($request->getMethod(), $request->getUrl(), ['headers' => $request->getHeaders(), 'body' => $request->getBody()]);
+        // Some servers (like testing Docker Images) does not supports `Transfer-Encoding: chunked` requests.
+        // The body is converted into string to prevent curl using `Transfer-Encoding: chunked` unless it really has to.
+        if (($requestBody = $request->getBody()) instanceof StringStream) {
+            $requestBody = $requestBody->stringify();
+        }
+
+        return $this->httpClient->request($request->getMethod(), $request->getUrl(), ['headers' => $request->getHeaders(), 'body' => 0 === $length ? null : $requestBody]);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Service/S3/tests/Integration/S3ClientTest.php
```diff
@@ -230,8 +230,8 @@ public function testHeadObject(): void
 
         $result->resolve();
 
-        self::assertSame('0', $result->getContentLength());
-        self::assertEquals('"d41d8cd98f00b204e9800998ecf8427e"', $result->getETag());
+        self::assertSame('7', $result->getContentLength());
```
jderusse: bug spotted

src/Core/src/AbstractApi.php
```diff
@@ -117,7 +118,11 @@ public function __construct($configuration = [], ?CredentialProvider $credential
             $request->setHeader('content-length', $length);
         }
 
-        return $this->httpClient->request($request->getMethod(), $request->getUrl(), ['headers' => $request->getHeaders(), 'body' => $request->getBody()]);
+        if (($body = $request->getBody()) instanceof StringStream) {
+            $body = $body->stringify();
+        }
```
Nyholm: So if the $body is a CallabeStream... then we pass that object to the `httpClient`, does that work?

Nyholm: Oh, sorry. My bad. 

A `Stream` is a `IteratorAggregate` which is Traversable. https://3v4l.org/Og6Bb

jderusse: Yes, because a stream is an `IteratorAggregate` which is handled by `HttpClient`: It yield on the iterator and send chunks. 

Issue is: some Docker Image does not support receiving chunked content. We didn't noticed it before, because it looks like php-curl sometime does not send the in chunks...

Please don't merge it yet, I'll try to use `transfert-encoding: identity`

Nyholm: > Issue is: some Docker Image does not support receiving chunked content. 

Please add that as a comment =)

jderusse: ouch.. in fact, there were a huge bug...

Because of Iterable body, `curl` converts the HEAD/GET method in PUT which break Rest APIs  

Please have a second review.

Nyholm: Why does curl do that?

src/Core/src/AbstractApi.php
```diff
@@ -113,11 +114,17 @@ public function __construct($configuration = [], ?CredentialProvider $credential
         $request = new Request($method, $this->fillEndpoint($endpoint), $headers, $stream);
         $this->getSigner()->sign($request, $this->credentialProvider->getCredentials($this->configuration));
 
-        if (!$request->hasHeader('content-length') && null !== $length = $request->getBody()->length()) {
+        if (null !== ($length = $request->getBody()->length()) && !$request->hasHeader('content-length')) {
```
Nyholm: Could you please write this in a more readable way?

src/Core/src/AbstractApi.php
```diff
@@ -113,11 +114,17 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
             $request->setHeader('content-length', $length);
         }
 
-        return $this->httpClient->request($request->getMethod(), $request->getUrl(), ['headers' => $request->getHeaders(), 'body' => $request->getBody()]);
+        // Some servers (like testing Docker Images) does not supports `Transfer-Encoding: chunked` requests.
+        // The body is converted into string to prevent curl using `Transfer-Encoding: chunked`
```
Nyholm: > The body is converted into string to prevent curl using `Transfer-Encoding: chunked` unless it really has to.

src/Core/src/AbstractApi.php
jderusse: Not sure, probably related to the way symfony/httpclient set the request body (`CURLOPT_READFUNCTION`)

src/Core/src/AbstractApi.php
jderusse: done. Thanks

jderusse:
LowesDep test fails because this PR fixes the bug

PR #291, opened 2020-03-09
Do not generate body when not needed

jderusse:


src/CodeGenerator/src/Definition/Operation.php
```diff
@@ -111,8 +111,17 @@ public function getHttpRequestUri(): ?string
...
+        if (isset($this->data['input']['method'])) {
+            throw new \InvalidArgumentException(sprintf('The operation "%s" should have an HTTP Method.', $this->getName()));
+        }
+
+        return $this->data['http']['method'];
+    }
+
+    public function hasBody(): bool
+    {
+        return \in_array($this->getHttpMethod(), ['PUT', 'POST']);
```
Nyholm: FYI: This is not always true for http request. All methods except OPTIONS may have a body if I remember correctly. 

 It might be true for all request we are doing.

jderusse: The generator should throw an exception in that case

Nyholm: Awesome

:+1:

PR #294, opened 2020-03-11
Update composer.json

swiffer:


Nyholm:
Thank you. Could you do the same thing with the composer.json in the root?

src/Integration/Flysystem/S3/tests/Integration/S3FilesystemV2Test.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace AsyncAws\Flysystem\S3\Tests;
+namespace AsyncAws\Flysystem\S3\Tests\Integration;
```
Nyholm: Thank you

PR #299, opened 2020-03-11
Remove fillEndpoint method from AbsreactApi

jderusse:
Now code is 100% generated, we now that `getResponse` is allays called with right arguments.
No need for such methods with optional arguments

I wonder if I shouldn't go further:
```php
final protected function getResponse(Request $request): ResponseInterface
{
    $request->setEndpoint($this->getEndpoint($request->uri(), $request->query());

    $this->getSigner()->sign($request, $this->credentialProvider->getCredentials($this->configuration));
    // ...
}


public function foo($input): Result
{
    $input = FooInput::create($input);
    $input->validate();

    $response = $this->getResponse(
        $input->request()
    );

    return new FooResult($response, $this->httpClient);
}
```

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -92,15 +92,11 @@ abstract class AbstractApi
      * @param string[]|string[][]                    $headers headers names provided as keys or as part of values
      * @param string|resource|callable|iterable|null $body
      */
-    final protected function getResponse(string $method, $body, $headers = [], ?string $endpoint = null): ResponseInterface
+    final protected function getResponse(string $method, $body, array $headers, string $endpoint): ResponseInterface
     {
-        if (!isset($headers['content-type'])) {
-            $headers['content-type'] = 'text/plain';
-        }
-
         $stream = StreamFactory::create($body);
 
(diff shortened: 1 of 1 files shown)


src/Core/src/AbstractApi.php
```diff
@@ -162,15 +161,4 @@ private function getSigner()
 
         return $this->signer;
     }
-
-    private function fillEndpoint(?string $endpoint): string
```
Nyholm: I think this was here for a reason.. 
Ie S3 could have a simpler `getEndpoint` if it didnt needed to bother about the region..

Nyholm:
"going further". That means that `Request` is no longer internal for `core`. (just an observation)

Hm.. I do like the idea though..

src/Core/src/AbstractApi.php
jderusse: Take your time to review, no hurry. I thougth it was here because of the generic AwsClient

Because it's currently not used, I prefer, removing it and re-adding it later when the use case is clear.

Nyholm:
I'll create a PR about "goign further"

PR #310, opened 2020-03-14
Test entire request

### LINKED ISSUE #252: Rest Input/Result test are not just about body
The automated test generate stubs for Body, but in RestJson and RestXml, Uri and Headers are important too. 

see for instance https://github.com/async-aws/aws/blob/master/src/Service/Lambda/tests/Unit/Input/AddLayerVersionPermissionRequestTest.php

jderusse:
Fixes #252

Better review whith whitespace hidden

### DIFF
--- a/src/Service/S3/src/Input/PutObjectRequest.php
+++ b/src/Service/S3/src/Input/PutObjectRequest.php
@@ -519,7 +519,7 @@ class PutObjectRequest
     public function request(): Request
     {
         // Prepare headers
-        $headers = ['content-type' => 'application/xml'];
+        $headers = [];
         if (null !== $this->ACL) {
             $headers['x-amz-acl'] = $this->ACL;
         }
--- a/src/Service/S3/tests/Unit/Input/PutObjectRequestTest.php
+++ b/src/Service/S3/tests/Unit/Input/PutObjectRequestTest.php
@@ -7,20 +7,20 @@ use AsyncAws\S3\Input\PutObjectRequest;
 
 class PutObjectRequestTest extends TestCase
 {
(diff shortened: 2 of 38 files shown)


src/Service/S3/src/Input/PutObjectRequest.php
```diff
@@ -519,7 +519,7 @@ public function getWebsiteRedirectLocation(): ?string
     public function request(): Request
     {
         // Prepare headers
-        $headers = ['content-type' => 'application/xml'];
+        $headers = [];
```
jderusse: That's a bug in the current implementation

src/Service/S3/tests/Unit/Input/PutObjectRequestTest.php
```diff
@@ -7,7 +7,7 @@
 
 class PutObjectRequestTest extends TestCase
 {
-    public function testRequestBody()
+    public function testRequest()
     {
         $bucket = 'foo';
         $key = 'bar/baz/biz';
```
Nyholm: These variables are only used once. Are they auto generated?

jderusse: nop, did you try `git blame`?..

Nyholm: They were used twice but this PR removed once instance. 

Im fine with this, I just wanted to make sure we dont generate variables that are only used once.

jderusse: ohh. my bad. Removed

PR #312, opened 2020-03-14
Make sure Request::$header is an array

Nyholm:


Nyholm:
Hm.. this is already set in the constructor...

But I get this error: https://github.com/async-aws/s3/runs/508372874

jderusse:
Because of https://github.com/async-aws/aws/pull/310/files#diff-bbfbf1f42de38b8d7d7988ae66920bdfR36

Nyholm:
Ah.. You added it just recently. 

Good =)

I was super confused how `$headers` could be null. =)

PR #314, opened 2020-03-15
Init DynamoDb

### LINKED ISSUE #302: AWS DynamoDB support
Like S3, DynamoDB is one of the core services of AWS. It's a serverless document database. It would be great to have it in this library.

Documentation: https://docs.aws.amazon.com/aws-sdk-php/v2/guide/service-dynamodb.html

There is a few methods to implement:

- [ ] BatchGetItem
- [ ] BatchWriteItem
- [x] CreateTable
- [x] DeleteItem
- [x] DeleteTable
- [x] DescribeTable
- [x] GetItem
- [x] ListTables
- [x] PutItem
- [x] Query
- [x] Scan
- [x] UpdateItem
- [x] UpdateTable

Generic use-cases:
* Session handler https://docs.aws.amazon.com/aws-sdk-php/v2/guide/feature-dynamodb-session-handler.html
* Distributed lock https://aws.amazon.com/blogs/database/building-distributed-locks-with-the-dynamodb-lock-client/
* Sémaphore https://boylesoftware.com/blog/convenient-semaphore-dynamodb/
* Others https://docs.aws.amazon.com/aws-sdk-php/v2/guide/service-dynamodb.html

Nyholm:
This is related to #302

### DIFF
--- a/src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
+++ b/src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\CodeGenerator\Generator\RequestSerializer;
+
+use AsyncAws\CodeGenerator\Definition\Operation;
+use AsyncAws\CodeGenerator\Definition\Shape;
+
+/**
+ * Serialize to AWS json.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ *
+ * @internal
+ */
+class JsonSerializer extends RestJsonSerializer
+{
+    public function getHeaders(Operation $operation): string
+    {
+        return strtr(<<<PHP
+[
+    'Content-Type' => 'application/x-amz-json-1.0',
+    'X-Amz-Target' => TARGET,
+];
+
+PHP,
+        [
+            'TARGET' => sprintf('"%s.%s"', $operation->getService()->getTargetPrefix(), $operation->getName()),
+        ]);
+    }
+
+    protected function dumpArrayBoolean(string $output, string $input, Shape $shape): string
+    {
+        return strtr('$payloadOUTPUT = (bool) INPUT;', [
+            'OUTPUT' => $output,
+            'INPUT' => $input,
+        ]);
+    }
+}
--- a/src/CodeGenerator/src/Generator/ResponseParser/JsonParser.php
+++ b/src/CodeGenerator/src/Generator/ResponseParser/JsonParser.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\CodeGenerator\Generator\ResponseParser;
+
+use AsyncAws\CodeGenerator\Definition\Shape;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ *
+ * @internal
+ */
+class JsonParser extends RestJsonParser
+{
+    protected function parseResponseTimestamp(Shape $shape, string $input, bool $required): string
+    {
+        $body = '\DateTimeImmutable::createFromFormat(\'U.u\', \sprintf(\'%.6F\', INPUT))';
+
+        if (!$required) {
+            $body = 'isset(INPUT) ? ' . $body . ' : null';
+        }
+
+        return strtr($body, ['INPUT' => $input]);
+    }
+}
--- a/src/CodeGenerator/src/Generator/ResponseParser/JsonParser.php
+++ b/src/CodeGenerator/src/Generator/ResponseParser/JsonParser.php
@@ -15,10 +15,10 @@ class JsonParser extends RestJsonParser
 {
     protected function parseResponseTimestamp(Shape $shape, string $input, bool $required): string
     {
-        $body = '\DateTimeImmutable::createFromFormat(\'U.u\', \sprintf(\'%.6F\', INPUT))';
-
-        if (!$required) {
-            $body = 'isset(INPUT) ? ' . $body . ' : null';
+        if ($required) {
+            $body = '/** @var \DateTimeImmutable $d */ $d = \DateTimeImmutable::createFromFormat(\'U.u\', \sprintf(\'%.6F\', INPUT))';
+        } else {
+            $body = '(isset(INPUT) && ($d = \DateTimeImmutable::createFromFormat(\'U.u\', \sprintf(\'%.6F\', INPUT)))) ? $d : null';
         }
 
         return strtr($body, ['INPUT' => $input]);
--- a/src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
+++ b/src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
@@ -130,12 +130,20 @@ class RestJsonParser implements Parser
             ]);
         }
 
-        return strtr('new CLASS_NAME([
+        $body = 'new CLASS_NAME([
             PROPERTIES
-        ])', [
-            'CLASS_NAME' => $this->namespaceRegistry->getObject($shape)->getName(),
-            'PROPERTIES' => implode("\n", $properties),
-        ]);
+        ])';
+
+        if (!$required) {
+            $body = 'empty(INPUT) ? null : ' . $body;
+        }
+
+        return strtr(
+            $body, [
+                'INPUT' => $input,
+                'CLASS_NAME' => $this->namespaceRegistry->getObject($shape)->getName(),
+                'PROPERTIES' => implode("\n", $properties),
+            ]);
     }
 
     private function parseResponseString(string $input, bool $required): string
--- a/src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
+++ b/src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
@@ -60,6 +60,17 @@ class RestJsonParser implements Parser
         return $body;
     }
 
+    protected function parseResponseTimestamp(Shape $shape, string $input, bool $required): string
+    {
+        $body = 'new \DateTimeImmutable((string) INPUT)';
+
+        if (!$required) {
+            $body = 'isset(INPUT) ? ' . $body . ' : null';
+        }
+
+        return strtr($body, ['INPUT' => $input]);
+    }
+
     private function getInputAccessorName(Member $member)
     {
         if ($member instanceof StructureMember) {
@@ -172,23 +183,6 @@ class RestJsonParser implements Parser
         return strtr('isset(INPUT) ? base64_decode((string) INPUT) : null', ['INPUT' => $input]);
     }
 
-    private function parseResponseTimestamp(Shape $shape, string $input, bool $required): string
-    {
-        if ('unixTimestamp' === $shape->get('timestampFormat')) {
-            if ($required) {
-                return strtr('\DateTimeImmutable::setTimestamp((string) INPUT)', ['INPUT' => $input]);
-            }
-
-            return strtr('isset(INPUT) ? \DateTimeImmutable::setTimestamp((string) INPUT) : null', ['INPUT' => $input]);
-        }
-
-        if ($required) {
-            return strtr('new \DateTimeImmutable((string) INPUT)', ['INPUT' => $input]);
-        }
-
-        return strtr('isset(INPUT) ? new \DateTimeImmutable((string) INPUT) : null', ['INPUT' => $input]);
-    }
-
     private function parseResponseList(ListShape $shape, string $input, bool $required): string
     {
         $shapeMember = $shape->getMember();
@@ -218,7 +212,7 @@ class RestJsonParser implements Parser
         }
 
         if (!$required) {
-            $body = '!INPUT ? [] : ' . $body;
+            $body = 'empty(INPUT) ? [] : ' . $body;
         }
 
         return strtr($body, [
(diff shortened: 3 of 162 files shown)


Nyholm:
This is a little tricky. Im happy to get some help.

Looking at: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DeleteItem.html#API_DeleteItem_Examples

There is an shape named `AttributeValue` which may contain a list of itself. (https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_AttributeValue.html)

src/Service/DynamoDb/src/Result/QueryOutput.php
```diff
@@ -0,0 +1,214 @@
...
+            throw new \InvalidArgumentException('missing client injected in paginated result');
+        }
+        if (!$this->input instanceof QueryInput) {
+            throw new \InvalidArgumentException('missing last request injected in paginated result');
+        }
+        $input = clone $this->input;
+        $page = $this;
+        while (true) {
+            if ($page->getLastEvaluatedKey()) {
+                $input->setExclusiveStartKey($page->getLastEvaluatedKey());
```
Nyholm: This is similar to what @jderusse and I talked about the other day, 

> Argument 1 of AsyncAws\DynamoDb\Input\QueryInput::setExclusiveStartKey expects array<array-key, AsyncAws\DynamoDb\Input\AttributeValue>, array<array-key, AsyncAws\DynamoDb\Result\AttributeValue> provided

src/Service/DynamoDb/src/Input/CreateTableInput.php
```diff
@@ -0,0 +1,511 @@
...
+                            }
+
+                            $x = (static function (array $input) use (&$payload, $indices) {
+                                $indices->kefae06f = -1;
+                                foreach ($input as $value) {
+                                    ++$indices->kefae06f;
+                                    $payload['LocalSecondaryIndexes'][$indices->kad3ea0f]['Projection']['NonKeyAttributes'][$indices->kefae06f] = $value;
+                                }
+                            })($input->getNonKeyAttributes());
+                        })($input->getProjection());
```
Nyholm: I dont understand this: 

> Argument 1 cannot be null, possibly null value provided

We check for null on 399

src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
```diff
@@ -0,0 +1,306 @@
...
+
+        return strtr($methodBody, $replaceData);
+    }
+
+    private function dumpArrayList(string $output, string $input, ListShape $shape): string
+    {
+        $memberShape = $shape->getMember()->getShape();
+
+        return strtr('
+$x = (static function(array $input) use (USE) {
```
Nyholm: This extra `$x = ` helps php-cs-fixer.

src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
```diff
@@ -0,0 +1,306 @@
...
+use AsyncAws\CodeGenerator\Generator\Naming\NamespaceRegistry;
+
+/**
+ * Serialize a request body to a nice nested array.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ *
+ * @internal
+ */
+class JsonSerializer implements Serializer
```
Nyholm: I copied the RestJsonSerializer. We can probably merge them later. But I'll do that in a separate PR.

src/Service/DynamoDb/src/Result/QueryOutput.php
jderusse: should be fixed by #316

src/Service/DynamoDb/src/Input/CreateTableInput.php
jderusse: maybe check on null value does not works on methods..

```php
public function getFoo(): ?string
{
  return rand(0,1) === 0 ? null : 'foo';
}
```

and psalm track value on variables only

Nyholm: True. 
I can add a variable.

src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
```diff
@@ -0,0 +1,303 @@
...
+{
+    private $recursionName = null;
+
+    private $recursionInitialized = false;
```
jderusse: Is there a way to avoid state?

What's about adding `requestBody()` in all objects (and not only in Root Input object), Like we do in `Validate`.
That would reduce complexity of the method in root object, and easily fix the recursion thing.

```php
public function requestBody()
{
  $payload = [];
  $payload['LocalProperty'] = $this->LocalProperty;
  if (null !== $this->ObjectProperty) {
     $payload['ObjectProperty'] = $this->ObjectProperty->requestBody();
  }
}
```

Nyholm: I thought about that. But I decided not to when I considered xml requests. It will be a bit more complicated but it may work.

jderusse: Your right, it could be done in another PR.

Xml should be easy:
```php
$doc = $document->createElement('Foo');
$doc->appendChild($this->Bar->requestBody()));
```

But, for Query Serializer that could be hard (for the record, the return array is flat: `$payload['Foo.Bar.0.member']`)

Nyholm:
Current status: Segmentation fault

src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
jderusse: should be fixed by #318

Nyholm:
We get stuck in an infinite loop here: 

https://github.com/async-aws/aws/blob/master/src/CodeGenerator/src/Generator/ObjectGenerator.php#L106-L134

src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
Nyholm: I do not create a separate Serializer now.

src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
Nyholm: Thank you

Nyholm:
I think Im pretty done. I just need to add tests to confirm

src/CodeGenerator/src/Generator/RequestSerializer/JsonSerializer.php
```diff
@@ -157,6 +157,22 @@ private function dumpArrayList(string $output, string $input, ListShape $shape):
             ]);
     }
 
+    private function dumpArrayMap(string $output, string $input, MapShape $shape): string
+    {
+        $memberCode = $this->dumpArrayElement($output . '[$name]', '$value', $shape->getValue()->getShape());
+
+        return strtr('
+if (is_array(INPUT)) {
```
jderusse: I think it should be the case. Constructors replaces null by array

src/CodeGenerator/src/Generator/ResponseParser/JsonParser.php
```diff
@@ -230,11 +230,26 @@ private function parseResponseList(ListShape $shape, string $input, bool $requir
 
     private function parseResponseMap(MapShape $shape, string $input, bool $required): string
     {
+        $shapeValue = $shape->getValue();
+
         if (null === $locationName = $shape->getKey()->getLocationName()) {
-            throw new \RuntimeException('This is not implemented yet');
+            // This is a map
+
+            $body = '(function(array $json): array {
```
jderusse: missing code when the input is not required (see bellow) 
```php
if (!$required) {
  $body = '!INPUT ? [] : ' . $body;
}
```

src/Service/DynamoDb/src/Enum/AttributeAction.php
```diff
@@ -0,0 +1,21 @@
...
+
+class AttributeAction
+{
+    public const ADD = 'ADD';
+    public const DELETE = 'DELETE';
+    public const PUT = 'PUT';
+
+    public static function exists(string $value): bool
+    {
+        $values = [
```
jderusse: couldn't it be `static $values` for permformance?

src/Service/DynamoDb/src/Input/CreateTableInput.php
```diff
@@ -0,0 +1,426 @@
...
+     *   BillingMode?: \AsyncAws\DynamoDb\Enum\BillingMode::*,
+     *   ProvisionedThroughput?: \AsyncAws\DynamoDb\ValueObject\ProvisionedThroughput|array,
+     *   StreamSpecification?: \AsyncAws\DynamoDb\ValueObject\StreamSpecification|array,
+     *   SSESpecification?: \AsyncAws\DynamoDb\ValueObject\SSESpecification|array,
+     *   Tags?: \AsyncAws\DynamoDb\ValueObject\Tag[],
+     * } $input
+     */
+    public function __construct(array $input = [])
+    {
+        $this->AttributeDefinitions = array_map(function ($item) { return AttributeDefinition::create($item); }, $input['AttributeDefinitions'] ?? []);
```
jderusse: Isn't it faster ?:
```diff
- array_map(function ($item) { return AttributeDefinition::create($item); }, $input['AttributeDefinitions'] ?? []);
+ array_map([AttributeDefinition::class, 'create'], $input['AttributeDefinitions'] ?? []);
```

src/Service/DynamoDb/src/Input/CreateTableInput.php
```diff
@@ -0,0 +1,426 @@
...
+     * @internal
+     */
+    private function requestBody(): array
+    {
+        $payload = [];
+
+        $index = -1;
+        foreach ($this->AttributeDefinitions as $mapValue) {
+            ++$index;
+            if (null !== $v = $mapValue) {
```
jderusse: thta's weird, it should not be null. Is it?

src/Service/DynamoDb/src/Enum/AttributeAction.php
Nyholm: Im not sure. It leaving that for a future update. 
I think it is just minor improvements though..

src/Service/DynamoDb/src/Input/CreateTableInput.php
Nyholm: This is not my change. But I reverted it now.

src/Service/DynamoDb/src/Input/CreateTableInput.php
Nyholm: Why, isnt it the exact same thing but different syntax?

jderusse: yes, it's equivalent, just a small performance optimization.
By wraping the Factory in a closure, we performs 2 methods call, when 1 is enougth.

Nyholm:
I saw from the tests/examples that there is a difference between rest-json and rest. 

I created a new serializer. Currently I've only found differences how one use booleans and the content type.

src/Service/DynamoDb/src/Input/CreateTableInput.php
jderusse: fixed by #322

src/CodeGenerator/src/Generator/ResponseParser/JsonParser.php
```diff
@@ -230,11 +230,30 @@ private function parseResponseList(ListShape $shape, string $input, bool $requir
...
+                $items = [];
+                foreach ($json as $name => $value) {
+                    $items[$name] = CLASS::create($value);
```
jderusse: Hmm here, you presume that this is a Map of struct, which is not always the case. For instances, `Tags` are often a map of strings.

Looks at parseResponseList (which is very close to map, expect array's Key)

src/Service/DynamoDb/src/Result/QueryOutput.php
```diff
@@ -0,0 +1,222 @@
...
+    {
+        $this->initialize();
+
+        return $this->Count;
+    }
+
+    /**
+     * @param bool $currentPageOnly When true, iterates over items of the current page. Otherwise also fetch items in the next pages.
+     *
+     * @return iterable<AttributeValue[]>
```
Nyholm: Isnt this a super weird return type?

Nyholm: It is correct though..

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
```diff
@@ -259,12 +277,12 @@ private function parseResponseMap(MapShape $shape, string $input, bool $required
         }
 
         if (!$required) {
-            $body = '!INPUT ? [] : ' . $body;
+            $body = '(!isset(INPUT) || !INPUT) ? [] : ' . $body;
```
Nyholm: If input is `$array['foo']['bar']` we need to check if `"foo"` exists.

jderusse: `(!isset(INPUT) || !INPUT)` => `empty(INPUT)` ?

src/CodeGenerator/src/Generator/ResponseParser/JsonParser.php
Nyholm: Thank you

Nyholm:
PHPStan issues are because: (simplified)

```
Expected: 

AttributeDefinitions?: array<array-key, AsyncAws\DynamoDb\ValueObject\AttributeDefinition>|null, 
CreationDateTime?: DateTimeInterface|null, 
GlobalSecondaryIndexes?: array<array-key, AsyncAws\DynamoDb\ValueObject\GlobalSecondaryIndexDescription>|null, 
KeySchema?: array<array-key, AsyncAws\DynamoDb\ValueObject\KeySchemaElement>|null, 
LocalSecondaryIndexes?: array<array-key, AsyncAws\DynamoDb\ValueObject\LocalSecondaryIndexDescription>|null, 
Replicas?: array<array-key, AsyncAws\DynamoDb\ValueObject\ReplicaDescription>|null, 
TableStatus?: null|string(ACTIVE)|string(ARCHIVED)|string(ARCHIVING)|string(CREATING)|string(DELETING)|string(INACCESSIBLE_ENCRYPTION_CREDENTIALS)|string(UPDATING)


Provided:
AttributeDefinitions: list<AsyncAws\DynamoDb\ValueObject\AttributeDefinition>, 
CreationDateTime: DateTimeImmutable|false|null, 
GlobalSecondaryIndexes: list<AsyncAws\DynamoDb\ValueObject\GlobalSecondaryIndexDescription>, 
KeySchema: list<AsyncAws\DynamoDb\ValueObject\KeySchemaElement>, 
LocalSecondaryIndexes: list<AsyncAws\DynamoDb\ValueObject\LocalSecondaryIndexDescription>, 
Replicas: list<AsyncAws\DynamoDb\ValueObject\ReplicaDescription>, 
TableStatus: null|string
````

Im not 100% about the lists. But I think they are fine. Which means: 


```
Expected: 
CreationDateTime?: DateTimeInterface|null, 


Provided:
CreationDateTime: DateTimeImmutable|false|null, 
````

This is because we use `\DateTimeImmutable::createFromFormat()` which may return false.  
Any points how to proceed?

Nyholm:
Im finally done with this PR. 

Thank you for all the reviews and other PRs that helped me complete this. 

@GromNaN, I thought I could make this PR in 10 minutes, I didn't realise it was this much work when I asked you to do it. ¯\_(ツ)_/¯

GromNaN:
Thank you for your dedication. I'm going to try.

PR #316, opened 2020-03-15
Move ValueObject in dedicated Namespace

jderusse:
This PR move ValueObject used by both Input and Result in a shared namespace.

The Object is immutable. Setter have been removed.
It allows invalid input, but  calling getter on it will throw an exception because of typehint. (WDYT?)

This PR allowed the following input:

```php
$foo = new Foo(['required' => 1, 'optional' => null]);

$foo = new Foo(['required' => null]); // let user input invalid data. It will be check by client with `$input->valide()`
$foo->getRequired(); // failed, becase return typehint does not allow null

$foo = new Foo([]); // let user input invalid data. It will be check by client with `$input->valide()`
$foo->getRequired(); // failed, becase return typehint does not allow null
```


src/Core/src/Sts/ValueObject/PolicyDescriptorType.php
```diff
@@ -1,25 +1,17 @@
 <?php
 
-namespace AsyncAws\Core\Sts\Input;
+namespace AsyncAws\Core\Sts\ValueObject;
 
 class PolicyDescriptorType
 {
-    /**
-     * The Amazon Resource Name (ARN) of the IAM managed policy to use as a session policy for the role. For more
-     * information about ARNs, see Amazon Resource Names (ARNs) and AWS Service Namespaces in the *AWS General Reference*.
```
Nyholm: Doesn’t it make sense to have comments on properties now?

jderusse: Given the object is used in response, the docblock should at least be on getters.
Otherwise, user will loose IDE autocompletion.

Screenshot from 2020-03-16 09-58-41

Or we could dupplicate dockbloc on each property/setters/getters

jderusse:
I don't know if removing setters will be an issue..

Actually, we provide Several way to do it.
```php
// Full array
[
    'AccessControlPolicy' => [
        'Grants' => [
            [
                'Grantee' => [
                    'DisplayName' => 'me',
                    'Type' => Type::CANONICAL_USER,
                ],
                'Permission' => Permission::FULL_CONTROL,
            ]
        ],
        'Owner' => [
            'DisplayName' => 'me',
        ],
    ],
];
```

```php
// Array with constructors
[
    'AccessControlPolicy' => new AccessControlPolicy([
        'Grants' => [new Grant([
            'Grantee' => new Grantee([
                'DisplayName' => 'me',
                'Type' => Type::CANONICAL_USER,
            ]),
            'Permission' => Permission::FULL_CONTROL,
        ])],
        'Owner' => new Owner([
            'DisplayName' => 'me',
        ]),
    ]),
];
```

```php
// intermediate objects
$owner =  new Owner([
    'DisplayName' => 'me',
]);
$grantee =  new Grantee([
    'DisplayName' => 'me',
    'Type' => Type::CANONICAL_USER,
]);
$grant = new Grant([
    'Grantee' => $grantee,
    'Permission' => Permission::FULL_CONTROL,
]);
[
    'AccessControlPolicy' => new AccessControlPolicy([
        'Grants' => [$grant],
        'Owner' => $owner,
    ]),
];
```

src/Core/src/Sts/ValueObject/PolicyDescriptorType.php
Nyholm: Let's duplicate them for now. 
I believe they are needed for me to understand what I should do with the input.

Nyholm:
Okey. I see there is no change it the Grantee stuff. Thank you

jderusse:
In fact, there are one change: You're no able anymore to do

```php
function updateGrantee(Grant $grant) {
  $grantee =  new Grantee([
      'DisplayName' => 'me',
      'Type' => Type::CANONICAL_USER,
  ]);

  $grant->setGrantee($grantee);
}

src/Core/src/Sts/ValueObject/PolicyDescriptorType.php
jderusse: Changed my mind. When writing entities, I don't write docbloc on getters neither on setter. Using property is the right place...

I applyed the rule on Input, ValueObject and Result

Nyholm:
:+1:

Im happy with this PR. 

Good job.

jderusse:
Thanks

PR #318, opened 2020-03-17
Split Requestbody

jderusse:
Should helps with recurssion in #314

### DIFF
--- a/src/CodeGenerator/src/Generator/ObjectGenerator.php
+++ b/src/CodeGenerator/src/Generator/ObjectGenerator.php
@@ -6,11 +6,13 @@ namespace AsyncAws\CodeGenerator\Generator;
 
 use AsyncAws\CodeGenerator\Definition\ListShape;
 use AsyncAws\CodeGenerator\Definition\MapShape;
+use AsyncAws\CodeGenerator\Definition\Shape;
 use AsyncAws\CodeGenerator\Definition\StructureShape;
 use AsyncAws\CodeGenerator\File\FileWriter;
 use AsyncAws\CodeGenerator\Generator\CodeGenerator\TypeGenerator;
 use AsyncAws\CodeGenerator\Generator\Naming\ClassName;
 use AsyncAws\CodeGenerator\Generator\Naming\NamespaceRegistry;
+use AsyncAws\CodeGenerator\Generator\RequestSerializer\SerializerProvider;
 use AsyncAws\Core\StreamableBodyInterface;
 use Nette\PhpGenerator\ClassType;
 use Nette\PhpGenerator\PhpNamespace;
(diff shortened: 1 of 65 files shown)


src/Service/S3/src/Input/DeleteObjectsRequest.php
```diff
@@ -128,8 +128,15 @@ public function request(): Request
         $uri['Bucket'] = $this->Bucket ?? '';
         $uriString = "/{$uri['Bucket']}?delete";
 
+        // Prepare Body
+
+        $document = new \DOMDocument('1.0', 'UTF-8');
+        $document->formatOutput = false;
+        $this->requestBody($document, $document);
```
Nyholm: This looks super weird. But I see why you are doing it. Smart!

src/Core/src/Sts/Input/AssumeRoleRequest.php
```diff
@@ -319,55 +322,48 @@ public function validate(): void
...
-
-                if (null !== $value) {
-                    (static function (PolicyDescriptorType $input) use (&$payload, $indices) {
-                        if (null !== $v = $input->getarn()) {
-                            $payload["PolicyArns.member.{$indices->kfc822c1}.arn"] = $v;
-                        }
-                    })($value);
+        $index = 0;
+        foreach ($this->PolicyArns as $mapValue) {
+            ++$index;
```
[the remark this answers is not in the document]
jderusse: Yes.
It's weird, but this is what AWS use QueryString (this Sample Request in https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html)

Note that RestJson starts a 0

src/Service/Lambda/src/Input/PublishLayerVersionRequest.php
```diff
@@ -118,8 +118,12 @@ public function request(): Request
         $uri['LayerName'] = $this->LayerName ?? '';
         $uriString = "/2018-10-31/layers/{$uri['LayerName']}/versions";
 
+        // Prepare Body
+        $bodyPayload = $this->requestBody();
+        $body = empty($bodyPayload) ? '{}' : json_encode($bodyPayload);
```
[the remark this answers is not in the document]
jderusse: yes, AWS expects a json Object, while `json_encode([])` returns `[]`

src/Core/src/Sts/Input/AssumeRoleRequest.php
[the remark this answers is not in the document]
Nyholm: haha okey. Good. 

I know I've seen indexes starting at 0 before. 

:+1:

src/Service/S3/src/Input/DeleteObjectsRequest.php
jderusse: $document is both the root node (first argument), and node creator (second argument).

I think, I can avoid the first argument. Let me try something else

Nyholm: Isnt this also needed if we are doing requests to value objects' requestBody()?

jderusse: hmm no, nevermind, the name of the tag is not related to the `shape`, but the `member`.
Meaning, the same object could be inside a `<Foo>` tag then a `<Bar>` tag. The node can not be created by the object itself, it have to be created by the caller.

jderusse:
Thank you for the review

src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -76,11 +89,50 @@ public function generate(StructureShape $shape): ClassName
...
+                if ($shape instanceof StructureShape) {
+                    foreach ($shape->getMembers() as $member) {
+                        $walk($member->getShape());
+                    }
+                } elseif ($shape instanceof ListShape) {
+                    $walk($shape->getMember()->getShape());
+                } elseif ($shape instanceof MapShape) {
+                    $walk($shape->getValue()->getShape());
+                }
+            };
```
Nyholm: This will trap us in an infinite loop if structures contain eachother. I'll provide a test to show it.

jderusse: oh yes, I forgot the exit statement

```diff
            $walk = function (Shape $shape) use (&$walk) {
+               if (isset($this->usedShapedInput[$shape->getName()]) return;
                $this->usedShapedInput[$shape->getName()] = true;
                if ($shape instanceof StructureShape) {
                    foreach ($shape->getMembers() as $member) {
                        $walk($member->getShape());
                    }
                } elseif ($shape instanceof ListShape) {
                    $walk($shape->getMember()->getShape());
                } elseif ($shape instanceof MapShape) {
                    $walk($shape->getValue()->getShape());
                }
            };
```

PR #319, opened 2020-03-17
Prevent infinite loop

jderusse:


Nyholm:
I pushed to your PR to fix two more infinite loops

PR #323, opened 2020-03-17
Fix compatibility with FlySystem 2

jderusse:


src/Integration/Flysystem/S3/src/S3FilesystemV2.php
```diff
@@ -230,7 +230,7 @@ public function fileSize(string $path): FileAttributes
         return $this->fetchFileMetadata($path, FileAttributes::ATTRIBUTE_FILE_SIZE);
     }
 
-    public function listContents(string $path, bool $recursive): Generator
+    public function listContents(string $path, bool $recursive): iterable
```
jderusse: Introduced in alpha2

src/Integration/Flysystem/S3/tests/version-bridge-tests.php
```diff
@@ -17,3 +17,27 @@ class FilesystemAdapterTestCase extends TestCase
         }
     }
 }
+
+namespace {
+    if (!function_exists('reset_function_mocks')) {
```
jderusse: because of https://github.com/thephpleague/flysystem/commit/a5735c905f5f86c47b8073f75c33b5fd2023288d#r37884566

.github/workflows/ci.yml
```diff
@@ -125,7 +125,7 @@ jobs:
             echo "$CURRENT_DIR/$COMPONENT"
             cd "$CURRENT_DIR/$COMPONENT"
 
-            sed -i -re 's/"require": \{/"minimum-stability": "dev","prefer-stable": true,"require": \{/' composer.json
+            sed -i -re 's/"require": \{/"minimum-stability": "alpha","prefer-stable": true,"require": \{/' composer.json
```
jderusse: I don't get why prefer-lowest failed without this change on downloading `2.0.0-alpha.2`

Nyholm: Hm.. Neither do I.. 

But this change will fail when S3 is require a version of core that is not released yet. (When it is only present in branch alias)

.github/workflows/ci.yml
```diff
@@ -112,21 +112,32 @@ jobs:
...
+      - name: Modify composer.json
+        run: |
+          CURRENT_DIR=$(pwd)
+          for COMPONENT in $(find src/Service -maxdepth 2 -type f -name phpunit.xml.dist -printf '%h\n' | sort)
+          do
+            echo ::group::$COMPONENT
+            echo "$CURRENT_DIR/$COMPONENT"
+            cd "$CURRENT_DIR/$COMPONENT"
+            sed -i -re 's/"async-aws\/core": "[^"]+",/"async-aws\/core": "*@dev",/' composer.json
+            sed -i -re 's/"require": \{/"repositories": [{"type": "path","url": "..\/..\/Core"}],"require": \{/' composer.json
```
jderusse: That breaks the purpose of `prefer-lowest` don't you think?

Nyholm: Yes. The code have been updated.

PR #325, opened 2020-03-18
Remove dependency to http-client

jderusse:
This PR removes the coupling between Services and http-client by exposing a new `Response` object to client

PRO:
- code in Result and Waiter have been merged
- remove dependencies to symfony/http-client in each services
- AbstractApi does not contains protected properties

CON
- BC break for all services, that have to require `core: ^0.5`

WDYT?

### DIFF
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -1,0 +1,161 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Core;
+
+use AsyncAws\Core\Exception\Http\ClientException;
+use AsyncAws\Core\Exception\Http\HttpException;
+use AsyncAws\Core\Exception\Http\NetworkException;
+use AsyncAws\Core\Exception\Http\RedirectionException;
+use AsyncAws\Core\Exception\Http\ServerException;
+use AsyncAws\Core\Exception\RuntimeException;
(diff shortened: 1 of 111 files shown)


Nyholm:
I like this idea. Because it hides complexity. 

I'll review this more carefully later.

src/Core/src/Response.php
```diff
@@ -0,0 +1,157 @@
...
+use AsyncAws\Core\Exception\Http\ServerException;
+use AsyncAws\Core\Exception\RuntimeException;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+/**
+ * The response provides a facade to manipulate HttpResponses.
+ */
+class Response
```
Nyholm: We could add the `@final` annotation on this response and the request. 

That will show that one should not extend these classes, but one can still mock them properly.

Nyholm:
Could you rebase please?

src/Core/src/Response.php
jderusse: good idea. thanks

jderusse:
thanks

PR #328, opened 2020-03-18
Bump branch-alias versions

jderusse:


src/Integration/Flysystem/S3/composer.json
```diff
@@ -27,7 +27,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "0.3-dev"
+            "dev-master": "0.4-dev"
```
jderusse: Every package have a different version. That's makes the things hard to maintain. 

What's about using the same version everywhere? Even when nothing change between 2 versions?

Nyholm: We only have this issue now when we are breaking BC all the time. 

After version 1 it will be smooth sailing

PR #329, opened 2020-03-18
Makes root repository not installable with composer

GromNaN:
If the main package is required, it replaces any subpackage.
Same as Symfony: https://github.com/symfony/symfony/blob/5.0/composer.json#L38


jderusse:
Could you please add a note in the [contribution doc](https://github.com/async-aws/aws/blob/master/docs/maintainer-notes.md#creating-a-new-client-operation) ?

GromNaN:
> Caution is advised when using replace for the sub-package purpose explained above. You should then typically only replace using `self.version` as a version constraint, to make sure the main package only replaces the sub-packages of that exact version, and not any other version, which would be incorrect.
https://getcomposer.org/doc/04-schema.md#replace

If there is no tag on this repository, composer cannot reliably decide which version it can replace. This main package cannot be used as a library and should not be published on packagist.

Nyholm:
I’m happy to remove it from Packagist tbh :)

GromNaN:
Then, the package name should be removed and the type changed to something different than `library`.

Nyholm:
Package is removed from Packagist.  

Could you update the PR with the changes you suggested?

Readme.md
```diff
@@ -112,12 +112,16 @@ echo $waiter->isSuccess(); // true
...
 | [async-aws/cloud-formation](https://github.com/async-aws/cloud-formation)     | [![Latest Stable Version](https://poser.pugx.org/async-aws/cloud-formation/v/stable)](https://packagist.org/packages/async-aws/cloud-formation)      [![Total Downloads](https://poser.pugx.org/async-aws/cloud-formation/downloads)](https://packagist.org/packages/async-aws/cloud-formation)         | [![](https://github.com/async-aws/cloud-formation/workflows/BC%20Check/badge.svg?branch=master)](https://github.com/async-aws/cloud-formation/actions)
+| [async-aws/dynamo-db](https://github.com/async-aws/dynamo-db)                 | [![Latest Stable Version](https://poser.pugx.org/async-aws/dynamo-db/v/stable)](https://packagist.org/packages/async-aws/dynamo-db)                  [![Total Downloads](https://poser.pugx.org/async-aws/lambda/downloads)](https://packagist.org/packages/async-aws/lambda)                           | [![](https://github.com/async-aws/dynamo-db/workflows/BC%20Check/badge.svg?branch=master)](https://github.com/async-aws/dynamo-db/actions)
```
GromNaN: Was missing.

Readme.md
```diff
@@ -112,12 +112,16 @@ echo $waiter->isSuccess(); // true
...
 | ----------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------- |
+| [async-aws/code-generator](https://github.com/async-aws/code-generator)       | [![Latest Stable Version](https://poser.pugx.org/async-aws/code-generator/v/stable)](https://packagist.org/packages/async-aws/code-generator)        [![Total Downloads](https://poser.pugx.org/async-aws/code-generator/downloads)](https://packagist.org/packages/async-aws/code-generator)            | [![](https://github.com/async-aws/code-generator/workflows/BC%20Check/badge.svg?branch=master)](https://github.com/async-aws/code-generator/actions)
```
Nyholm: This is not ready do be displayed or versioned

jderusse: I'm not sure this package should be exposed on main page. People are looking for AWS services don't you think?

PR #330, opened 2020-03-18
Small refactor and get content-type version from definition

Nyholm:


src/CodeGenerator/src/Generator/RequestSerializer/JsonRpcSerializer.php
```diff
@@ -14,19 +14,20 @@
  *
  * @internal
  */
-class JsonSerializer extends RestJsonSerializer
+class JsonRpcSerializer extends RestJsonSerializer
```
Nyholm: To be more similar to the official API.

PR #338, opened 2020-03-19
[DynamoDB] Add waiter for table not exists after delete

GromNaN:
This waiter is in the official SDK.
https://github.com/aws/aws-sdk-php/blob/3.133.39/src/data/dynamodb/2012-08-10/waiters-2.json

Usage: 
```php
$dynamodbClient->deleteTable(['TableName' => 'errors']);
$dynamodbClient->tableNotExists(['TableName' => 'errors'])->wait();
```

### DIFF
--- a/src/CodeGenerator/src/Generator/WaiterGenerator.php
+++ b/src/CodeGenerator/src/Generator/WaiterGenerator.php
@@ -182,6 +182,8 @@ class WaiterGenerator
                 return $this->getAcceptorStatusBody($acceptor);
             case WaiterAcceptor::MATCHER_PATH:
                 return $this->getAcceptorPathBody($acceptor);
+            case WaiterAcceptor::MATCHER_ERROR:
+                return $this->getAcceptorErrorBody($acceptor);
             default:
                 throw new \RuntimeException(sprintf('Acceptor matcher "%s" is not yet implemented', $acceptor->getMatcher()));
         }
@@ -214,8 +216,11 @@ class WaiterGenerator
 
     private function getAcceptorErrorBody(ErrorWaiterAcceptor $acceptor): string
     {
-        $checks = ['$exception !== null'];
+        $checks = [];
         $error = $acceptor->getError();
+        if (null !== $expected = $acceptor->getExpected()) {
+            $checks[] = '$exception->getAwsCode() === ' . \var_export($expected, true);
+        }
         if (null !== $code = $error->getCode()) {
             $checks[] = '$exception->getAwsCode() === ' . \var_export($code, true);
         }
@@ -223,8 +228,12 @@ class WaiterGenerator
             $checks[] = '$exception->getCode() === ' . \var_export($code, true);
         }
 
+        if (0 === \count($checks)) {
+            return '';
+        }
+
         return strtr('
-            if (EXPECTED) {
+            if ($exception !== null && EXPECTED) {
                 return self::BEHAVIOR;
             }
         ', [
--- a/src/CodeGenerator/src/Generator/WaiterGenerator.php
+++ b/src/CodeGenerator/src/Generator/WaiterGenerator.php
@@ -218,14 +218,15 @@ class WaiterGenerator
     {
         $checks = [];
         $error = $acceptor->getError();
-        if (null !== $expected = $acceptor->getExpected()) {
+        if ($error->hasError()) {
+            if (null !== $code = $error->getCode()) {
+                $checks[] = '$exception->getAwsCode() === ' . \var_export($code, true);
+            }
+            if (null !== $code = $error->getStatusCode()) {
+                $checks[] = '$exception->getCode() === ' . \var_export($code, true);
+            }
+        } elseif (null !== $expected = $acceptor->getExpected()) {
             $checks[] = '$exception->getAwsCode() === ' . \var_export($expected, true);
-        }
-        if (null !== $code = $error->getCode()) {
-            $checks[] = '$exception->getAwsCode() === ' . \var_export($code, true);
-        }
-        if (null !== $code = $error->getStatusCode()) {
-            $checks[] = '$exception->getCode() === ' . \var_export($code, true);
         }
 
         if (0 === \count($checks)) {
--- a/src/CodeGenerator/src/Generator/WaiterGenerator.php
+++ b/src/CodeGenerator/src/Generator/WaiterGenerator.php
@@ -182,8 +182,6 @@ class WaiterGenerator
                 return $this->getAcceptorStatusBody($acceptor);
             case WaiterAcceptor::MATCHER_PATH:
                 return $this->getAcceptorPathBody($acceptor);
-            case WaiterAcceptor::MATCHER_ERROR:
-                return $this->getAcceptorErrorBody($acceptor);
             default:
                 throw new \RuntimeException(sprintf('Acceptor matcher "%s" is not yet implemented', $acceptor->getMatcher()));
         }
@@ -220,13 +218,13 @@ class WaiterGenerator
         $error = $acceptor->getError();
         if ($error->hasError()) {
             if (null !== $code = $error->getCode()) {
-                $checks[] = '$exception->getAwsCode() === ' . \var_export($code, true);
+                $checks[] = \var_export($code, true) . ' === $exception->getAwsCode()';
             }
             if (null !== $code = $error->getStatusCode()) {
-                $checks[] = '$exception->getCode() === ' . \var_export($code, true);
+                $checks[] = \var_export($code, true) . ' === $exception->getCode()';
             }
         } elseif (null !== $expected = $acceptor->getExpected()) {
-            $checks[] = '$exception->getAwsCode() === ' . \var_export($expected, true);
+            $checks[] = \var_export($expected, true) . ' === $exception->getAwsCode()';
         }
 
         if (0 === \count($checks)) {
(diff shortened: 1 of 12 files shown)


src/Service/DynamoDb/src/Result/TableExistsWaiter.php
```diff
@@ -11,18 +11,14 @@
 class TableExistsWaiter extends Waiter
 {
     protected const WAIT_TIMEOUT = 500.0;
-    protected const WAIT_DELAY = 20.0;
+    protected const WAIT_DELAY = 5.0;
```
GromNaN: Whereas the default time is 20 seconds, the table are actually created in less that 5 seconds.

src/Service/DynamoDb/src/Result/TableNotExistsWaiter.php
```diff
@@ -0,0 +1,41 @@
...
+use AsyncAws\Core\Exception\Http\HttpException;
+use AsyncAws\Core\Response;
+use AsyncAws\Core\Waiter;
+use AsyncAws\DynamoDb\DynamoDbClient;
+use AsyncAws\DynamoDb\Input\DescribeTableInput;
+
+class TableNotExistsWaiter extends Waiter
+{
+    protected const WAIT_TIMEOUT = 500.0;
+    protected const WAIT_DELAY = 2.0;
```
GromNaN: Whereas the default time is 20 seconds, the empty table are actually deleted in less that 2 seconds.

src/Service/DynamoDb/src/Result/TableExistsWaiter.php
```diff
@@ -11,18 +11,14 @@
...
 
     protected function extractState(Response $response, ?HttpException $exception): string
     {
         if (200 === $response->getStatusCode() && 'ACTIVE' === ($response->toArray()['Table']['TableStatus'] ?? null)) {
             return self::STATE_SUCCESS;
         }
 
-        if (null !== $exception) {
```
GromNaN: This condition is the same as the following ternary condition.

src/Service/DynamoDb/src/Result/TableExistsWaiter.php
jderusse: this code is generated, you can't change it (CI fails)

this change have to be proposed in AWS reposity. Or we should think about a way to override it

src/Service/DynamoDb/src/Result/TableExistsWaiter.php
jderusse: please open an issue.

when the last check is about the exception, the ternary is not needed

GromNaN:
The CI realigned me to the correct path. I'm updating the code generator & exception parser.

Nyholm:
Make sure to update the generator in a second PR. That will make it easier to review and to merge quickly.

src/Core/src/Response.php
```diff
@@ -87,18 +92,17 @@ public function __destruct()
...
-
-            if (400 <= $statusCode) {
-                throw $this->resolveResult = new ClientException($this->response);
-            }
-
-            if (300 <= $statusCode) {
-                throw $this->resolveResult = new RedirectionException($this->response);
-            }
+        if (500 <= $statusCode) {
+            $this->exception = new ServerException($this->response);
```
jderusse: Just realized that this code were wrong, calling this method several time with $throw=false then true does not return the rigth result. You don't need `this->exception`, we can use `this->resolveResult` to store the exception.

I think a best fix would be to remove the `$throw` parameter in this class,  and use try/catch in waiter to get the exception and silent errors

src/Core/src/Exception/Http/HttpExceptionTrait.php
```diff
@@ -85,6 +85,9 @@ public function __construct(ResponseInterface $response)
                 $this->awsMessage = $body['Message'];
                 $message .= "\n\n" . $body['Message'] . "\n\n";
             }
+            if (isset($body['__type'])) {
+                list(, $this->awsCode) = explode('#', $body['__type'], 2);
```
jderusse: SDK uses `$data['code'] = isset($parts[1]) ? $parts[1] : $parts[0];` meaning, it fallback to `parts[0]`

src/Service/Sqs/src/Result/QueueExistsWaiter.php
```diff
@@ -19,7 +19,7 @@ protected function extractState(Response $response, ?HttpException $exception):
             return self::STATE_SUCCESS;
         }
 
-        if (null !== $exception && 'AWS.SimpleQueueService.NonExistentQueue' === $exception->getAwsCode() && 400 === $exception->getCode()) {
+        if (null !== $exception && 'QueueDoesNotExist' === $exception->getAwsCode() && 'AWS.SimpleQueueService.NonExistentQueue' === $exception->getAwsCode() && 400 === $exception->getCode()) {
```
GromNaN: Effet de bord.

src/CodeGenerator/src/Generator/WaiterGenerator.php
```diff
@@ -214,17 +216,24 @@ private function getAcceptorPathBody(WaiterAcceptor $acceptor): string
 
     private function getAcceptorErrorBody(ErrorWaiterAcceptor $acceptor): string
     {
-        $checks = ['$exception !== null'];
+        $checks = [];
         $error = $acceptor->getError();
+        if (null !== $expected = $acceptor->getExpected()) {
```
jderusse: IMHO this should be done only when the Exception does not have `error` attributes (The `QueueExistsWaiter` should not change)

Moreover, this is always true, as `getError()` called above is extract from `expected` value.

Do understand how it works, the `expected=ResourceNotFoundException` refer to a shape:
```yaml
    "ResourceNotFoundException":{
      "type":"structure",
      "members":{
        "message":{"shape":"ErrorMessage"}
      },
      "exception":true
    },
```
for SQS, this shape looks like
```
    "BatchEntryIdsNotDistinct":{
      "type":"structure",
      "members":{
      },
      "error":{
        "code":"AWS.SimpleQueueService.BatchEntryIdsNotDistinct",
        "httpStatusCode":400,
        "senderFault":true
      },
      "exception":true
    },
```

Is suggest to add the method `ExceptionShape::hasError`
```php
    public function hasError(): bool
    {
        return isset($this->data['error']);
    }
```

and change your code
```php
if (!$error->hasError()) {
    $checks[] = '$exception->getAwsCode() === ' . \var_export($expected, true);
} else {
    if (null !== $code = $error->getCode()) {
        $checks[] = '$exception->getAwsCode() === ' . \var_export($code, true);
    }
    if (null !== $code = $error->getStatusCode()) {
        $checks[] = '$exception->getCode() === ' . \var_export($code, true);
    }
}

File: src/Core/src/Response.php (hunk 4 above)
Comment (GromNaN) [reply src/Core/src/Response.php]: I've removed the `$throw` argument.

File: src/CodeGenerator/src/Generator/WaiterGenerator.php (hunk 7 above)
Comment (GromNaN) [reply src/CodeGenerator/src/Generator/WaiterGenerator.php]: Done.

File: src/Core/src/Exception/Http/HttpExceptionTrait.php (hunk 5 above)
Comment (GromNaN) [reply src/Core/src/Exception/Http/HttpExceptionTrait.php]: Done.

File: src/Core/src/Waiter.php [hunk 8]
```diff
@@ -120,13 +126,12 @@ public function __destruct()
...
      *
-     * @return bool whether the request is executed or not
-     *
      * @throws NetworkException
+     * @throws HttpException
      */
     final public function resolve(?float $timeout = null): bool
     {
-        return $this->response->resolve($timeout, false);
+        return $this->response->resolve($timeout);
```
jderusse: You should catch here

src/Core/src/Waiter.php
```diff
@@ -64,8 +64,10 @@ public function __construct(Response $response, AbstractApi $awsClient, $request
 
     public function __destruct()
     {
-        if (null === $this->resolveResult) {
-            $this->resolve();
+        try {
+            $this->response->resolve();
```
jderusse: you can call `$this->resolve()` that should catch the exception

src/Core/src/Waiter.php
```diff
@@ -163,8 +168,12 @@ public function __destruct()
 
         $start = \microtime(true);
         while (true) {
-            if (!$this->resolve($timeout - (\microtime(true) - $start))) {
-                break;
+            try {
+                if (!$this->response->resolve($timeout - (\microtime(true) - $start))) {
```
jderusse: use `$this->resolve()` that shoud catch the exception

src/CodeGenerator/src/Generator/WaiterGenerator.php
```diff
@@ -214,17 +214,25 @@ private function getAcceptorPathBody(WaiterAcceptor $acceptor): string
...
     private function getAcceptorErrorBody(ErrorWaiterAcceptor $acceptor): string
     {
-        $checks = ['$exception !== null'];
+        $checks = [];
         $error = $acceptor->getError();
-        if (null !== $code = $error->getCode()) {
-            $checks[] = '$exception->getAwsCode() === ' . \var_export($code, true);
+        if ($error->hasError()) {
+            if (null !== $code = $error->getCode()) {
+                $checks[] = \var_export($code, true) . ' === $exception->getAwsCode()';
```
GromNaN: I changed the order `value === variable` to align with what cs-fixer does.

Nyholm:
Thank you for this PR.

PR #341, opened 2020-03-20
PreSign request

### LINKED ISSUE #336: Generating a signed URL for a S3 object
When storing objects with a private ACL, a common need is to generate a signed URL to allow temporary access to this object to the client getting this URL. It would be great if this could be available in this SDK.

jderusse:
Fixes #336 

usage:
```php
$input = new SendMessageRequest([]);
$url = $client->presign($input);
```
```php
$s3 = new S3Client(['region' => 'eu-west-1']);
$req = $s3->presign(new GetObjectRequest(['Bucket' => 'foo', 'Key' => 'big']));
echo (string) $req->getUri();
```

### DIFF
--- a/docs/compare.md
+++ b/docs/compare.md
@@ -4,7 +4,7 @@
 |---|-------------|-----------|
 | [Async](#async-experience)                    | | :heavy_check_mark: |
 | [Pagination](#pagination-experience)          | | :heavy_check_mark: |
-| [Presign](#presign-experience)          | | :heavy_check_mark: |
+| [Presign](#presign-experience)                | | :heavy_check_mark: |
 | [Weight](#dependencies-size)                  | | :heavy_check_mark: |
 | [Developer Experience](#developer-experience) | | :heavy_check_mark: |
 | [Mock / Proxy](#mock-and-proxy)               | | :heavy_check_mark: |
@@ -112,7 +112,7 @@ $content = $s3Client->getObject($input);
 
 // Presign Url
 $url = $s3Client->presign($input, new \Datetime('+60 min'));
-echo (string) $url;
+echo $url;
 ```
 
 **Official AWS PHP SDK:**
--- a/docs/compare.md
+++ b/docs/compare.md
@@ -4,6 +4,7 @@
 |---|-------------|-----------|
 | [Async](#async-experience)                    | | :heavy_check_mark: |
 | [Pagination](#pagination-experience)          | | :heavy_check_mark: |
+| [Presign](#presign-experience)          | | :heavy_check_mark: |
 | [Weight](#dependencies-size)                  | | :heavy_check_mark: |
 | [Developer Experience](#developer-experience) | | :heavy_check_mark: |
 | [Mock / Proxy](#mock-and-proxy)               | | :heavy_check_mark: |
@@ -95,6 +96,41 @@ foreach ($pages as $page) {
 *note*: Even if pagination is automatically handled, AsyncAws let you fetch
 only result for the current page.
 
+## Presign eXperience
+
+AWS allow pre-generating sign url that let user access to a resource
+without exposing the key. For instance, provide a link to Download a S3 Object.
+AsycAWS provides a fancy way to generate such url by reusing the same objects
+used in the standard way.
+
+**AsyncAws:**
+```php
+$input = new GetObjectRequest(['Bucket' => 'my-bucket', 'Key' => 'test]);
+
+// Sign on the fly
+$content = $s3Client->getObject($input);
+
+// Presign Url
+$psr7 = $s3Client->presign($input, new \Datetime('+60 min'));
+echo (string) $psr7->getUri();
+```
+
+**Official AWS PHP SDK:**
+```php
+
+// Sign on the fly
+$content = $s3Client->getObject(['Bucket' => 'my-bucket', 'Key' => 'test]);
+
+// Presign Url
+$command = $s3Client->getCommand('GetObject', ['Bucket' => 'my-bucket', 'Key' => 'test]);
+$psr7 = $s3Client->createPresignedRequest($cmd, '+60 min');
+echo (string) $psr7->getUri();
+
+```
+
+*note*: While official AWS PHP SDK provide methods to presign S3 methods only,
+AsyncAws let you presign requests for every services.
+
 ## Developer eXperience
 
 Ever get the error `PHP Fatal error: Missing required client configuration
@@ -103,8 +139,7 @@ to blindly copy/paste `['version' => '2006-03-01']`? AsyncAws saved you from
 this complexity and use the right version for you.
 
 AsyncAws also provides real classes with documented getter and methods, while
- the
-official AWS PHP SDK uses magic methods and undocumented array accessor.
+the official AWS PHP SDK uses magic methods and undocumented array accessor.
 
 | AWS PHP SDK | AsyncAws |
 |-------------|-----------|
--- a/docs/compare.md
+++ b/docs/compare.md
@@ -111,8 +111,8 @@ $input = new GetObjectRequest(['Bucket' => 'my-bucket', 'Key' => 'test]);
 $content = $s3Client->getObject($input);
 
 // Presign Url
-$psr7 = $s3Client->presign($input, new \Datetime('+60 min'));
-echo (string) $psr7->getUri();
+$url = $s3Client->presign($input, new \Datetime('+60 min'));
+echo (string) $url;
 ```
 
 **Official AWS PHP SDK:**
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -14,7 +14,11 @@ use AsyncAws\Core\Credentials\WebIdentityProvider;
 use AsyncAws\Core\Exception\InvalidArgument;
 use AsyncAws\Core\Signer\Signer;
 use AsyncAws\Core\Signer\SignerV4;
+use AsyncAws\Core\Stream\ResourceStream;
 use AsyncAws\Core\Stream\StringStream;
+use AsyncAws\Core\Tests\Unit\Stream\ResourceStreamTest;
+use Nyholm\Psr7\Factory\Psr17Factory;
+use Psr\Http\Message\RequestInterface;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Symfony\Component\HttpClient\HttpClient;
@@ -85,6 +89,47 @@ abstract class AbstractApi
     abstract protected function getSignatureVersion(): string;
 
     abstract protected function getSignatureScopeName(): string;
+
+    final public function presign(Input $input): RequestInterface
+    {
+        if (!\class_exists(Psr17Factory::class)) {
+            throw new \LogicException('You cannot presign requests as the "nyholm/psr7" package is not installed. Try running "composer require nyholm/psr7".');
+        }
+        $request = $input->request();
+        $request->setEndpoint($this->getEndpoint($request->getUri(), $request->getQuery()));
+
+        // todo sign
+        // $this->getSigner()->presign($request, $this->credentialProvider->getCredentials($this->configuration));
+
+        $length = $request->getBody()->length();
+        if (null !== $length && !$request->hasHeader('content-length')) {
+            $request->setHeader('content-length', $length);
+        }
+
+        // Some servers (like testing Docker Images) does not supports `Transfer-Encoding: chunked` requests.
+        // The body is converted into string to prevent curl using `Transfer-Encoding: chunked` unless it really has to.
+        if (($requestBody = $request->getBody()) instanceof StringStream) {
+            $requestBody = $requestBody->stringify();
+        }
+
+        $factory = new Psr17Factory();
+        $psr = $factory->createRequest(
+            $request->getMethod(),
+            $request->getEndpoint()
+        );
+        foreach ($request->getHeaders() as $headerName => $headerValue) {
+            $psr->withAddedHeader($headerName, $headerValue);
+        }
+        if ($length > 0) {
+            if ($requestBody instanceof ResourceStream) {
+                $psr->withBody($factory->createStreamFromResource($requestBody->getResource()));
+            } else {
+                $psr->withBody($factory->createStream($requestBody->stringify()));
+            }
+        }
+
+        return $psr;
+    }
 
     final protected function getResponse(Request $request): Response
     {
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -85,7 +85,9 @@ abstract class AbstractApi
         $request = $input->request();
         $request->setEndpoint($this->getEndpoint($request->getUri(), $request->getQuery()));
 
-        $this->getSigner()->presign($request, $this->credentialProvider->getCredentials($this->configuration), $expires ? $expires : new \DateTimeImmutable('+60min'));
+        if (null !== $credentials = $this->credentialProvider->getCredentials($this->configuration)) {
+            $this->getSigner()->presign($request, $credentials, new RequestContext(['expirationDate' => $expires]));
+        }
 
         return $request->getEndpoint();
     }
@@ -96,11 +98,14 @@ abstract class AbstractApi
 
     abstract protected function getSignatureScopeName(): string;
 
-    final protected function getResponse(Request $request, ?string $operation = null): Response
+    final protected function getResponse(Request $request, ?RequestContext $context = null): Response
     {
         $request->setEndpoint($this->getEndpoint($request->getUri(), $request->getQuery()));
 
-        $this->getSigner()->sign($request, $this->credentialProvider->getCredentials($this->configuration), $operation);
+        if (null !== $credentials = $this->credentialProvider->getCredentials($this->configuration)) {
+            $this->getSigner()->sign($request, $credentials, $context ?? new RequestContext());
+        }
+
         $length = $request->getBody()->length();
         if (null !== $length && !$request->hasHeader('content-length')) {
             $request->setHeader('content-length', $length);
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -16,7 +16,6 @@ use AsyncAws\Core\Signer\Signer;
 use AsyncAws\Core\Signer\SignerV4;
 use AsyncAws\Core\Stream\ResourceStream;
 use AsyncAws\Core\Stream\StringStream;
-use AsyncAws\Core\Tests\Unit\Stream\ResourceStreamTest;
 use Nyholm\Psr7\Factory\Psr17Factory;
 use Psr\Http\Message\RequestInterface;
 use Psr\Log\LoggerInterface;
@@ -84,13 +83,7 @@ abstract class AbstractApi
         return $this->configuration;
     }
 
-    abstract protected function getServiceCode(): string;
-
-    abstract protected function getSignatureVersion(): string;
-
-    abstract protected function getSignatureScopeName(): string;
-
-    final public function presign(Input $input): RequestInterface
+    final public function presign(Input $input, ?\DateTimeInterface $expires = null): RequestInterface
     {
         if (!\class_exists(Psr17Factory::class)) {
             throw new \LogicException('You cannot presign requests as the "nyholm/psr7" package is not installed. Try running "composer require nyholm/psr7".');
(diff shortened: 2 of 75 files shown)


Nyholm:
Yes, it seams fair.

src/Core/src/Signer/SignerV4.php
```diff
@@ -56,6 +56,82 @@ public function __construct(string $scopeName, string $region)
...
+        }
+
+        if (false === $parsedUrl = parse_url($request->getEndpoint())) {
+            throw new InvalidArgument(sprintf('The endpoint "%s" is invalid.', $request->getEndpoint()));
+        }
+        if (null !== $sessionToken = $credentials->getSessionToken()) {
+            $request->setHeader('x-amz-security-token', $sessionToken);
+        }
+
+        $this->convertPostToGet($request);
```
Nyholm: Can you tell me why we do this?

src/Core/src/AbstractApi.php
```diff
@@ -79,6 +82,40 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
+            $request->setHeader('content-length', $length);
+        }
+
+        $factory = new Psr17Factory();
+        $psr = $factory->createRequest(
+            $request->getMethod(),
+            $request->getEndpoint()
+        );
+        foreach ($request->getHeaders() as $headerName => $headerValue) {
+            $psr->withAddedHeader($headerName, $headerValue);
```
Nyholm: $psr = 

Since it is immutable.

src/Core/src/AbstractApi.php
```diff
@@ -79,6 +82,40 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
+        }
+        if ($length > 0) {
+            if (($requestBody = $request->getBody()) instanceof ResourceStream) {
+                $psr->withBody($factory->createStreamFromResource($requestBody->getResource()));
+            } else {
+                $psr->withBody($factory->createStream($requestBody->stringify()));
```
Nyholm: Same

src/Core/src/AbstractApi.php
```diff
@@ -79,6 +82,40 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
+        $psr = $factory->createRequest(
+            $request->getMethod(),
+            $request->getEndpoint()
+        );
+        foreach ($request->getHeaders() as $headerName => $headerValue) {
+            $psr->withAddedHeader($headerName, $headerValue);
+        }
+        if ($length > 0) {
+            if (($requestBody = $request->getBody()) instanceof ResourceStream) {
+                $psr->withBody($factory->createStreamFromResource($requestBody->getResource()));
```
Nyholm: Same

src/Core/src/Signer/SignerV4.php
jderusse: That's convert `x-www-form-urlencoded` to `query-string`. In order to presign non-rest requests (ie SQS). 
AWS does not support mixing queryString and form post. But they support Form send as GET

nb: This is a copy/paste from official AWS SDK on EC2 and RDS service. 

I'm not such a should call this for every services. It would robably break S3 Upload. I'll dig into it today.

src/Core/src/AbstractApi.php
jderusse: good catch

Nyholm:
Sorry for asking.. But what is the user expected to do with the PSR request? 
If I look at the official docs, the only thing I see is to get the URL. What else?

What Im asking is: Do we need PSR7?
One could:
A) Use our Request. But that is a bad idea since it is internal and we like to keep it that way. 
B) Use Symfony's Request. But Im not sure how much you may change that or use it. 
C) Return only the URL
D) Return a new model with URL and headers
E) Return `[$url, $headers]`, this is ugly.

jderusse:
I ended to the same conclusion.

returning string (option C) does not cover 100% of cases but should be enough for a first version IMHO.

docs/compare.md
```diff
@@ -4,6 +4,7 @@
 |---|-------------|-----------|
 | [Async](#async-experience)                    | | :heavy_check_mark: |
 | [Pagination](#pagination-experience)          | | :heavy_check_mark: |
+| [Presign](#presign-experience)          | | :heavy_check_mark: |
```
Nyholm: This makes my OCD sense tingling. 

Please align this =)

jderusse:
I finally rewrite a LOT the Signer (and the Stream to avoid conversion to string when computing hash)

src/Core/src/AbstractApi.php
```diff
@@ -79,18 +79,27 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
+    }
+
     abstract protected function getServiceCode(): string;
 
     abstract protected function getSignatureVersion(): string;
 
     abstract protected function getSignatureScopeName(): string;
 
-    final protected function getResponse(Request $request): Response
+    final protected function getResponse(Request $request, ?string $operation = null): Response
```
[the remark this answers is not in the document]
jderusse: in fact, this is not a `or`, but an extra header required by some S3 operation.
The information is not available in metadata file, and can't not be generated :-(

Nevertheless, I like the idea of object to avoid futures BC break.

src/Core/src/Signer/Signer.php
```diff
@@ -12,5 +12,7 @@
  */
 interface Signer
 {
-    public function sign(Request $request, ?Credentials $credentials): void;
+    public function sign(Request $request, ?Credentials $credentials, ?string $operation = null, \DateTimeInterface $now = null): void;
+
+    public function presign(Request $request, ?Credentials $credentials, \DateTimeInterface $expires, ?\DateTimeInterface $now = null): void;
```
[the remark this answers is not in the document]
jderusse: I injected `now` to create reproductible test. .That could be moved in configuration indeed

docs/compare.md
jderusse: sorry. Fixed

docs/compare.md
```diff
@@ -95,6 +96,41 @@ foreach ($pages as $page) {
...
+**AsyncAws:**
+```php
+$input = new GetObjectRequest(['Bucket' => 'my-bucket', 'Key' => 'test]);
+
+// Sign on the fly
+$content = $s3Client->getObject($input);
+
+// Presign Url
+$url = $s3Client->presign($input, new \Datetime('+60 min'));
+echo (string) $url;
```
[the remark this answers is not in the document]
jderusse: good catch

src/Core/src/AbstractApi.php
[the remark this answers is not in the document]
jderusse: added a ContextRequest object

src/Core/src/Signer/Signer.php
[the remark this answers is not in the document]
jderusse: No is inject vie requestObject for tests

src/Core/src/RequestContext.php
```diff
@@ -0,0 +1,69 @@
...
+        'currentDate' => true,
+    ];
+
+    /**
+     * @var string|null
+     */
+    private $operation;
+
+    /**
+     * @var \DateTimeImmutable|null
```
Nyholm: Should be `DateTimeInterface`. Or else it will not match the signature of `presign()`

src/Core/src/AbstractApi.php
```diff
@@ -85,7 +85,9 @@ public function __construct($configuration = [], ?CredentialProvider $credential
         $request = $input->request();
         $request->setEndpoint($this->getEndpoint($request->getUri(), $request->getQuery()));
 
-        $this->getSigner()->presign($request, $this->credentialProvider->getCredentials($this->configuration), $expires ? $expires : new \DateTimeImmutable('+60min'));
+        if (null !== $credentials = $this->credentialProvider->getCredentials($this->configuration)) {
+            $this->getSigner()->presign($request, $credentials, new RequestContext(['expirationDate' => $expires]));
```
Nyholm: Will this work? We dont have operation here.

jderusse: the operation is not a strong requirement, it avoid computing md5 on some S3 endpoint.

src/Core/src/RequestContext.php
jderusse: Good catch. I switched to DateTimeInterface 
=> it adds complexity to generate a default expiration date. Because `\DatetimeInterface::add` does not exists

jderusse:
> Ie $request->context(). That will make the getResponse signature cleaner.

I'm not sure, the 2 instances `request` and `requestContext`are not built at the same not by the same classes.
For instance in presign, `request` is built by `Input`, and `requestContext` is built by the AbstractApi with the user's input...

src/Core/src/AbstractApi.php
Nyholm: Yes. But can I presign those endpoints if I dont pass "operation"?

jderusse: Yes it will work.
Because body is not used to compute the signature (special case for S3 using a special hash `UNSIGNED-PAYLOAD`) we don't have have to know the operation, because we don't have to MD5 the body

Nyholm: Okey. I understand. 
Thank you

src/Core/src/Stream/RewindableStream.php
```diff
@@ -0,0 +1,88 @@
...
+
+    public function getIterator(): \Traversable
+    {
+        if (null !== $this->fallback) {
+            yield from $this->fallback;
+
+            return;
+        }
+
+        $resource = \tmpfile();
```
Nyholm: Is it overkill to check the length first? If the length is less that 1Mb then maybe we should store it in memory?

src/Core/src/Stream/RewindableStream.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->fallback = ResourceStream::create($resource);
+
+        foreach ($this->content as $chunk) {
+            \fwrite($resource, $chunk);
+            yield $chunk;
```
Nyholm: If I do : 

```php
$readOnlySteam = // read once stream
echo $readOnlySteam->length(); // 50Mb
$stream = RewindableStream::create($readOnlySteam);
echo $readOnlySteam->length(); // 50Mb

foreach ($stream as $part) {
   break;
}

echo $stream->length(); // 64 kb (or size of chunk)

```

Is this something we should consider?

src/Core/src/Stream/StreamFactory.php
```diff
@@ -17,7 +17,7 @@ public static function create($content): Stream
             return StringStream::create($content ?? '');
         }
         if (\is_callable($content)) {
-            return CallableStream::create($content);
+            return RewindableStream::create(CallableStream::create($content));
```
Nyholm: Why are we always writing a callable to disk?
Will we always read it twice or is that only a special case?

jderusse:
I think I'm done with this one. Would you min to have a last review?

src/Core/src/Stream/RewindableStream.php
jderusse: Good idea !!

src/Core/src/Stream/RewindableStream.php
jderusse: If it wasn't internal, yes, we should check if the stream have been consumed until the end (same for `hash` ans stringify).

Given it's internal, I'm not sure it worth it.

Nyholm: Add a comment so we dont forget it in the future. 

> This is for internal use only. If iterating over a stream, the full stream must be consumed. One cannot iterate only over a few items of the stream.

src/Core/src/Stream/StreamFactory.php
jderusse: Sadly yes, to compute hash in signature, we have to read the stream first THEN to read it a second time when sending the request.

Note: previous implementation ends to stringify the requests which were worse IMO.

BUT, I've updated the implementaiton to always use memory AND fallback to disk when the buffer exceed 1Mo. That should fixes your concerns

src/Core/src/Signer/SignerV4.php
```diff
@@ -56,87 +57,206 @@ public function __construct(string $scopeName, string $region)
...
+        }
+
+        if ('UNSIGNED-PAYLOAD' === $hash) {
+            $request->setHeader('x-amz-content-sha256', $hash);
+        }
+
+        return $hash;
+    }
+
+    private function handleSignature(Request $request, ?Credentials $credentials, \DateTimeInterface $now, \DateTimeInterface $expires, bool $isPresign = false): void
```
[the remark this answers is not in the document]
jderusse: thanks

src/Core/src/Stream/RewindableStream.php
jderusse: I gone further by always using memory, and then switched to disk later when memory buffer exceeded 1Mo.

src/Core/src/Stream/StreamFactory.php
Nyholm: Perfect!

src/Core/src/Signer/SignerV4.php
```diff
@@ -87,7 +87,7 @@ protected function buildBodyDigest(Request $request, bool $isPresign): string
         return $hash;
     }
 
-    private function handleSignature(Request $request, ?Credentials $credentials, \DateTimeInterface $now, \DateTimeInterface $expires, bool $isPresign = false): void
+    private function handleSignature(Request $request, ?Credentials $credentials, \DateTimeInterface $now, \DateTimeInterface $expires, bool $isPresign): void
```
Nyholm:
```suggestion
    private function handleSignature(Request $request, Credentials $credentials, \DateTimeInterface $now, \DateTimeInterface $expires, bool $isPresign): void
```

src/Core/src/Stream/RewindableStream.php
```diff
@@ -63,11 +63,26 @@ public function getIterator(): \Traversable
...
         $this->fallback = ResourceStream::create($resource);
 
+        $size = 0;
+        $inmemory = true;
         foreach ($this->content as $chunk) {
             \fwrite($resource, $chunk);
+            if ($inmemory) {
+                $size += \strlen($chunk);
+                // switch to filesystem
+                if ($size > 1) {
```
Nyholm:
```suggestion
                // If string larger that 1 Mb
                if ($this->fallback->length() > 1024*1024) {
```

We also need to close the memory resource

src/Core/src/Stream/RewindableStream.php
jderusse: done

src/Core/src/Stream/RewindableStream.php
jderusse: haha debug spoted...

I didn't used `$this->fallback->length()` because it's more expensive that a simple counter

src/Core/src/Signer/SignerV4.php
jderusse: fixed. Thanks

src/Core/src/Signer/SignerV4.php
```diff
@@ -56,87 +57,202 @@ public function __construct(string $scopeName, string $region)
...
+                throw new InvalidArgument('The expiration date of presigned URL must be less than one week');
+            }
+            if ($duration < 0) {
+                throw new InvalidArgument('The expiration date of presigned URL must be in the future');
+            }
 
-        $signingKey = 'AWS4' . $credentials->getSecretKey();
-        foreach ($credentialScope as $scopePart) {
-            $signingKey = hash_hmac('sha256', $scopePart, $signingKey, true);
+            $request->setQueryAttribute('X-Amz-Date', gmdate('Ymd\THis\Z', $now->getTimestamp()));
```
stof: Why not using `$now-format()` here ?

src/Core/src/Signer/SignerV4.php
```diff
@@ -56,87 +57,202 @@ public function __construct(string $scopeName, string $region)
...
+            // Should be called before `buildBodyDigest` because this method may alter the body
+            $this->convertBodyToQuery($request);
+        } else {
+            $this->convertBodyToStream($request, $now, $credentialString, $signingKey, $signature);
+        }
+
+        $bodyDigest = $this->buildBodyDigest($request, $isPresign);
+
+        if ($isPresign) {
+            // Should be called after `buildBodyDigest` because header this method may remove `x-amz-content-sha256`
```
stof: there is a mistake in this sentence. It does not make sense.

src/Core/src/Signer/SignerV4.php
```diff
@@ -56,87 +57,202 @@ public function __construct(string $scopeName, string $region)
...
+        $this->sanitizeHostForHeader($request);
+        $this->assignAmzQueryValues($request, $credentials, $isPresign);
+
+        $this->buildTime($request, $now, $expires, $isPresign);
+        $credentialScope = $this->buildCredentialString($request, $credentials, $now, $isPresign);
+        $credentialString = \implode('/', $credentialScope);
+        $signingKey = $this->buildSigningKey($credentials, $credentialScope);
+
+        // signature is passed by reference to convertBodyToStream
+        $signature = '';
```
stof: This variable is never used elsewhere than in the `convertBodyToStream` call (it is then always overridden on line 120 before being used again). So why defining it outside the `if/else` block ?

src/Core/src/Stream/ResourceStream.php
```diff
@@ -66,4 +66,33 @@ public function getIterator(): \Traversable
...
+        }
+
+        $ctx = hash_init($algo);
+        while (!feof($this->content)) {
+            hash_update($ctx, fread($this->content, 1048576));
+        }
+
+        $out = hash_final($ctx, $raw);
+        if (-1 === fseek($this->content, $pos)) {
+            throw new InvalidArgument('Unable to seek the content.');
```
stof: Isn't this a runtime exception rather than an invalid argument ?

src/Core/src/Stream/RewindableStream.php
```diff
@@ -0,0 +1,110 @@
...
+    public static function create($content): RewindableStream
+    {
+        if ($content instanceof self) {
+            return $content;
+        }
+        if ($content instanceof Stream) {
+            return new self($content);
+        }
+
+        throw new InvalidArgument(sprintf('Expect content to be a "Stream". "%s" given.', \is_object($content) ? \get_class($content) : \gettype($content)));
```
stof: Why not typehinting the argument then ?

src/Core/src/Stream/RewindableStream.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->fallback = ResourceStream::create($resource);
+
+        $size = 0;
+        $inmemory = true;
+        foreach ($this->content as $chunk) {
+            \fwrite($resource, $chunk);
+            if ($inmemory) {
+                $size += \strlen($chunk);
+                // switch to filesystem if string larger that 1 Mb
+                if ($size > 1024 * 1024) {
```
stof: You could use `php://temp`, which starts in memory and automatically turns it into a file when reaching a size threshold (2MB by default, but that can be configured when creating the stream). This would simplify the code a lot, as you are basically reimplementing it on top of `php://memory` and `tmpfile()`

src/Core/src/Signer/SignerV4.php
jderusse: thanks

src/Core/src/Signer/SignerV4.php
jderusse: indeed, the $signature could have been initialized in the `else` branch.
It have to be defined before the call to `convertBodyToStream` for matching the typehinting.

stof: Shouldn't we actually move the line 120 here, to pass the computed signature instead of computing it twice ?

src/Core/src/Stream/ResourceStream.php
jderusse: I believed that such case occurs only when user sent invalid resource that is not able to rewind.
Do you mean, that it could occurs even with a "valid" resource (valid = should be able to rewind)

src/Core/src/Stream/RewindableStream.php
jderusse: good catch

src/Core/src/Stream/RewindableStream.php
jderusse: I wasn't ware of that! thanks for the tips!

src/Core/src/Signer/SignerV4.php
jderusse: sadly not, there are complex circular dependency.
The line 120 `buildSignature` need the headers and the headers may be updated by this `buildBodyDigest`.

src/Core/src/Stream/RewindableStream.php
Nyholm: Cool. I didnt know that either.

src/Core/src/Signer/SignerV4.php
jderusse: Fixed. 
Added a `->setTimezone('GMT')` in `sign`/`presign` methods

PR #343, opened 2020-03-21
Add integration tests on DynamoDbClient::tableNotExsits

GromNaN:
Missing from #338


src/Core/tests/Unit/Exception/Http/ClientExceptionTest.php
```diff
@@ -38,4 +38,15 @@ public function testRestXmlError()
         self::assertEquals('NoSuchKey', $exception->getAwsCode());
         self::assertEquals('The specified key does not exist.', $exception->getAwsMessage());
     }
+
+    public function testDynamoDbError()
+    {
```
Nyholm: Good. This class needs more testing. 

(Maybe a cleanup in the future)

PR #346, opened 2020-03-21
Updating final keyword

Nyholm:
Im not sure Im 100% correct here. 

* I've added `final` on the Credentials. That is because they all have an interface. 
* I've also removed `final` on `@internal` classes/functions. Nobody should use them anyways.

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -24,6 +24,7 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
  * Base class most APIs are inheriting.
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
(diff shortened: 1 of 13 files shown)


Nyholm:
Question: Is `Request` and `Response` really internal?

All src/service/Foo/Input/* and src/service/Foo/Result/* classes are using them. So, "internal" to what?
- The project? Yes
- The core package? No. 

What are the pros/cons of doing `Request` and `Response` internal?

Nyholm:
Pros: 
- We dont need to care about BC. We can change the class whenever. 

Cons: 
- Nobody can use them. If a user decides to implement their own S3Client or their on FooClient, they cannot do that because they cannot use our Request/Response classes.

Nyholm:
Thank you

PR #348, opened 2020-03-21
Move StreamableBody to AsyncAws\Core\Stream

Nyholm:
I also renamed: 
- `StreamableBody` to `ResponseBodyStream`
- `StreamableBodyInterface` to `ResultStream`


Nyholm:
Should we rename `Stream` to `InputStream?`

jderusse:
Yes, I think we should not mix both (because all instances in `Stream` are related to the `Stream` interface) That could be confusion.

I vote for `InputStream` or `RequestBodyStream` or `RequestStream` or `BodyStream`

Nyholm:
If you don’t want them in the same namespace. What namespace do you suggest for the ResultStream?

jderusse:
What's about `RequestStream`+`ResultStream` or `Stream/Request`+`Stream/Result`
`Stream` interface should should be renamed into `RequestStream` for consistency

Nyholm:
Im happy with RequestStream+ResultStream. 

Is that putting them in the same namespace?

PR #350, opened 2020-03-21
Added comment on Stream

### LINKED ISSUE #349: Should Core\Stream be using multibyte support?
Does it matter if we are using `substr` and `strlen` compared to `mb_substr` and `mb_strlen`?

We do have a `Stream::length`, is that length in bytes or chars? 
I dont think it matters if `getIterator()` is using `mb_*` or not.

Nyholm:
This will fix #349


jderusse:
thanks you

PR #351, opened 2020-03-22
Removed deprecated code from #348

Nyholm:
In  #348 I started to rename a bunch of classes. I did it in a way so we dont break BC. The Flysystem tests were helpful to find out a BC breaking free path. 

If we want to, we can keep the current changes in master around for a while. Ie just before we release 1.0.0.  OR we can clean them up now. 

Ps. I code have added this commit to #348 but that would have made the CI fail on my PR.


Nyholm:
Could I get a second review? 

I found it hard to wrote unit tests for Flysystem 1 if I couldn't mock `Result::resolve()`. 

It this okey?

src/Integration/Flysystem/S3/tests/Unit/S3FilesystemV1Test.php
```diff
@@ -39,7 +40,7 @@ public function testWrite()
 
         $result = $this->getMockBuilder(PutObjectOutput::class)
```
jderusse: What's about using the `ResultMockFactory`?

src/Integration/Flysystem/S3/tests/Unit/S3FilesystemV1Test.php
```diff
@@ -39,7 +40,7 @@ public function testWrite()
...
             ->disableOriginalConstructor()
-            ->onlyMethods(['resolve'])
+            //->onlyMethods(['resolve'])
```
jderusse: don't forget to remove this

Nyholm: Thank you

src/Integration/Flysystem/S3/tests/Unit/S3FilesystemV1Test.php
Nyholm: I forgot about ResultMockFactory =)

But I still have issues with pagination.. I'll see if I can solve it.

jderusse: If you can't, Im fine with move the final from method to class

PR #352, opened 2020-03-22
Optimize Rewindable usage

jderusse:
This PR uses the Rewindable only when needed. It also fixes several things:

- generator ar traversed once
- fix issue when length is unknown


src/Core/src/Stream/StreamFactory.php
```diff
@@ -17,9 +17,13 @@ public static function create($content): RequestStream
             return StringStream::create($content ?? '');
         }
         if (\is_callable($content)) {
-            return RewindableStream::create(CallableStream::create($content));
+            return CallableStream::create($content);
         }
         if (\is_iterable($content)) {
+            if ($content instanceof \Generator) {
+                return GeneratorIterableStream::create($content);
```
jderusse: This is a special case for \Generator (read-once) while simple array can be read several time.

I wonder if ArrayObject should be read once too?....

MayBe I should use
```php
if (is_array) 
  return ArrayStream();
else
  return IterableStream();
```

src/Core/src/Signer/SignerV4.php
```diff
@@ -77,6 +79,12 @@ protected function buildBodyDigest(Request $request, bool $isPresign): string
         if ($request->hasHeader('x-amz-content-sha256')) {
             $hash = ((array) $request->getHeader('x-amz-content-sha256'))[0];
         } else {
+            $body = $request->getBody();
+            if ($body instanceof ReadOnceResultStream) {
+                dd($body);
```
Nyholm: Debug code

src/Core/src/Stream/GeneratorIterableStream.php
```diff
@@ -0,0 +1,27 @@
...
+use AsyncAws\Core\Exception\InvalidArgument;
+
+/**
+ * Convert an iterator into a Stream.
+ *
+ * @author Jérémy Derussé <jeremy@derusse.com>
+ *
+ * @internal
+ */
+final class GeneratorIterableStream extends IterableStream implements ReadOnceResultStream
```
Nyholm: Did we introduce this class just to add the `ReadOnceResultStream`?

src/Core/src/Stream/RewindableStream.php
```diff
@@ -107,4 +107,11 @@ public function hash(string $algo = 'sha256', bool $raw = false): string
 
         return hash_final($ctx, $raw);
     }
+
+    public function travers(): void
```
Nyholm: Maybe use a simpler name? `read()`?

src/Core/src/Stream/StreamFactory.php
Nyholm: Or just add the `ReadOnceResultStream` on `IterableStream`?

src/Core/src/Signer/SignerV4.php
```diff
@@ -245,6 +253,13 @@ private function convertBodyToStream(Request $request, \DateTimeInterface $now,
             $contentLength = $body->length();
         }
 
+        // If content length is unknown, use the rewindable stream to read it once locally in order to get the length
+        if (null === $contentLength) {
+            $request->setBody($body = RewindableStream::create($body));
+            $body->travers();
```
Nyholm: This function could be private, right?

src/Core/src/Stream/GeneratorIterableStream.php
jderusse: The purpose of this class is to distinguish array (rewindable) from generator (read once)

If you think it don't worth it, I can remove it.

src/Core/src/Signer/SignerV4.php
jderusse: not sure how. 
For instance here I want to get the exact value of length. While in AbstractApi, it's better to get length, but if wwe don't, it's not compulsotry.

Nyholm: In the `RewindableStream::length()`:

```php
    public function length(): ?int
    {
        if (null !== $this->fallback) {
            $this->read():
        }

         return $this->fallback->length();
    }
```

Is this a good idea? Will it work for `stringify()` and `hash()` too?


Hm... maybe no. Because that will always duplicate the stream. Sometimes there is no need to.

src/Core/src/Stream/IterableStream.php
```diff
@@ -7,15 +7,16 @@
...
+ * @internal
+ * @final
  */
-final class IterableStream implements RequestStream
+class IterableStream implements RequestStream, ReadOnceResultStream
 {
     private $content;
 
-    private function __construct(iterable $content)
+    protected function __construct(iterable $content)
```
Nyholm: This can be private (and the class final)

src/Core/src/Stream/RewindableStream.php
```diff
@@ -107,4 +107,11 @@ public function hash(string $algo = 'sha256', bool $raw = false): string
 
         return hash_final($ctx, $raw);
     }
+
+    public function read(): void
+    {
+        foreach ($this as $chunk) {
+            // do noting.
```
Nyholm: This is not true =)

> Use `getIterator()` to read stream content to $this->fallback

src/Core/src/Signer/SignerV4.php
jderusse: yes, that were my point but ou explained it better than me :)

src/Core/src/Stream/IterableStream.php
jderusse: indeed, forgot this

Nyholm:
THank you

PR #353, opened 2020-03-22
Minor cleanup, adding comments and renaming properties

Nyholm:
I thought it is clearer. What do you think?


src/Core/src/Response.php
```diff
@@ -25,20 +25,21 @@
  */
 class Response
 {
-    private $response;
+    private $httpResponse;
```
Nyholm: We currently have `Result::$response` and `Response::$response`. This makes it a little bit more clear.

PR #355, opened 2020-03-22
Updated readme about async stuff

Nyholm:
This might be more clear.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -40,9 +40,9 @@ $sqsClient = new SqsClient([
 
 // Call a client's method with an array
 $result = $sqsClient->createQueue(['QueueName' => 'bar']);
-// Make sure the request is sent
-$result->resolve();
 
+// Request is automatically sent when reading the result
+echo $result->getMessageId();
 
 // You can also call a client's method with an input object
 $input = new SendMessageRequest();
@@ -50,10 +50,10 @@ $input
     ->setQueueUrl('https://foo.com/bar')
     ->setMessageBody('foobar');
 
-$result = $sqsClient->sendMessage($input);
+// Since we dont use a $result to get the return value,
+// the HTTP request is sent automatically.
+$sqsClient->sendMessage($input);
 
-// Request is automatically sent when reading the response
-echo $result->getMessageId();
 ```
 
 ## How is it async first?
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -40,20 +40,20 @@ $sqsClient = new SqsClient([
...
 // You can also call a client's method with an input object
 $input = new SendMessageRequest();
 $input
     ->setQueueUrl('https://foo.com/bar')
     ->setMessageBody('foobar');
 
-$result = $sqsClient->sendMessage($input);
+// Since we dont use a $result to get the return value,
+// the HTTP request is sent automatically.
+$sqsClient->sendMessage($input);
```
Nyholm: I want to show that we actually send the HTTP request on `__desctruct()`. 

A call to `resolve()` is rarely needed.

Readme.md
```diff
@@ -97,13 +97,20 @@ Similar to Official AWS PHP SDK, AsyncAws provides waiters to let you wait
...
-$waiter = $sqsClient->queueExists(['QueueName' => 'fooBar']);
+// create a new table. It normally takes around 5 seconds to complete this action.
+$dbClient->createTable(['TableName' => 'foobar', /* ...*/]);
+
+$waiter = $dbClient->tableExists(['TableName' => 'foobar']);
 echo $waiter->isSuccess(); // false
+
 $waiter->wait();
 echo $waiter->isSuccess(); // true
+
```
Nyholm: I use a DynamoDB example here. 

@GromNaN, please correct me if Im wrong. Calling `createTable()` will get a response within a second. But you cannot use the table yet because it is not actually created. That is why you need the waiter.

Readme.md
jderusse: it's not needed, but it helps troubleshouting exception. 
When `resolve` is called on destruct and trigger an exception, that sometime is hard to understand (from a user point of view)

Nyholm: But in this case it is the exact opposite, right? 

Doing `$sqsClient->sendMessage($input);` will actually fire the request, If you get a 400 you will get an exception on this line. Ie, this is pretty much synchronous. 

The tricky part is when you do something like: 

```
function foo() {
   $result = $sqsClient->sendMessage($input);

  // Bunch of other code 

}
// the result will be destructed at the end of the block. 
``` 

But this is rare, either you need the result and use it, or you dont need it.

jderusse: My comment were just a reminder that `__destructor ` are hides sometime complexity in debugging.

Nyholm: Yeah, true. We should document `resolve()` but not here in the readme.

Readme.md
GromNaN: Sorry, I forgot to answer. This is totally correct.

PR #358, opened 2020-03-23
Fix PreSign request

jderusse:
This PR take comments from #341 into account

### DIFF
--- a/src/Core/src/Signer/SignerV4.php
+++ b/src/Core/src/Signer/SignerV4.php
@@ -59,15 +59,25 @@ class SignerV4 implements Signer
 
     public function presign(Request $request, Credentials $credentials, RequestContext $context): void
     {
-        $now = $context->getCurrentDate() ?? new \DateTimeImmutable();
-        $expires = $context->getExpirationDate() ?? (new \DateTimeImmutable($now->format(\DateTimeInterface::ATOM)))->add(new \DateInterval('PT1H'));
+        if (null === $now = $context->getCurrentDate()) {
+            $now = new \DateTimeImmutable();
+        } else {
+            $now = new \DateTimeImmutable($now->format(\DateTimeInterface::ATOM));
+        }
+        $now = $now->setTimezone(new \DateTimeZone('GMT'));
+        $expires = $context->getExpirationDate() ?? $now->add(new \DateInterval('PT1H'));
 
         $this->handleSignature($request, $credentials, $now, $expires, true);
     }
 
     public function sign(Request $request, Credentials $credentials, RequestContext $context): void
     {
-        $now = $context->getCurrentDate() ?? new \DateTimeImmutable();
+        if (null === $now = $context->getCurrentDate()) {
+            $now = new \DateTimeImmutable();
+        } else {
+            $now = new \DateTimeImmutable($now->format(\DateTimeInterface::ATOM));
+        }
+        $now = $now->setTimezone(new \DateTimeZone('GMT'));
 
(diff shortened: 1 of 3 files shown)


stof:
The NotExistsWaiter commit does not belong to this PR, right ?

src/Core/src/Signer/SignerV4.php
```diff
@@ -59,15 +59,25 @@ public function __construct(string $scopeName, string $region)
...
     public function presign(Request $request, Credentials $credentials, RequestContext $context): void
     {
-        $now = $context->getCurrentDate() ?? new \DateTimeImmutable();
-        $expires = $context->getExpirationDate() ?? (new \DateTimeImmutable($now->format(\DateTimeInterface::ATOM)))->add(new \DateInterval('PT1H'));
+        if (null === $now = $context->getCurrentDate()) {
+            $now = new \DateTimeImmutable();
+        } else {
+            $now = new \DateTimeImmutable($now->format(\DateTimeInterface::ATOM));
+        }
+        $now = $now->setTimezone(new \DateTimeZone('GMT'));
```
stof: why `GMT` rather than `UTC` (which is what `Z` as timezone indicator means) ?

src/Core/src/Stream/RewindableStream.php
```diff
@@ -69,27 +64,11 @@ public function getIterator(): \Traversable
             return;
         }
 
-        $resource = \fopen('php://memory', 'r+b');
+        $resource = \fopen('php://temp', 'r+b');
```
Nyholm: This is awesome. I didn't know this. 

Thank you @stof.

src/Core/src/Signer/SignerV4.php
jderusse: dummy copy/paste from official AWS SDK that use GMT.

Indeed, the doc refer to UTC https://docs.aws.amazon.com/general/latest/gr/sigv4-date-handling.html.

Fixed

src/Core/src/Signer/SignerV4.php
```diff
@@ -98,19 +110,19 @@ private function handleSignature(Request $request, Credentials $credentials, \Da
...
 
-        // signature is passed by reference to convertBodyToStream
-        $signature = '';
         if ($isPresign) {
             // Should be called before `buildBodyDigest` because this method may alter the body
             $this->convertBodyToQuery($request);
         } else {
+            // $signature does not exists but passed by reference then computed buildSignature
+            $signature = '';
             $this->convertBodyToStream($request, $now, $credentialString, $signingKey, $signature);
```
Nyholm: Why don’t we return the signature instead of passing it as reference?

jderusse: this method need the signature generated by the BuildSignature method Bellow.
And the BuildSignature method need headers generated by this method.
Their is a circular dependency between both method

stof: but then, does it work properly at all ?

jderusse: yes it works. Thanks to the reference.

here is the simplified workflow
```php
function convertBodyToStream() {
  $request->setHeader('xxx'); // change something in the header

  // request getBody will be called later when httpClient will sent the request
  $request->setBody(function() use ($signature) {
   //do someting with the signature
  });
}
function buildSignature()
{
  // need the headers to build the signature
  $signature = xx($request->getHeaders(), $args)
}
```

PR #361, opened 2020-03-23
FixedSizeStream decorating FixedSizeStream

### LINKED ISSUE #359: FixedSizeStream::create does not respect the chunk size when wrapping a FixedSizeStream
The factory method has an optimization to return the wrapped stream directly if it is already a FixedSizeStream. But that should be done only if the provided FixedSizeStream has the same fixed size than the requested one IMO.

jderusse:
Fixes #359

### DIFF
--- a/src/Core/tests/Unit/Stream/FixedSizeStreamTest.php
+++ b/src/Core/tests/Unit/Stream/FixedSizeStreamTest.php
@@ -43,6 +43,16 @@ class FixedSizeStreamTest extends TestCase
         self::assertSame($expected, \iterator_to_array($stream));
     }
 
+    /**
+     * @dataProvider provideChunks
+     */
+    public function testDecoratingFixedSize(RequestStream $content, int $size, array $expected): void
+    {
+        $stream = FixedSizeStream::create(FixedSizeStream::create($content, \random_int(1, 10)), $size);
+
+        self::assertSame($expected, \iterator_to_array($stream));
+    }
+
     public function provideLengths(): iterable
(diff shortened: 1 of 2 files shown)


src/Core/tests/Unit/Stream/FixedSizeStreamTest.php
```diff
@@ -43,6 +43,16 @@ public function testChunk(RequestStream $content, int $size, array $expected): v
         self::assertSame($expected, \iterator_to_array($stream));
     }
 
+    /**
+     * @dataProvider provideChunks
+     */
+    public function testDecoratingFixedSize(RequestStream $content, int $size, array $expected): void
+    {
+        $stream = FixedSizeStream::create(FixedSizeStream::create($content, \random_int(1, 10)), $size);
```
Nyholm: We don’t need the inner stream to be random here. Do we?

We don’t want our tests to fail randomly

jderusse: it don't have to be random, but it should work whatever the value of the inner chunk size.
if the test failed randomly, then their is a bug

stof: well, we should have 2 non-random tests: one using the same size, and one using always a different size.

jderusse: removed.

Nyholm:
Could you rebase and merge please?

PR #365, opened 2020-03-26
Use DateTimeImmutable instead of DateTimeInterface

### LINKED ISSUE #363: [RFC] Use DateTimeImmutable everywhere instead of DateTimeInterface
`DateTimeInterface` only contains the methods to format dates (`format`) and getters (`getTimestamp`, `getTimezone`, ...), as there are the only methods working exactly the same in `DateTime` and `DateTimeImmutable`. This makes the code a lot more complex when needing to manipulate dates (see the preSign logic for instance).
But allowing any DateTimeInterface has another drawback: the immutable value objects are not actually immutable. If I instantiate one of them and I provide a DateTime object, I can then actually mutate the date in this object.

I propose the following change:

- use `DateTimeImmutable` as type for all properties and return values in the library
- maybe keep accepting `DateTime` as input in constructors (either as a deprecated BC layer, or as a supported API), but convert the value to `DateTimeImmutable` (as done for string values)
- change argument types in internal APIs to use `DateTimeImmutable` (for the SignerV4 private APIs, I think that could actually already be done today, as it performs some conversion in public entrypoints anyway

What do you think @jderusse @Nyholm ?

Nyholm:
This will fix #363.


src/Core/src/Signer/SignerV4.php
```diff
@@ -63,9 +63,8 @@ public function presign(Request $request, Credentials $credentials, RequestConte
     {
         if (null === $now = $context->getCurrentDate()) {
             $now = new \DateTimeImmutable();
-        } else {
-            $now = new \DateTimeImmutable($now->format(\DateTimeInterface::ATOM));
         }
+
```
jderusse:
```suggestion
        $now = $context->getCurrentDate() ?? new \DateTimeImmutable();
```

src/Core/src/Signer/SignerV4.php
```diff
@@ -77,9 +76,8 @@ public function sign(Request $request, Credentials $credentials, RequestContext
...
```
jderusse:
```suggestion
        $now = $context->getCurrentDate() ?? new \DateTimeImmutable();
```

Nyholm:
> I just wonder if we shouldn't allow \DateTimeInteface in input then convert the user input in \DateTimeImmutable

You mean Input objects, right? 
Because "inputs" like arguments to `presign` just caused these issues. 

Im not sure it is worth the trouble. Either way it can be added in the future without breaking BC.

src/Core/src/Signer/SignerV4.php
Nyholm: I could not apply these with the GUI

Screenshot 2020-03-27 at 00 47 45

PR #366, opened 2020-03-26
Created website with docs

Nyholm:
It is not pretty, but that can be fixed later. 

https://async-aws.com/

I think I like this structure. No content is generated. 

---------

The generator looks at the menu structure in couscous.yml, then it looks at all Markdown files in /docs and converts them to HTML. 

The "design" is in website/template. It is normal twig. 
The assets lives in website/assets, it is normal webpack.


.github/workflows/website.yml
```diff
@@ -0,0 +1,56 @@
+name: Build website
```
Nyholm: Im not 100% sure this is working. I've not tested it

.github/workflows/website.yml
```diff
@@ -0,0 +1,56 @@
...
+      run: |
+        curl -OS http://couscous.io/couscous.phar
+        chmod +x couscous.phar
+
+    - name: Generate website
+      run: |
+        ./couscous.phar generate --target website aws
+        echo "async-aws.com" > CNAME
+
+    - name: Create Pull Request
```
Nyholm: I test this with a PR. I want to push it automatically later.

.github/workflows/website.yml
Nyholm: I've tested it

Nyholm:
The only thing missing now is permission for `secrets.SSH_PRIVATE_KEY` to push to https://github.com/async-aws/async-aws.github.io. 

@jderusse can you merge this PR when that is fixed?

PR #368, opened 2020-03-27
[Lambda] Invoke several concurrent Lambdas (non blocking)

gmorel:
#### Pain Point

We would like to use lambda this way:
https://twitter.com/matthieunapoli/status/1239960294736031744

Basically spawning 10 lambdas and don't process lambda replies in the order they were sent.
Lambda example:
```php
<?php declare(strict_types=1);

require __DIR__ . '/vendor/autoload.php';

return function ($event) {
    $rand = rand (1, 4);
    sleep($rand);

    return 'Hello ' . ($event['name'] ?? 'world') . ': ' . $rand . 's';
};

```

Introduce LambdaClient->invokeBatchAsync()
    Using `HttpClient->stream()`
https://symfony.com/doc/current/components/http_client.html#multiplexing-responses
    
    If you have 10 Lambda triggered: 1, 2, 3, ..., 9, 10
    And each of theses Lambda can reply from 1sec to 4sec
    It won't wait these 10 Lambda to complete to get their result
    For example: 4, 2, 6, ..., 5, 1
    And will call `BatchAsyncClosureInterface` closure async
    
    Usage:
```php
            $requests = [];
            for ($i = 0; $i < 10; ++$i) {
                $requests[] = new InvocationRequest([
                    'FunctionName' => 'app-dev-hellow_world',
                    'Payload' => "{\"name\": $i}",
                ]);
            }
    
            $callback = new BatchAsyncClosure();
    
            $client = $this->createClient($this->httpClient);
            $client->invokeBatchAsync($this->httpClient, $requests, $callback);
```   

```php
class BatchAsyncClosure implements BatchAsyncClosureInterface
{
    public function __invoke(InvocationResponse $invocationResponse): void
    {
        dump($invocationResponse->getPayload());
    }
} 
```

    "Hello 4: 1s"
    "Hello 2: 1s"
    "Hello 6: 1s"
    "Hello 3: 2s"
    "Hello 9: 2s"
    "Hello 7: 2s"
    "Hello 8: 3s"
    "Hello 5: 3s"
    "Hello 0: 4s"
    "Hello 1: 4s"

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -19,6 +19,7 @@ use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Symfony\Component\HttpClient\HttpClient;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Contracts\HttpClient\ResponseInterface;
 
 /**
  * Base class all API clients are inheriting.
@@ -100,6 +101,17 @@ abstract class AbstractApi
 
     final protected function getResponse(Request $request, ?RequestContext $context = null): Response
     {
+        return new Response(
+            $this->createHttpResponse($request, $context),
+            $this->httpClient
+        );
+    }
+
+    /**
+     * @throws \Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface
+     */
+    final protected function createHttpResponse(Request $request, ?RequestContext $context = null): ResponseInterface
+    {
         $request->setEndpoint($this->getEndpoint($request->getUri(), $request->getQuery()));
 
         if (null !== $credentials = $this->credentialProvider->getCredentials($this->configuration)) {
@@ -117,9 +129,7 @@ abstract class AbstractApi
             $requestBody = $requestBody->stringify();
         }
 
-        $response = $this->httpClient->request($request->getMethod(), $request->getEndpoint(), ['headers' => $request->getHeaders(), 'body' => 0 === $length ? null : $requestBody]);
-
-        return new Response($response, $this->httpClient);
+        return $this->httpClient->request($request->getMethod(), $request->getEndpoint(), ['headers' => $request->getHeaders(), 'body' => 0 === $length ? null : $requestBody]);
     }
 
     /**
--- a/src/Service/Lambda/src/Closure/BatchAsyncClosureInterface.php
+++ b/src/Service/Lambda/src/Closure/BatchAsyncClosureInterface.php
@@ -1,0 +1,29 @@
+<?php
+declare(strict_types=1);
+
+namespace AsyncAws\Lambda\Closure;
+
+use AsyncAws\Lambda\Result\InvocationResponse;
+
+/**
+ * Ease invokeBatchAsync implementation
+ * Allow unit testing of the callable implementation
+ *
+ * Example:
+ *
+ * // Instead of untestable Business Logic
+ * $callback = function(InvocationResponse $invocationResponse) {
+ *     // Business Logic
+ * };
+ * $callback($invocationResponse);
+ *
+ * // Testable Business Logic
+ * $callback = new BatchAsyncClosure(); // Implements BatchAsyncClosureInterface
+ * $callback($invocationResponse);
+ *
+ * @author Guillaume MOREL <me@gmorel.io>
+ */
+interface BatchAsyncClosureInterface
+{
+    public function __invoke(InvocationResponse $invocationResponse): void;
+}
--- a/src/Service/Lambda/src/LambdaClient.php
+++ b/src/Service/Lambda/src/LambdaClient.php
@@ -4,14 +4,17 @@ namespace AsyncAws\Lambda;
 
 use AsyncAws\Core\AbstractApi;
 use AsyncAws\Core\RequestContext;
+use AsyncAws\Core\Response;
 use AsyncAws\Lambda\Input\AddLayerVersionPermissionRequest;
 use AsyncAws\Lambda\Input\InvocationRequest;
 use AsyncAws\Lambda\Input\ListLayerVersionsRequest;
 use AsyncAws\Lambda\Input\PublishLayerVersionRequest;
+use AsyncAws\Lambda\Closure\BatchAsyncClosureInterface;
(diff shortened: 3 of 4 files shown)


src/Core/src/AbstractApi.php
```diff
@@ -99,6 +100,17 @@ public function __construct($configuration = [], ?CredentialProvider $credential
     abstract protected function getSignatureScopeName(): string;
 
     final protected function getResponse(Request $request, ?RequestContext $context = null): Response
```
gmorel: No BC break

src/Core/src/AbstractApi.php
```diff
@@ -99,6 +100,17 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
+        return new Response(
+            $this->createHttpResponse($request, $context),
+            $this->httpClient
+        );
+    }
+
+    /**
+     * @throws \Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface
+     */
+    final protected function createHttpResponse(Request $request, ?RequestContext $context = null): ResponseInterface
```
gmorel: New method to get the HttpResponse without wrapper
I was stuck because of https://github.com/async-aws/aws/blob/master/src/Core/src/Response.php#L46-L51

src/Service/Lambda/src/LambdaClient.php
```diff
@@ -62,6 +65,41 @@ public function invoke($input): InvocationResponse
...
+     * @see https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-lambda-2015-03-31.html#invoke
+     * @see https://symfony.com/doc/current/components/http_client.html#concurrent-requests
+     *
+     * @param InvocationRequest[] $inputs
+     *
+     * @return InvocationResponse[]
+     *
+     * @throws \Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface
+     */
+    public function invokeBatchAsync(HttpClientInterface $httpClient, array $inputs, BatchAsyncClosureInterface $callback): array
```
gmorel: New method

gmorel: I have to inject HttpClientInterface as it is in private in the AbstractAPI

src/Service/Lambda/tests/Integration/LambdaClientTest.php
```diff
@@ -44,7 +44,7 @@ public function testInvoke(): void
             'FunctionName' => 'Index',
             'Payload' => '{"name": "jderusse"}',
         ]);
-        $result = $client->Invoke($input);
+        $result = $client->Invoke($input); // @todo Find why the capital I ?
```
gmorel: Shall I rename to camelCase ?

src/Service/Lambda/tests/Integration/LambdaClientTest.php
```diff
@@ -55,6 +55,19 @@ public function testInvoke(): void
         self::assertSame('$LATEST', $result->getExecutedVersion());
     }
 
+    public function testInvokeBatchAsync(): void
+    {
+        $client = $this->getClient();
+
+//        $input = new InvocationRequest([
```
gmorel: If you have time I would need help to understand how you test this
- Run tests
- How I can lead `http://localhost:9001` to return something. Or shall I prefer using mocking ?

src/Service/Lambda/src/Closure/BatchAsyncClosureInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * // Testable Business Logic
+ * $callback = new BatchAsyncClosure(); // Implements BatchAsyncClosureInterface
+ * $callback($invocationResponse);
+ *
+ * @author Guillaume MOREL <me@gmorel.io>
+ */
+interface BatchAsyncClosureInterface
+{
+    public function __invoke(InvocationResponse $invocationResponse): void;
```
gmorel: Implementation example
```php
class BatchAsyncClosure implements BatchAsyncClosureInterface
{
    /**
     * @inheritDoc
     */
    public function __invoke(InvocationResponse $invocationResponse): void
    {
        dump(
            $invocationResponse->getPayload() 
        );
    }
}
```

Nyholm:
Thank you for this PR. Make note that most of the code in this library is generated. See https://async-aws.com/contribute/ That is also one reason the CI fails. (and that we have some weird code style)


I like the idea of batching requests like this. Dont you achieve the same result with code like: 

```php

            $results = [];
            for ($i = 0; $i < 10; ++$i) {
                $results[] = $client->invoke(new InvocationRequest([
                    'FunctionName' => 'app-dev-hellow_world',
                    'Payload' => "{\"name\": $i}",
                ]));
            }
    
            while (!empty($results)) {
                      if ($result->resolve(0)) {
                          echo $result->getPayload();
                          unset($results[$i]);
                     }
              }
         }
```

gmorel:
Very clever @Nyholm, it works :rocket: 
Screenshot from 2020-03-27 15-29-17

Without the `while (!empty($results)) {` it is no more working. Is it the trick in order to avoid being garbage collected ?

Thanks :+1: I will remove the code and update the documentation instead.

Keep up the good work !

Nyholm:
I took the liberty to add it to the docs: https://github.com/async-aws/aws/blob/master/docs/features/async.md

> Without the while (!empty($results)) { it is no more working. Is it the trick in order to avoid being garbage collected ?

If you dont have the while loop you will just go over your responses once. Since it is non-blocking none will print any output. Try adding a loop counter and print it each time you restart the while loop. You are running that plenty of iterations.

gmorel:
Indeed it took 1263 iterations.

I close the PR then.

Nyholm:
Is there a difference if you do `if ($result->resolve(0.01)) {` instead?

Ie move CPU power from PHP to curl. It is not "non-blocking" since it blocks for 0.01 seconds.

gmorel:
I got this with 100 concurrent lambdas
Both cases are working.

`if ($result->resolve(0.0)) {`

0 0-x100

-----------------------------

`if ($result->resolve(0.01)) {`

0 01-x100

Nyholm:
Have a look at #417. 

This feature is implemented there and it is soon to be merged.

PR #370, opened 2020-03-27
 Adding a bit nicer commit message

Nyholm:
This will show what commit on async-aws/aws that triggered the website update

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -44,5 +44,5 @@ jobs:
         cd website
         git config --local user.email "github@async-aws.com"
         git config --local user.name "AsyncAWS Bot"
-        git commit -m "Update website" -a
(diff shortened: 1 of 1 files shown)


.github/workflows/website.yml
```diff
@@ -44,5 +44,5 @@ jobs:
         cd website
         git config --local user.email "github@async-aws.com"
         git config --local user.name "AsyncAWS Bot"
-        git commit -m "Update website" -a
+        git commit -m "Update website from commit `git --git-dir=./../aws/.git rev-parse HEAD`" -a
```
jderusse: Couldn't it be `${{ github.ref }}`?

Nyholm: I could maybe use `github.sha`

PR #371, opened 2020-03-27
Make sure commit reference works across repos

Nyholm:
¯\\_(ツ)_/¯

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -14,6 +14,7 @@ jobs:
 
       - name: Download dependencies
         run: |
+          echo "Update website from commit async-aws/aws@${{ github.sha }}"
           composer update --no-interaction --prefer-dist --optimize-autoloader
 
       - name: PHPStan
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -12,11 +12,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Download dependencies
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/website.yml
```diff
@@ -44,5 +44,5 @@ jobs:
         cd website
         git config --local user.email "github@async-aws.com"
         git config --local user.name "AsyncAWS Bot"
-        git commit -m "Update website from commit `git --git-dir=./../aws/.git rev-parse HEAD`" -a
+        git commit -m "Update website from commit async-aws/aws@`git --git-dir=./../aws/.git rev-parse HEAD`" -a
```
Nyholm: Should I update this to `github.sha`?

jderusse: if it work, yes, it's more readable IMO

.github/workflows/static.yml
```diff
@@ -12,10 +12,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Download dependencies
-        run: |
-          composer update --no-interaction --prefer-dist --optimize-autoloader
```
Nyholm: There is no need for composer update here. 

See https://github.com/OskarStark/phpstan-ga/blob/master/entrypoint.sh#L8

Nyholm: Hm... I was clearly wrong.. strange

jderusse: because of no-dev.

but looks like they added a new `$REQUIRE_DEV` variables.

PR #375, opened 2020-03-27
Use fork in automated PR

jderusse:


.github/workflows/watch.yml
```diff
@@ -1,6 +1,9 @@
 name: Watch for changes
 
 on:
+  push:
+    branches:
+    - bot-fork  # do not forget to revert
```
Nyholm: This will only work on your local fork. Do you have all secrets there? 

Or should you push to async-aws/aws:bot-fork instead?

jderusse: > Or should you push to async-aws/aws:bot-fork instead
This is what I do

jderusse:
RFR

.github/workflows/watch.yml
```diff
@@ -2,7 +2,7 @@ name: Watch for changes
 
 on:
   schedule:
-    - cron: '0 6 * * *'
+  - cron: '0 6 * * *'
```
Nyholm: Invalid indentation. Please revert.

PR #376, opened 2020-03-27
Adding naive implementation of multiplexing responses

Nyholm:
This is an alternative to what #368 is trying to do. 

Im not sure this is a good idea because: 
1. It exposes Symfony responses to the user
2. The user has no idea what response that is related to what result
3. There is no result classes anymore

Could we build on to this somehow? 

I dont want to merge it in the current state

### DIFF
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -15,6 +15,7 @@ use AsyncAws\Core\Stream\ResultStream;
 use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
+use Symfony\Contracts\HttpClient\ResponseStreamInterface;
 
 /**
  * The response provides a facade to manipulate HttpResponses.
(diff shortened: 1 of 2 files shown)


jderusse:
I 100% agreed. We should not return HttpClient namespace to the user.

What are we trying to solve? Does it make sens to give such granularity to the user? I means does it make send to return StreamInterface to let user acccess to chunks?

IMHO users maybe interested by:
- status code (status/failure)
- full response (perform something with the content of the response)

Status Code: could be achieved by the snippet you gave in https://github.com/async-aws/aws/pull/368#issuecomment-604945602
We don't provide tools for the second one. I see 3 options:
- provide a `resolveFull` method that wait for last chunk (instead of first chunk actually)
- add a parameter `resolve(float $timeout, bool $full false)`
- change the behavior of `resolve` that always wait for last chunk

---

In addition, the loop over all responses are performed by user (the user can resolve 1 response a the time), We could provide a method that resolve all the response in parallel.

actually, we have to does such code
```php
while (!empty($results)) {
    if ($result->resolve(0.01)) {
      echo $result->getPayload();
      unset($results[$i]);
    }
  }
}
```
`resolve(0.01)` is not optimal: 0.01 avoid CPU burning, but if you have 100 responses we can wait up to a second.

A better solution would be
```php
foreach (Result::resolveAll($results, 30.0) => $result) {

}
```
resolveAll yields resolved responses within the given timeout.
```php
public static function resolveAll(iterable $responses, float $timeout = null): iterable
{
    $httpResponses = [];
    $httpClient = null;
        if (null === $httpClient) {
            $httpClient = $response->httpClient;
        }
        $httpResponses[$response->httpResponse] = $response;
    }

    foreach ($httpClient->stream(array_keys($httpResponses), $timeout) as $httpResponse => $chunk) {
        if ($chunk->isTimeout()) {
          continue; // return?
        }
        if ($chunk->isFirst()) {
          yield $httpResponses[$httpResponse];
        }
    }
}
```

Nyholm:
Thank you. I'll reply on your PRs.

Nyholm:
Im closing this in favor of #417

PR #385, opened 2020-03-28
Client and integration docs

Nyholm:
I feel like we have a good start with our docs now. There are still much that could be improved but after this PR is merged we have both a good structure and some basic docs.


Nyholm:
Finally done with the PR. Im now very happy with the docs =)

I've read all articles many times. I cannot produce anything better at this time.

If only the website was updated automatically.... Waiting for github support to get back to me.

PR #386, opened 2020-03-28
Use custom twig extensions to use manifest.json

Nyholm:
Blocked by https://github.com/CouscousPHP/Couscous/pull/248

UPDATE:
This PR uses a forked couscous.phar.

### DIFF
--- a/website/Twig/AsyncAwsTwigExtension.php
+++ b/website/Twig/AsyncAwsTwigExtension.php
@@ -82,7 +82,7 @@ class AsyncAwsTwigExtension extends AbstractExtension
 
     public function renderStyle(string $name)
     {
-        return $this->renderTag($name, 'css', '<link rel="stylesheet" href="%s"/>');
+        return $this->renderTag($name, 'css', '<link rel="stylesheet" href="%s">');
     }
 
     public function renderScript(string $name)
--- a/website/Twig/AsyncAwsTwigExtension.php
+++ b/website/Twig/AsyncAwsTwigExtension.php
@@ -100,7 +100,7 @@ class AsyncAwsTwigExtension extends AbstractExtension
         }
 
         foreach ($entrypoints['entrypoints'][$name][$prefix] as $file) {
-            $output.= sprintf($template, $file);
+            $output.= sprintf($template, htmlspecialchars($file));
         }
 
         return $output;
--- a/website/Twig/AsyncAwsTwigExtension.php
+++ b/website/Twig/AsyncAwsTwigExtension.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Website;
+
+use Twig\Extension\AbstractExtension;
+
+class AsyncAwsTwigExtension extends AbstractExtension
+{
+    /**
+     * @var array|null
+     */
+    private $manifestCache;
+
+    public function getFunctions()
+    {
+        return [
+            new \Twig_SimpleFunction('asset', [$this, 'getAssetUrl'])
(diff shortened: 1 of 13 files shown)


website/Twig/AsyncAwsTwigExtension.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        $output = '';
+
+        $entrypoints = $this->getEntrypoints();
+        if (!isset($entrypoints['entrypoints'][$name])) {
+            throw new \RuntimeException(sprintf('Could not find entrypoints.json entry for "%s"', $name));
+        }
+
+        foreach ($entrypoints['entrypoints'][$name][$prefix] as $file) {
+            $output.= sprintf($template, $file);
```
stof: this is missing the escaping of `$file` in case it contains special chars.

Nyholm: Can $file contain special chars? I believe `entrypoints.json` contain the correct urls

Nyholm: Maybe you are correct. But as long as we are using "app" as our entry name, arn't we safe? 

```
{
  "entrypoints": {
    "app": {
      "css": [
        "/assets/0.3639a826.css",
        "/assets/app.9f119eed.css"
      ],
      "js": [
        "/assets/0.24ed7736.js",
        "/assets/app.c7303778.js"
      ]
    }
  }
}
```

stof: Well, I would not guarantee that no chars in it need HTML escaping.

As we know our templates, we could do `htmlspecialchars($file)`

Nyholm: Cool. I've updated the PR.

website/Twig/AsyncAwsTwigExtension.php
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * @var array|null
+     */
+    private $entrypointsCache;
+
+    public function getFunctions()
+    {
+        return [
+            new \Twig_SimpleFunction('asset', [$this, 'getAssetUrl']),
+            new \Twig_SimpleFunction('renderStyle', [$this, 'renderStyle'], [
```
stof, line 25: use `\Twig\TwigFunction` to be foward-compatible with new Twig versions (Twig 3 uses only namespaced classes)

website/Twig/AsyncAwsTwigExtension.php
```diff
@@ -0,0 +1,108 @@
...
+            new \Twig_SimpleFunction('renderScript', [$this, 'renderScript'], [
+                'is_safe' => ['html'],
+            ]),
+        ];
+    }
+
+    public function getFilters()
+    {
+        return [
+            new \Twig_SimpleFilter('asset', [$this, 'parseAssetUrls']),
```
stof, line 37: same here for `\Twig\TwigFilter`

website/Twig/AsyncAwsTwigExtension.php
Nyholm: Couscous is using Twig 1. Is there a forward bridge from Twig 1 so I can use namespaces?

Nyholm: Yes, it does. https://github.com/async-aws/aws/pull/427

stof: thus, you already use the namespace for the AbstractExtension base class :smile:

PR #390, opened 2020-03-29
Fixed broken CI

Nyholm:


.github/workflows/website.yml
```diff
@@ -38,6 +38,7 @@ jobs:
     - name: Generate website
       run: |
         ./couscous.phar generate --target website aws
+        cd aws/website && npm run highlight `pwd`/../../website/.couscous/generated
```
Nyholm: Is this the correct path?

website/hightlight/main.js
```diff
@@ -51,12 +51,14 @@ walk(outputDir, function(err, files) {
         .forEach(function(file) {
             JSDOM.fromFile(file, {}).then(dom => {
                 dom.window.document.querySelectorAll('pre code').forEach((block) => {
-                    highlight.highlightBlock(block);
+                    if (!block.classList.contains('hljs')) {
+                        highlight.highlightBlock(block);
+                    }
```
Nyholm: This if block make sure we can run highlight multiple times on the same HTML

Makefile
```diff
@@ -39,5 +39,7 @@ website/template/assets/app.css: website/node_modules website/assets/*
 website/node_modules: website/package.json
 	cd website && npm install
 
-website-highlight:
-	cd website && npm run highlight `pwd`/../.couscous/generated
+website-highlight: website/node_modules
+ifneq (,$(wildcard .couscous/generated/index.html))
+	cd website && npm run highlight ${PWD}/.couscous/generated;
+endif
```
Nyholm: Check if file exists. 

This make sure CI dont fail. When generating the docs on CI there will not be any files in this folder.

.github/workflows/website.yml
```diff
@@ -24,11 +24,11 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
       with:
-        path: 'aws'
+        path: 'source'
```
Nyholm: Now then main repo will be checked out in `/home/runner/work/aws/aws/source` instead of `/home/runner/work/aws/aws/aws`.

PR #398, opened 2020-03-29
Use Bot's token instead of SSH Key

jderusse:
Use Bot's Personal Access Token instead of SSH Key (reduce number of secrets)

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -16,19 +16,15 @@ jobs:
         with:
           php-version: 7.4
 
-      - name: Set up SSH
-        uses: webfactory/ssh-agent@v0.2.0
-        with:
-          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
-
       - name: Checkout code
         uses: actions/checkout@v2
         with:
           path: 'source'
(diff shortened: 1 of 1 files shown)


.github/workflows/website.yml
```diff
@@ -16,19 +16,15 @@ jobs:
...
         uses: actions/checkout@v2
         with:
           path: 'source'
+          token: ${{ secrets.BOT_GITHUB_TOKEN }}
 
-      - name: Checkout Website repo via SSH
+      - name: Checkout Website repo
         run: |
-          git clone --branch master git@github.com:async-aws/async-aws.github.io.git output
+          git clone --branch master http://github.com/async-aws/async-aws.github.io output
```
Nyholm: Could you update the URL?

jderusse: thanks, fixed.

Nyholm: Could you rebase, it seams like there is a conflict?

Nyholm:
This seamed to fail: 

https://github.com/async-aws/aws/runs/543246103?check_suite_focus=true#step:7:36

PR #399, opened 2020-03-29
Show that we "intend" to add all files

Nyholm:
https://stackoverflow.com/questions/855767/can-i-use-git-diff-on-untracked-files

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -51,6 +51,7 @@ jobs:
           git status
           echo "::endgroup::"
 
+          git add -N .
           if [[ $(git diff --numstat | wc -l) -eq 0 ]]; then
             echo "No changes found. Exiting."
             exit 0;


.github/workflows/website.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           git status
           echo "::endgroup::"
 
+          git add -N .
```
jderusse: should we use same for `watch.yaml`?

Nyholm:
Woho!. It worked!

https://async-aws.com/

.github/workflows/website.yml
Nyholm: Sure, why not? 

But it is unlikely that the watch command generates a new file.

jderusse: Replace string by Enum class for instance..
But yeah, in that case, at least one file would probably changed too

PR #402, opened 2020-03-29
Dont use token for website build

Nyholm:
Im not sure there is any benefit. Also, using token makes the build fail =)


Nyholm:
Using token is fine.

PR #403, opened 2020-03-29
[S3] MultipartUpload

### LINKED ISSUE #337: Add support for uploading big files on S3
`PutObject` is limited to 5GB. Bigger objects (up to 5TB) require using a multipart upload. The official SDK has a `upload()` method, which handles this transparently (by default, it will switch to multipart much earlier than the 5GB threshold, probably because they found it more efficient or more reliable to upload in chunks). Does it make sense to provide the same kind of feature in this SDK ?

Nyholm:
When looking as #337 I found these API endpoints. This PR has just generated the endpoints. I have not tested them yet. 

This will fix #337


Nyholm:
So, this adds endpoints for handling multipart uploads. https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html

There is no "automatic switching" between uploading files <5Gb and files >5Gb. There is also no nice helper class with a pleasant API. 

I think that should be a separate package if we want it. It should not be in async-aws/s3. But I dont mind to have it in the organization.

src/CodeGenerator/src/Generator/RequestSerializer/RestXmlSerializer.php
```diff
@@ -105,6 +105,7 @@ private function dumpXmlShape(Member $member, Shape $shape, string $output, stri
         switch ($shape->getType()) {
             case 'blob':
             case 'string':
+            case 'integer':
```
Nyholm: This is the only change I made in the generator. 

Im not 100% convinced that it is correct. I'll check that later.

jderusse: should work. Except if people provide non-base10 integer 

Maybe you could add test that include one of those
```
$a = 0123;
$a = 0x1A;
$a = 0b11111111;
```

src/Service/S3/tests/Unit/Result/CompleteMultipartUploadOutputTest.php
```diff
@@ -0,0 +1,32 @@
...
+ <ETag>"3858f62230ac3c915f300c664312c11f-9"</ETag>
+</CompleteMultipartUploadResult>');
+
+        $client = new MockHttpClient($response);
+        $result = new CompleteMultipartUploadOutput(new Response($client->request('POST', 'http://localhost'), $client));
+
+        self::assertSame('http://Example-Bucket.s3.eu-central-1.amazonaws.com/Example-Object', $result->getLocation());
+        self::assertSame('Example-Bucket', $result->getBucket());
+        self::assertSame('Example-Object', $result->getKey());
+        self::assertSame('3858f62230ac3c915f300c664312c11f-9', $result->getETag());
```
jderusse: should be `"3858f62230ac3c915f300c664312c11f-9"` (see my comment https://github.com/async-aws/aws/issues/404#issuecomment-605656957)

src/Service/S3/src/ValueObject/CompletedPart.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @param array{
+     *   ETag?: null|string,
+     *   PartNumber?: null|int,
+     * } $input
+     */
+    public function __construct(array $input)
+    {
+        $this->ETag = $input['ETag'] ?? null;
```
Nyholm: Should we somehow verify the `Etag` that is is quoted? Or dont we care?

jderusse: how does other SDK works?

Nyholm: From what I can see, the official API do nothing at all regarding ETags. I dont think we should either

src/CodeGenerator/src/Generator/RequestSerializer/RestXmlSerializer.php
Nyholm: Those should be handled the same in the XML.

PR #406, opened 2020-03-29
Add documentation on clients

jderusse:


jderusse:
agreed, just the most common endpoint, or advance usages

Nyholm:
Yeah. I like the extra docs on s3 regarding streaming.

PR #407, opened 2020-03-29
Replace all occurrences of foobar by real examples

jderusse:


jderusse:
don't ! I'm on the same boat :-)

PR #411, opened 2020-03-30
Make ResponseBody rewindable

jderusse:
calling several time `ResponseBodyStream::getContentAsString` throws an exception (should be visible in prefer-lowest test).
This PR:
- use HttpClient `ResponseTrait::toStream` method that provide a stream when available
- otherwise use the `ResponseBodyStream` but store the stream in a `php://temp` resource in order to be re-read

### DIFF
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -160,7 +160,7 @@ class Response
     {
         $this->resolve();
 
-        if (is_callable([$this->httpResponse, 'toStream'])) {
+        if (\is_callable([$this->httpResponse, 'toStream'])) {
             return new ResponseBodyResourceStream($this->httpResponse->toStream());
         }
 
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -158,12 +158,11 @@ class Response
 
     public function toStream(): ResultStream
     {
+        $this->resolve();
+
         if (is_callable([$this->httpResponse, 'toStream'])) {
             return new ResponseBodyResourceStream($this->httpResponse->toStream());
         }
-
-        // ensure no exceptions are thrown
-        $this->resolve();
 
         return new ResponseBodyStream($this->httpClient->stream($this->httpResponse));
     }
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -10,6 +10,7 @@ use AsyncAws\Core\Exception\Http\NetworkException;
 use AsyncAws\Core\Exception\Http\RedirectionException;
 use AsyncAws\Core\Exception\Http\ServerException;
 use AsyncAws\Core\Exception\RuntimeException;
+use AsyncAws\Core\Stream\ResponseBodyResourceStream;
 use AsyncAws\Core\Stream\ResponseBodyStream;
 use AsyncAws\Core\Stream\ResultStream;
 use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
@@ -157,6 +158,13 @@ class Response
 
     public function toStream(): ResultStream
     {
+        if (\method_exists($this->httpResponse, 'toStream')) {
+            return new ResponseBodyResourceStream($this->httpResponse->toStream());
+        }
+
+        // ensure no exceptions are thrown
+        $this->resolve();
+
         return new ResponseBodyStream($this->httpClient->stream($this->httpResponse));
     }
 }
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -158,7 +158,7 @@ class Response
 
     public function toStream(): ResultStream
     {
-        if (\method_exists($this->httpResponse, 'toStream')) {
+        if (is_callable([$this->httpResponse, 'toStream'])) {
             return new ResponseBodyResourceStream($this->httpResponse->toStream());
         }
 
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -130,6 +130,10 @@ class Response
         $this->resolveResult = false;
     }
 
+    /**
+     * @throws NetworkException
+     * @throws HttpException
+     */
     public function getHeaders(): array
     {
         $this->resolve();
@@ -137,6 +141,10 @@ class Response
         return $this->httpResponse->getHeaders(false);
     }
 
+    /**
+     * @throws NetworkException
+     * @throws HttpException
+     */
     public function getContent(): string
     {
         $this->resolve();
@@ -144,6 +152,10 @@ class Response
         return $this->httpResponse->getContent(false);
     }
 
+    /**
+     * @throws NetworkException
+     * @throws HttpException
+     */
     public function toArray(): array
     {
         $this->resolve();
@@ -156,6 +168,10 @@ class Response
         return $this->httpResponse->getStatusCode();
     }
 
+    /**
+     * @throws NetworkException
+     * @throws HttpException
+     */
     public function toStream(): ResultStream
     {
         $this->resolve();
--- a/src/Core/src/Stream/ResponseBodyResourceStream.php
+++ b/src/Core/src/Stream/ResponseBodyResourceStream.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace AsyncAws\Core\Stream;
 
+use AsyncAws\Core\Exception\RuntimeException;
+
 /**
  * Stream a resource body.
  *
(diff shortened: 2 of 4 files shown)


src/Core/src/Stream/ResponseBodyResourceStream.php
```diff
@@ -0,0 +1,74 @@
...
+        $pos = \ftell($this->responseStream);
+
+        try {
+            if (!\rewind($this->responseStream)) {
+                throw new \InvalidArgumentException('Failed to rewind the stream');
+            }
+
+            while (!\feof($this->responseStream)) {
+                yield \fread($this->responseStream, 64 * 1024);
+            }
```
jderusse: I wonder If I shouldn't ftell/fseek between each yield, to ensure that:
- reset the pointer to the original place, in case the user call something that need to read/write the same stream
- restore the pointer to the last read position (for the same reason)

src/Core/src/Stream/ResponseBodyStream.php
```diff
@@ -33,45 +44,47 @@ public function __toString()
...
     public function getContentAsResource()
     {
-        $resource = \fopen('php://temp', 'rw+');
-
-        try {
-            foreach ($this->responseStream as $chunk) {
-                fwrite($resource, $chunk->getContent());
+        if (null === $this->fallback) {
+            // Use getChunks() to read stream content to $this->fallback
+            foreach ($this->getChunks() as $chunk) {
```
jderusse: That's a dummy implementation. A better solution would be to register a stream wrapper, and read the chunk ONLY when the stream is consumed

src/Core/src/Stream/ResponseBodyStream.php
```diff
@@ -10,6 +10,7 @@
  * Stream a HTTP response body.
```
Nyholm: We should add a comment here and on `ResponseBodyResourceStream` so one would understand the difference between them.

src/Core/src/Stream/ResponseBodyResourceStream.php
```diff
@@ -0,0 +1,94 @@
...
+            if (\feof($this->resource)) {
+                if (\fseek($this->resource, $outPos)) {
+                    throw new RuntimeException('The stream is not seekable');
+                }
+                yield $content;
+
+                break;
+            }
+
+            // move the cursor to original position and remember the reading position
```
Nyholm: If I read the stream, we are actually reading (or moving the cursor) the stream multiple times. 
If the stream in 10*64*1024 bytes (10 chunks). We are setting the cursor to the beginning 10 times. That is not needed. 


We dont need to reset the cursor on the "original position". We should make sure we read everything and don't care about other's reading the stream.

src/Core/src/Stream/ResponseBodyStream.php
```diff
@@ -33,45 +44,47 @@ public function __toString()
...
+            $this->resource = \fopen('php://temp', 'rb+');
+        }
         foreach ($this->responseStream as $chunk) {
-            yield $chunk->getContent();
+            $chunkContent = $chunk->getContent();
+            \fwrite($this->resource, $chunkContent);
+            yield $chunkContent;
         }
+
+        $this->fallback = new ResponseBodyResourceStream($this->resource);
```
Nyholm: What if I start reading the stream. Stop half ways and restart?

src/Core/src/Response.php
```diff
@@ -157,6 +158,13 @@ public function getStatusCode(): int
 
     public function toStream(): ResultStream
     {
+        if (\method_exists($this->httpResponse, 'toStream')) {
```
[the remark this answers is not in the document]
jderusse: nop, but it's not part of the interface: https://github.com/symfony/symfony/issues/34834

In 5.1, we the streamWrapper will be able to create the stream for us

Did you get Why I did that?
The `client->stream()` store is not rewindable. BUT it store the content in a private `php://temp` resource.
Here, I create a `ResultStream` using this private resource instead of the `$client->stream` callback

src/Core/src/Stream/ResponseBodyStream.php
jderusse: thanks

src/Core/src/Stream/ResponseBodyResourceStream.php
jderusse: Why not it would be much more easier.
but originally, my issue was: what if others trigger the code (HttpClient::steam) that write in the stream?

src/Core/src/Stream/ResponseBodyStream.php
jderusse: indeed... we have to consume the entire `responseStream`.. That defect the purpose of this class. we don't stream anythings. This should be deprecated an replaced by `ResponseBodyResourceStream`

Nyholm: Im happy to remove `ResponseBodyStream` if we always can use `ResponseBodyResourceStream` . 

If we cant, we should add a note that `ResponseBodyStream` is only a BC layer and document the edge case if you stop half ways.

jderusse: but we don't have choice... Moreover this class should not be used, people using a standard httpClient will use the other one.

Nyholm: That is okey. But it should be documented on the class. 
I was trying to say that by:

> we should add a note that ResponseBodyStream is only a BC layer and document the edge case if you stop half ways.

jderusse: comment added

src/Core/src/Stream/ResponseBodyStream.php
```diff
@@ -8,8 +8,11 @@
 
 /**
  * Stream a HTTP response body.
+ * This class is a NC layer for Http Client that does not supports toStream
```
Nyholm, line 11: BC

src/Core/src/Stream/ResponseBodyStream.php
```diff
@@ -8,8 +8,11 @@
...
+ * When calling `getChunks` or `getContentAsResource` it first, fully read download the Response Body in a blocking way.
```
Nyholm, line 12: I was fine with the previous implementation. I just wanted a comment that said that we could t restart reading half way. 

I’m fine with leaving that “bug” since it is an edge case

src/Core/src/Stream/ResponseBodyResourceStream.php
jderusse: nvm, the resource is used only by us. it should not be manipulated by somebody else

src/Core/src/Stream/ResponseBodyStream.php
jderusse: reverted. 
- added commetnts about the edge case
- added a check that throw an exception if user call getChunk several time before the end of the first

src/Core/src/Response.php
```diff
@@ -157,6 +158,12 @@ public function getStatusCode(): int
 
     public function toStream(): ResultStream
     {
+        $this->resolve();
```
Nyholm, line 161: Im okey with resolving this first. It could throw exceptions. 

We should add the exceptions to the doc block.

src/Core/src/Stream/ResponseBodyStream.php
```diff
@@ -4,12 +4,17 @@
 
 namespace AsyncAws\Core\Stream;
 
+use AsyncAws\Core\Exception\LogicException;
 use Symfony\Contracts\HttpClient\ResponseStreamInterface;
 
 /**
  * Stream a HTTP response body.
+ * This class is a BC layer for Http Response that does not supports toStream.
```
Nyholm, line 12:
```suggestion
 * This class is a BC layer for Http Response that does not support toStream().
```

Removed `supports` and make sure `toStream` looks like a function.

src/Core/src/Stream/ResponseBodyStream.php
```diff
@@ -33,45 +45,50 @@ public function __toString()
      */
     public function getChunks(): iterable
     {
+        if (null !== $this->fallback) {
+            return $this->fallback->getChunks();
+        }
+        if ($this->partialRead) {
+            throw new LogicException(\sprintf('You can not call "%s". Another process doesn\'t reading "getChunks" till  the end.', __METHOD__));
```
Nyholm, line 52: Nice!

src/Core/src/Response.php
jderusse: indeed. Fixed

Nyholm:
Thank you

PR #412, opened 2020-03-30
Resolve the full response

jderusse:
Provide a way to resolve the entire body, instead of just the first bytes.

That provide full async until the end of the response.

Partially fixes #376

### DIFF
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -38,11 +38,18 @@ class Response
     private $resolveResult;
 
     /**
-     * State of the response stream.
+     * State of the response chunk readability.
      *
      * @var bool
      */
     private $streamable = true;
+
+    /**
+     * State of the response fully streamed.
+     *
+     * @var bool
+     */
+    private $fullyResolved = false;
 
     public function __construct(ResponseInterface $response, HttpClientInterface $httpClient)
     {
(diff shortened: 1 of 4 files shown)


src/Core/src/Response.php
```diff
@@ -80,7 +88,17 @@ public function resolve(?float $timeout = null): bool
...
                 }
-                if ($chunk->isFirst()) {
+                if ($fullResponse) {
+                    if ($chunk->isLast()) {
+                        $this->streamable = false;
+
+                        break;
+                    }
+
+                    if ($this->streamable && '' !== $chunk->getContent()) {
```
Nyholm: Can you read a chunk multiple times?

jderusse: yes, you can read multiple time the same chunk, but can't rewind the stream

src/Core/src/Response.php
```diff
@@ -53,15 +60,16 @@ public function __destruct()
...
      *
      * @return bool whether the request is executed or not
      *
      * @throws NetworkException
      * @throws HttpException
      */
-    public function resolve(?float $timeout = null): bool
+    public function resolve(?float $timeout = null, bool $fullResponse = false): bool
     {
         if (null !== $this->resolveResult) {
```
jderusse: should check if fully resolved here

Nyholm:
Hm.. I dont think I like this. Im not sure you should care if the full response is downloaded. I think we could make some changes to #417 to take care of this problem.

Nyholm:
I might be wrong about this...

Nyholm:
I think this should be closed. It makes sense to have it when downloading more than one response. 
https://github.com/async-aws/aws/pull/417#discussion_r401197603

If you really really really want to use this. Then you can do `Result::multiplex([$result], 2.0);`

jderusse:
closed in favor of #417

PR #417, opened 2020-03-30
Add multiplexing method

jderusse:
Replacement of #376

If https://github.com/async-aws/aws/pull/412 is merged, this method should add the same `fullResponse` parameter

### DIFF
--- a/docs/features/async.md
+++ b/docs/features/async.md
@@ -151,6 +151,7 @@ their result. The result that is downloaded first will be printed first. The ord
 the requests are created do not matter.
 
 ```php
+use AsyncAws\Core\Result;
 use AsyncAws\Lambda\LambdaClient;
 use AsyncAws\Lambda\Input\InvocationRequest;
 
@@ -164,13 +165,8 @@ for ($i = 0; $i < 10; ++$i) {
     ]));
 }
 
-while (!empty($results)) {
-    foreach ($results as $i => $result) {
-        if ($result->resolve(0.01)) {
-            echo $result->getPayload();
-            unset($results[$i]);
-        }
-    }
+foreach (Result::resolveAll($results, null) as $result => $resolved) {
+    echo $result->getPayload();
 }
 ```
 
--- a/docs/features/async.md
+++ b/docs/features/async.md
@@ -158,9 +158,9 @@ The function has a `?float $timeout = null` argument. If the timeout is set to
 is received, the function will yield the response. If the timeout is reached, or
 all responses are resolved, the loop stops.
 
-The function has also a `$fullResponse` arguments. When false (default value)
-the HTTP client will wait for only the first bytes. Otherwise, the HTTP client
-will wait until receiving the entire response.
+The function has also a `$fullResponse` argument. When `false` (default value)
+the HTTP client will wait only for HTTP status code and headers. When `true`,
+the HTTP client will wait until receiving the full response body.
 
 ```php
 use AsyncAws\Core\Result;
--- a/docs/features/async.md
+++ b/docs/features/async.md
@@ -165,7 +165,7 @@ for ($i = 0; $i < 10; ++$i) {
     ]));
 }
 
-foreach (Result::resolveAll($results, null) as $result => $resolved) {
+foreach (Result::multiplex($results) as $result) {
     echo $result->getPayload();
 }
 ```
--- a/docs/features/async.md
+++ b/docs/features/async.md
@@ -150,6 +150,18 @@ Consider the following example. It is creating 10 `InvocationRequest`s and print
 their result. The result that is downloaded first will be printed first. The order
 the requests are created do not matter.
 
+The `Result::multiplex` function will iterate over provided results, and yield
+the the response as soon as it has been resolved.
+
+The function has a `?float $timeout = null` argument. If the timeout is set to
+`2.0`, the HTTP client will wait for 2 seconds for responses. Each time a response
+is received, the function will yield the response. If the timeout is reached, or
+all responses are resolved, the loop stops.
+
+The function has also a `$fullResponse` arguments. When false (default value)
+the HTTP client will wait for only the first bytes. Otherwise, the HTTP client
+will wait until receiving the entire response.
+
 ```php
 use AsyncAws\Core\Result;
 use AsyncAws\Lambda\LambdaClient;
@@ -165,7 +177,7 @@ for ($i = 0; $i < 10; ++$i) {
     ]));
 }
 
-foreach (Result::multiplex($results) as $result) {
+foreach (Result::multiplex($results, null, true) as $result) {
     echo $result->getPayload();
 }
 ```
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -36,9 +36,9 @@ class Response
 
     /**
      * A Result can be resolved many times. This variable contains the last resolve result.
-     * Null means that the result has never been resolved. Array contains material to create an exception
+     * Null means that the result has never been resolved. Array contains material to create an exception.
      *
-     * @var bool|array|null
+     * @var bool|HttpException|NetworkException|array|null
      */
     private $resolveResult;
 
@@ -57,7 +57,7 @@ class Response
     private $streamStarted = false;
 
     /**
-     * A flag that indicated that an exception has been thrown to the user
+     * A flag that indicated that an exception has been thrown to the user.
      */
     private $didThrow = false;
 
@@ -130,7 +130,7 @@ class Response
         $httpResponses = [];
         $httpClient = null;
         foreach ($responses as $index => $response) {
-            if (null !== $response->resolveResult && ($response->resolveResult !== true || !$downloadBody || $response->bodyDownloaded)) {
+            if (null !== $response->resolveResult && (true !== $response->resolveResult || !$downloadBody || $response->bodyDownloaded)) {
                 yield $index => $response;
 
                 continue;
@@ -346,10 +346,12 @@ class Response
 
         if (\is_array($this->resolveResult)) {
             [$class, $args] = $this->resolveResult;
+            /** @psalm-suppress PropertyTypeCoercion */
             $this->resolveResult = new $class(...$args);
         }
         if ($this->resolveResult instanceof Exception) {
             $this->didThrow = true;
+
             throw $this->resolveResult;
         }
 
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -10,6 +10,7 @@ use AsyncAws\Core\Exception\Http\HttpException;
 use AsyncAws\Core\Exception\Http\NetworkException;
 use AsyncAws\Core\Exception\Http\RedirectionException;
 use AsyncAws\Core\Exception\Http\ServerException;
+use AsyncAws\Core\Exception\InvalidArgument;
 use AsyncAws\Core\Exception\RuntimeException;
 use AsyncAws\Core\Stream\ResponseBodyResourceStream;
 use AsyncAws\Core\Stream\ResponseBodyStream;
@@ -125,9 +126,13 @@ class Response
             $responseMap[\spl_object_id($response->httpResponse)] = $response;
         }
 
-        // response is empty
-        if (null === $httpClient) {
+        // no reponse provided (or all responses already resolved)
+        if (empty($httpResponses)) {
             return;
+        }
+
+        if (null === $httpClient) {
+            throw new InvalidArgument('At least one response should have contains an Http Client');
         }
 
         try {
@@ -138,6 +143,7 @@ class Response
                 if ($chunk->isFirst()) {
                     $response = $responseMap[\spl_object_id($httpResponse)] ?? null;
 
+                    // Check if null, just in case symfony yield several time a `first` chunk for the same response
                     if (null !== $response) {
                         try {
                             $response->handleStatus();
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -15,6 +15,7 @@ use AsyncAws\Core\Exception\RuntimeException;
 use AsyncAws\Core\Stream\ResponseBodyResourceStream;
 use AsyncAws\Core\Stream\ResponseBodyStream;
 use AsyncAws\Core\Stream\ResultStream;
+use Symfony\Component\HttpClient\Exception\TransportException;
 use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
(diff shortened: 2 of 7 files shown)


src/Core/src/Response.php
```diff
@@ -85,24 +86,83 @@ public function resolve(?float $timeout = null): bool
...
+        /** @var self[] $responseMap */
+        $responseMap = [];
+        $httpResponses = [];
+        $httpClient = null;
+        foreach ($responses as $response) {
+            if (null !== $response->resolveResult) {
+                if (\is_bool($response->resolveResult)) {
+                    yield $response => $response->resolveResult;
+                } else {
+                    yield $response => true;
```
jderusse: I'm not happy with this. it hides exceptions and user may think (wrongly) the response is successfully proceeded.

I can't return false neither, user may think that response is not yet proceeded and call this method again and again...

returning the exception is not really usable.

I think I'll rename this method into `multiplex`, yielding only the response. and let user calling `resolve` (or fetching a getter) on each item

jderusse: fixed

src/Core/src/Result.php
```diff
@@ -62,6 +62,32 @@ public function __destruct()
...
+     * @param self[]     $results
+     * @param float|null $timeout Duration in seconds before aborting. When null wait
+     *                            until the end of execution. Using 0 means non-blocking
+     *
+     * @return iterable<self>
+     *
+     * @throws NetworkException
+     * @throws HttpException
+     */
+    final public static function multiplex(iterable $results, float $timeout = null): iterable
```
Nyholm: You suggested `resolveAll()` in a comment before. I like that better.

src/Core/src/Response.php
```diff
@@ -85,24 +86,74 @@ public function resolve(?float $timeout = null): bool
...
+            }
+            $httpResponses[] = $response->httpResponse;
+            $responseMap[\spl_object_id($response->httpResponse)] = $response;
         }
 
-        if (300 <= $statusCode) {
-            throw $this->resolveResult = new RedirectionException($this->httpResponse);
+        // response is empty
+        if (null === $httpClient) {
+            return;
```
Nyholm: Throw an error here instead. 

And at the very top of the function do a check `if (empty($responses)) { return;}`

src/Core/src/Response.php
```diff
@@ -85,24 +86,74 @@ public function resolve(?float $timeout = null): bool
...
-        return $this->resolveResult = true;
+        try {
+            foreach ($httpClient->stream($httpResponses, $timeout) as $httpResponse => $chunk) {
+                if ($chunk->isTimeout()) {
+                    continue;
+                }
+                if ($chunk->isFirst()) {
+                    $response = $responseMap[\spl_object_id($httpResponse)] ?? null;
+
+                    if (null !== $response) {
```
Nyholm: What if it is null? Then the SymfonyHttpClient has made a request but it does not belong to the `$responses`, so we should ignore it? Is that correct?

I would really like a comment about this.

src/Core/src/Response.php
```diff
@@ -85,24 +86,74 @@ public function resolve(?float $timeout = null): bool
...
+                        try {
+                            $response->handleStatus();
+                        } catch (Exception $e) {
+                        }
+                        yield $response;
```
Nyholm: `$response->resolveResult instanceof NetworkException|HttpException`, right?

src/Core/src/Response.php
```diff
@@ -85,24 +86,74 @@ public function resolve(?float $timeout = null): bool
...
+                    }
+                    unset($responseMap[\spl_object_id($httpResponse)]);
+                    if (empty($responseMap)) {
+                        // early exit if all statusCode are known. We don't have to wait for all responses bodies
+                        return;
+                    }
+                }
+            }
+        } catch (TransportExceptionInterface $e) {
+            throw new NetworkException('Could not contact remote server.', 0, $e);
```
Nyholm: Hm.. When will this be thrown?

src/Core/src/Response.php
```diff
@@ -85,24 +86,74 @@ public function resolve(?float $timeout = null): bool
...
+     * @param self[]     $responses
+     * @param float|null $timeout   Duration in seconds before aborting. When null wait
+     *                              until the end of execution. Using 0 means non-blocking
+     *
+     * @return iterable<self>
+     *
+     * @throws NetworkException
+     * @throws HttpException
+     */
+    final public static function multiplex(iterable $responses, float $timeout = null): iterable
```
Nyholm: I think it makes sense to add `$fullResponse` or similar to this method. But not the "resolve" function. Hm.. or am I wrong?

src/Core/src/Response.php
```diff
@@ -85,24 +86,74 @@ public function resolve(?float $timeout = null): bool
...
+        // response is empty
+        if (null === $httpClient) {
+            return;
         }
 
-        return $this->resolveResult = true;
+        try {
+            foreach ($httpClient->stream($httpResponses, $timeout) as $httpResponse => $chunk) {
+                if ($chunk->isTimeout()) {
+                    continue;
```
Nyholm: This means all http responses are inactive. We should return here, not `continue`.

src/Core/src/Response.php
Nyholm: This catch should be inside the loop. 

One request may throw a TransportException but that should not effect all of them, right?

Nyholm:
Question: 

Does it make sense to add a `body_downloaded` to the `$result->info()`?

We know the header are downloaded when the result is resolved or then `$chunk->isFirst()`. The response is downloaded when: 
- If no stream, we called `populateResult()`
- If stream, when the stream is read...

-----------------

hm...

src/Core/src/Result.php
jderusse: issue with `resolveAll`, is: it sounds like `resolve`, people may expect the same behavior:
- when yielding true => request resolve in given timeout
- when yielding false => request not resolved in given timeout
- What if exception occures?.. We can't yield an exception and continue processing other reponse.

any idea to solve this?

src/Core/src/Response.php
jderusse: thanks

Nyholm:
I continue to write down my thoughts..

So the problem we want to solve: 
Prove a way to allow curl to download multiple requests at the same time and return the results in the order they are downloaded. 
- We will not return any files from the Symfony namespace. We
- We should use our own request objects

There are two scenarios. 
1) When I do care about the payload
2) When I dont care about the payload. 

If I download 5 responses. They are of size 1kb, 1kb, 10MB, 2MB, 2MB. The large response takes 5 seconds to download. If I call `resolveAll($responses, 2.0)`. Then some responses will be fully downloaded and some will not be. I need a way to see which one that are completely downloaded.

src/Core/src/Response.php
jderusse: Just defensive programming. It should not happens, unless symfony http client yield several times a "first chunk"

Comment added

src/Core/src/Response.php
jderusse: the purpose of this `multiplex` method is yielding each time a response is `resolved` (success or failure)

Do you suggest: `yield $response => $isSuccessfull`:
```php
foreach (Result::multiplex($results) as $result => $success) {
  if (!$success) {
    // deal with error
  }
}
```

could also be `yield $response => $exceptionOrNull`

WDYT?

src/Core/src/Result.php
Nyholm: Hm.. lets make sure we dont throw anything... Ie, just name this `download()`. 

```php
while(!empty/$results) {
  Result::download($results, 2.0);
    if ($result->info()['resolved']) {
      unset($results[$i];
      $result->getPayload(); // This may throw exception. 
    }
  }
}

```

src/Core/src/Response.php
jderusse: it throw by iterrating on `$httpClient->stream()`

> One request may throw a TransportException

yes you're right. Should I play with `next`/`current` to manually move the cursor?

src/Core/src/Response.php
Nyholm: I like why it is there. I just want a comment to explain that. 

This is fairly complicated for me and I spent plenty of time on this library on Symfony HTTP client. Im just trying to be friendly to the next person reviewing this. (or nice to myself 6 months from now) =)

src/Core/src/Response.php
jderusse: good idea! I like it!
it avoid adding another parameter to `resolve` and focus multiplexing on that method only.

src/Core/src/Response.php
Nyholm: No, sorry. 
I just want to confirm the state of the $response. I use my words: 

Is it true, that on this line. The `$response->resolveResult` will always be an instance of `NetworkException` or `HttpException`?

src/Core/src/Response.php
jderusse: Sure abut that? Can't we have a bunch of responses in the buffer, were the first is inactive and other unfetched buffer?

nb: I tested with `0` and I didn"t get an infinite loop, if you afraid about that

Nyholm: From the docs: 

> If all responses become inactive for the given duration, the method will yield a special chunk whose isTimeout() will return true:

Nyholm:
I kind of like this: https://github.com/async-aws/aws/pull/417#discussion_r401224694

Nyholm:
This is growing on me. 
Im more and more convinced that we: 

1. Need a parameter like $downloadFullResponse
2. Should not throw any exceptions
3. Should extend `$result->info()` with a new key to say if the full response is downloaded. It should have type `?bool`. 

I currently dont see any issues.

Nyholm:
But oh no.. The example here (https://github.com/async-aws/aws/pull/417#discussion_r401224694) is blocking for 2 seconds... We want it to yield a response. 

```php
$downloadFullResponse = false;

while(!empty($results) {
  foreach (Result::download($results, 2.0, $downloadFullResponse) as $i => $result) {
    echo $i;
    unset($results[$i]);
    if ($result->getStatus() >= 300) {
      // handle error
    }

   // This call will be blocking as it may download the 4Gb response
   $result->getPayload()
  }
}
```

As `$i` is not in order, but the same order and values as `$results`, the code above may print: 

```
3
2
foobar
12
```

If we do download response. 
```php
$downloadFullResponse = true;

while(!empty($results) {
  foreach (Result::download($results, 2.0, $downloadFullResponse) as $i => $result) {
    echo $i;
    unset($results[$i]);
    if ($result->getStatus() >= 300) {
      // handle error
    }

   // This call is not blocking
   $result->getPayload()
  }
}
```

Nyholm:
Last example of using `$result->info()`. This is rare, but I may want to do this if I expect very different sizes of responses. 

```php
$results = [ /* 10 results */];
$count = 0;
while(!empty($results) {
  $count++;
  foreach (Result::download($results, 2.0, false) as $i => $result) {
    if ($result->getStatus() >= 300) {
      // Handle errors early. 
    }
   
    if ($result->getInfo()['body_downloaded']) {
       echo $result->getPayload();
       unset($results[$i]);
    }
  }
}

echo $count; // 1230
```

I burn the CPU on curl until I get all the respons' first chunk. Then I burn CPU on PHP when I wait on the body for all requests.

src/Core/src/Result.php
Nyholm: This example is no good. See https://github.com/async-aws/aws/pull/417#issuecomment-606894093

src/Core/src/Response.php
jderusse: oh yes, you're right

jderusse: euhh no... it's either `true|NetworkException|HttpException`

Nyholm: You are correct. 
It cannot be null.

jderusse:
I think we agreed on https://github.com/async-aws/aws/pull/417#issuecomment-606894093
This is almost what I've implemented.
Almost:
- yield **index** => response
- provides a fullBodyDownloaded in info
- rename the method (I'm not convinced by `multiplexe` neither by `download` :/)

src/Core/src/Response.php
jderusse: hmm, we can't do it.
nb: in fact nothing can throw an exception inside the loop (we just manipulat DTO objects)

The exception can be thrown between 2 yield of HttpClient: when moving the cursor from a response to another (internally calling `iterator::next`).
If such exception occurred we can not ignore this failed response and witch to another, given the exception is thrown while moving the cursor.

jderusse: Just figured out that `$chunk->timeout()` (and other methods of chunk) will throw the exception.
It's possible to continue the loop. I've fixed it in the last commit

src/Core/src/Response.php
jderusse: fixed,thank for the pointer

jderusse:
I think we are good here :)

Nyholm:
But how about the name?
I like download more that multiplex. Other suggestions may be:

- wait (sounds like a promise)
- stream (confusing with other methods)
- resolveAll (Same as stream)
- fetch
- load

src/Core/src/Response.php
```diff
@@ -140,14 +245,20 @@ public function getContent(): string
     {
         $this->resolve();
 
-        return $this->httpResponse->getContent(false);
+        $content = $this->httpResponse->getContent(false);
+        $this->bodyDownloaded = true;
+
+        return $content;
```
Nyholm, line 251: This is a trick @nicolas-grekas has showed me =)

```suggestion
        try {
            return $this->httpResponse->getContent(false);
        } finally {
            $this->bodyDownloaded = true;
        }
```

src/Core/src/Response.php
```diff
@@ -140,14 +245,20 @@ public function getContent(): string
...
 
     public function toArray(): array
     {
         $this->resolve();
 
-        return $this->httpResponse->toArray(false);
+        $content = $this->httpResponse->toArray(false);
+        $this->bodyDownloaded = true;
+
+        return $content;
```
Nyholm, line 261:
```suggestion
        try {
            return $this->httpResponse->toArray(false);
        } finally {
            $this->bodyDownloaded = true;
        }
```

src/Core/src/Response.php
```diff
@@ -157,6 +268,33 @@ public function getStatusCode(): int
 
     public function toStream(): ResultStream
     {
+        if ($this->streamStarted) {
+            throw new RuntimeException('Can not create a ResultStream because the body started being downloaded. Do not call Result::multiplex');
+        }
+
```
Nyholm, line 274: We should add `$this->bodyDownloaded = true;`, right? 

We don't know for sure. But we have started streaming the response... or how do we handle this?

src/Core/src/Result.php
```diff
@@ -62,13 +62,41 @@ public function __destruct()
...
+     * @throws NetworkException
+     * @throws HttpException
+     */
+    final public static function multiplex(iterable $results, float $timeout = null, bool $downloadBody = false): iterable
+    {
+        $resultMap = [];
+        $responses = [];
+        foreach ($results as $index => $result) {
+            $responses[$index] = $result->response;
+            $resultMap[\spl_object_id($result->response)] = $result;
```
Nyholm, line 84: We dont need `\spl_object_id($result->response)` here. We can use use `$index`.

src/Core/tests/Unit/ResultTest.php
```diff
@@ -123,6 +122,19 @@ public function testMultiplexWithException()
         $result1->resolve();
     }
 
+    public function testMultiplexWithExceptionOnDestruct()
```
Nyholm, line 125: Excellent. I wound never have thought about this.

docs/features/async.md
```diff
@@ -164,13 +177,8 @@ for ($i = 0; $i < 10; ++$i) {
...
 }
 
-while (!empty($results)) {
-    foreach ($results as $i => $result) {
-        if ($result->resolve(0.01)) {
-            echo $result->getPayload();
-            unset($results[$i]);
-        }
-    }
+foreach (Result::multiplex($results, null, true) as $result) {
```
gmorel, line 180: 👍

src/Core/src/Response.php
```diff
@@ -85,39 +111,130 @@ public function resolve(?float $timeout = null): bool
...
+                    unset($indexMap[$hash]);
+                    yield $index => $response;
+                }
+            }
+            if ($chunk->isFirst()) {
+                try {
+                    // call handleStatus to set internal state: `resolveResult` and ignore errors
+                    $response->handleStatus();
+                } catch (Exception $e) {
+                    // unset resolveResult because of refcount: the exception contains a reference to the `$response`
```
jderusse, line 205: this one was very hard to find!!! see https://3v4l.org/GI9rU

When an object contains an exception created by a static method, it creates a circular reference that prevent the __destructor being called

src/Core/src/Result.php
```diff
@@ -62,13 +62,41 @@ public function __destruct()
         return $this->response->resolve($timeout);
     }
 
+    /**
+     * Make sure all provided requests are executed.
```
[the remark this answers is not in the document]
jderusse: Added a comment, and an exception to check it

src/Core/src/Response.php
Nyholm: Oh. wow. I had no idea. 

How did you find it?

src/Core/src/Response.php
```diff
@@ -85,39 +102,127 @@ public function resolve(?float $timeout = null): bool
...
+        foreach ($httpClient->stream($httpResponses, $timeout) as $httpResponse => $chunk) {
+            $hash = \spl_object_id($httpResponse);
+
+            try {
+                if ($chunk->isTimeout()) {
+                    // Receiving a timeout mean all responses are inactive.
+                    break;
+                }
+            } catch (TransportException $e) {
+                unset($indexMap[$hash]);
```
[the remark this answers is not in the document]
jderusse: good catch fixed.

src/Core/src/Response.php
stof: not equivalent in case an error happens, though.

jderusse: ahahah.. I love that syntax, but thought you didn't like it.. 
happy to apply!

src/Core/src/Response.php
jderusse: we don't know if the body is downloaded...

But you're right, better saying that we're done on our side

src/Core/src/Result.php
jderusse: good catch!!

src/Core/src/Response.php
jderusse: Added a test to assert that exception is triggered on destruct => test failed

Spent 4 hours with php meminfo, dump, to understand what happens.

Note: I refactored this code. Instead of unset, I store the "components" of the exception in an array and instanciate it when calling "resolve"

jderusse:
remaining naming... 
I'm not conviced by multiplex, neither download/

I think `resolveAll` is the best options. Now exceptions are thrown on destruct, it's close the the behavior of `resolve`

Nyholm:
I like `download` over `resolveAll` because this code makes sense:

```php
foreach (Result::download($results) as $result) {
    try {
         $result->resolve();
    } catch (AsyncAwsException $e) {
        // Error handling
    }
    echo $result->payload();
}

```

But to name it `resolveAll()` then in the loop call `resolve()`feels weird. 

Anyhow. Rename it to `resolveAll` and I can submit a new PR with a new name if I start to feel strongly for anything else.

Nyholm:
Thank you

jderusse:
thanks for the review and your time

PR #428, opened 2020-04-02
Try to mitigate random crashes in the tests

Nyholm:
I noticed that this test was failing, so I restarted it and was green...

This PR does not fix the issue, but it reduces the change of getting a failure.

### DIFF
--- a/src/Service/Lambda/tests/Integration/LambdaClientTest.php
+++ b/src/Service/Lambda/tests/Integration/LambdaClientTest.php
@@ -67,11 +67,11 @@ class LambdaClientTest extends TestCase
 
         $results = [];
         $expected = [];
-        for ($i = 0; $i < 5; ++$i) {
+        for ($i = 0; $i < 7; ++$i) {
             $expected[] = '"hello ' . $i . '"';
             $results[] = $client->Invoke(new InvocationRequest([
                 'FunctionName' => 'Index',
-                'Payload' => \json_encode(['name' => $i, 'delay' => 10000*($i % 3)]),
+                'Payload' => \json_encode(['name' => $i, 'delay' => 80 * ($i % 3)]),
             ]));
         }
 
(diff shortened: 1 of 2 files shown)


src/Service/Lambda/tests/Integration/LambdaClientTest.php
```diff
@@ -67,7 +67,7 @@ public function testInvokeWait(): void
 
         $results = [];
         $expected = [];
-        for ($i = 0; $i < 10; ++$i) {
+        for ($i = 0; $i < 20; ++$i) {
```
jderusse, line 70: If it occurs again, we can add a sleep in the JS methods
```php
'FunctionName' => 'Index',
'Payload' => \json_encode(['name' => $i, 'sleep' => $i%5]),
```

Nyholm:
I've updated the test. The JS part is working

Im not sure the docker (or our script) is working.... The PR contains a lot of debug changes. I run the test and it takes 40s to get the following output: 

```
Array
(
    [0] => "hello 2. Sleep 20000"
    [1] => "hello 1. Sleep 10000"
    [2] => "hello 0. Sleep 1"
    [3] => "hello 3. Sleep 1"
    [4] => "hello 4. Sleep 10000"
)
```

Which makes me believe that docker executes these in sequence and not parallel. 
Im not sure why order is not consistent. I believe it is because of chance..

Nyholm:
It is docker that fails us... 

It takes 20 seconds to execute on AWS. 
```
Array
(
    [0] => "hello 3. Sleep 1"
    [1] => "hello 0. Sleep 1"
    [2] => "hello 1. Sleep 10000"
    [3] => "hello 4. Sleep 10000"
    [4] => "hello 2. Sleep 20000"
)
```

Nyholm:
I removed the debug code.

Im not sure what we should do here. Docker is failing us again. Should we create a new AWS account and run some integration tests (like this one) on pushes to master?

jderusse:
let's see with 20 iterations. 
if still failing, maybe using another service  (Sqs, S3)

Nyholm:
PR is updated.

PR #435, opened 2020-04-02
Improve pagination documentation

### LINKED ISSUE #432: [Documentation] Update Pagination example
`S3Client::listObjectsV2` is a bit special since it returns two types of objects. We can make this example simpler by using a more normal result class.

jderusse:
fix #432

### DIFF
--- a/docs/features/pagination.md
+++ b/docs/features/pagination.md
@@ -4,34 +4,63 @@ category: features
 
 # Pagination
 
-Some API Results are lists of items, like the result of `S3Client::listObjectsV2()`.
+Some API Results are lists of items, like the result of `CloudFormationClient::DescribeStacks()`.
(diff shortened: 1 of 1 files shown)


docs/features/pagination.md
```diff
@@ -4,34 +4,63 @@ category: features
...
 boolean `true` as first argument.
 
 ```diff
--foreach($result as $file) {
-+foreach($result->getContents(true) as $file) {
+-foreach($result->getStacks() as $stack) {
++foreach($result->getStacks(true) as $stack) {
+```
+
+Some Endpoints return several lists of items in the same response, like the
```
Nyholm, line 41:
```suggestion
Some endpoints return several lists of items in the same response, like the
```

PR #437, opened 2020-04-03
Adding CloudWatch::PutMetricAlarm

Nyholm:
Im planning to build a small tool that creates alarms automatically for my applications. I think this is the only action I need. 

This PR is in WIP because I need to write the tests. This contains only generated code. 

Fun fact: The api specification seams to be both "monitoring" and "logs". I guess there are two different APIs for CloudWatch..


Nyholm:
I changed my mind. I dont need this.

PR #438, opened 2020-04-03
Remove useless brackets for consistency with map keys

jderusse:


website/Twig/AsyncAwsTwigExtension.php
```diff
@@ -62,6 +62,16 @@ public function parseAssetUrls(string $content)
         return $content;
     }
 
+    public function renderStyle(string $name)
```
Nyholm, line 65: This was fixed in master earlier today.

Nyholm:
:+1:

website/Twig/AsyncAwsTwigExtension.php
jderusse: thanks, rebased

PR #447, opened 2020-04-04
Hide copy button mobile devices

jderusse:


website/assets/style/code.scss
```diff
@@ -64,3 +64,9 @@ code {
         background-image: url(./res/copy-error.svg);
     }
 }
+
+@media (max-width: 500px) {
```
Nyholm, line 68: Lets use our predefined breakpoints. 

```suggestion
@include media-breakpoint-down(md){
```

PR #448, opened 2020-04-04
Adding header links

Nyholm:
This will make it easier to copy a link to a header.

I did also refactor the post-processors so we easily can add more.

Screenshot 2020-04-04 at 11 54 26


Nyholm:
I thought about that. It could, Byt symfony.com has decided not to. Im not sure why. But I assume it is because a11y.

jderusse:
maybe @stof now why?

I found it very useful, but visually not that pretty :/

not: on Symfony the don't color had a very low contrast

stof:
I guess you will need to ask @javiereguiluz to know why this was done. The symfony.com website is managed by Symfony SAS, and I'm not part of it.

website/post-process/header-links.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function(dom) {
+    dom.window.document.querySelectorAll('h2, h3').forEach((headline) => {
+
+        let id = headline.getAttribute('id');
+        let link = dom.window.document.createElement('a');
+        link.setAttribute('href', '#' + id);
+        link.setAttribute('title', 'Permalink to this headline');
+        link.setAttribute('class', 'headerlink');
+        link.innerHTML = '¶';
+        headline.innerHTML = headline.innerHTML + link.outerHTML;
```
stof, line 10: why not `headline.appendChild(link)` ?

website/post-process/walk.js
```diff
@@ -0,0 +1,31 @@
+const fs = require('fs');
+const path = require('path');
+
+/**
+ * Read directory for files.
+ * @see https://stackoverflow.com/a/5827895/1526789
+ */
+let walk = function(dir, done) {
```
stof, line 8: should be a `const` as you never override it.

Nyholm: You are correct. Both these was fixed in https://github.com/async-aws/aws/pull/461

javiereguiluz:
At symfony.com we originally displayed them (although very faded, to not disturb reading) because otherwise it was difficult to discover them. But then we kept them because of accessibility, as Tobias said. I can't find the exact discussion related about accessibility, but I remember being related to that.

But maybe it's time to revisit this decision and hide them by default?

PR #455, opened 2020-04-04
Minor code cleanup

Nyholm:


src/Core/src/Result.php
```diff
@@ -24,7 +25,7 @@ class Result
      */
     protected $input;
 
-    protected $initialized = false;
+    private $initialized = false;
```
Nyholm, line 28: This is currently only used in the `Result` class.

src/Core/src/Waiter.php
```diff
@@ -48,13 +48,6 @@ class Waiter
      */
     private $finalState;
 
-    /**
-     * A Result can be resolved many times. This variable contains the last resolve result.
-     *
-     * @var bool|NetworkException|null
-     */
-    private $resolveResult;
```
Nyholm, line 56: This was never used. I guess we missed removing it when introducing the `Response` class

jderusse:
thanks

Nyholm:
Thank you for merging

PR #456, opened 2020-04-04
Make contribution icon absolute instead of fixed

Nyholm:
There was an issue with the contribution icon and code examples. I fixed it by remove `position:fixed`. 

I also make sure the HTML for this is at the bottom of the page. That will make our content to be read first with screen readers. 

Screenshot 2020-04-04 at 15 47 56


jderusse:
yeah, I wasn't sure fixed was a good idea.

Nyholm:
Thank you for merging

PR #457, opened 2020-04-04
Add missing dependencies

jderusse:


src/Integration/Symfony/Bundle/composer.json
```diff
@@ -37,6 +37,5 @@
         "psr-4": {
             "AsyncAws\\Symfony\\Bundle\\Tests\\": "tests/"
         }
-    },
-    "minimum-stability": "dev"
```
Nyholm, line 41: If we keep this line, we dont need to make another PR to remove `@dev` before releasing a new version.

jderusse: it was for consistency with FlySystem integration. don't we have to create PR to add `^0.n` for each minor version (before 1.0)?

Nyholm: This line was removed from Flysystem because it was downloading dev-master from flysystem v2. That broke CI time to time. 

> don't we have to create PR to add ^0.n for each minor version (before 1.0)?

We do. But I dont want to make a PR to remove `@dev` before a new release.

jderusse: reverted

PR #462, opened 2020-04-04
Generate copy button html at build time

jderusse:


website/post-process/copy-code.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function (dom) {
+    for (const elem of dom.window.document.querySelectorAll('code.hljs:not(.language-text,.language-diff)')) {
```
Nyholm, line 2: Good to filter these

website/assets/script/code.js
```diff
@@ -1,25 +1,21 @@
...
-            console.error(e);
-            button.classList.add('btn-copy-error')
-            setTimeout(() => {
-                button.classList.remove('btn-copy-error')
-            }, 3000);
-        }
-    });
-    elem.prepend(button);
-}
+window.onCopyClick = (button) => {
```
Nyholm, line 1: We would never to a global function in PHP. 

Im not sure what is best here. Either we are happy with the global function, or we register a bunch of event listeners (which takes time) and that we may never need..

Hm.. Im fine with a global function.

jderusse: you're right. I'm fixing it

PR #466, opened 2020-04-04
Add supports for multi-region

### LINKED ISSUE #464: Add support for multi-region
The official AWS SDK provide a "MultiRegionClient", that enables users to specify which AWS Region to send a command to by providing an `@region` input parameter

https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/s3-multiregion-client.html

We probably don't need to provide 2 Clients classes, but we can check for `@region` parameter in request.

jderusse:
fixes #464

### DIFF
--- a/src/Core/src/Sts/Input/GetCallerIdentityRequest.php
+++ b/src/Core/src/Sts/Input/GetCallerIdentityRequest.php
@@ -8,6 +8,11 @@ use AsyncAws\Core\Stream\StreamFactory;
 
 final class GetCallerIdentityRequest extends Input
 {
+    /**
+     * @param array{
+     *   @region?: string,
+     * } $input
+     */
     public function __construct(array $input = [])
     {
         parent::__construct($input);
--- a/src/Core/src/Sts/Input/GetCallerIdentityRequest.php
+++ b/src/Core/src/Sts/Input/GetCallerIdentityRequest.php
@@ -6,11 +6,16 @@ use AsyncAws\Core\Input;
 use AsyncAws\Core\Request;
 use AsyncAws\Core\Stream\StreamFactory;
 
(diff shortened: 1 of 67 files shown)


src/Core/src/Sts/Input/GetCallerIdentityRequest.php
```diff
@@ -6,11 +6,16 @@
 use AsyncAws\Core\Request;
 use AsyncAws\Core\Stream\StreamFactory;
 
-class GetCallerIdentityRequest implements Input
+class GetCallerIdentityRequest extends Input
 {
+    public function __construct(array $input = [])
```
Nyholm, line 11: There is no doc block here. Do we care?

src/Core/src/Sts/StsClient.php
```diff
@@ -77,7 +81,8 @@ public function assumeRoleWithWebIdentity($input): AssumeRoleWithWebIdentityResp
      */
     public function getCallerIdentity($input = []): GetCallerIdentityResponse
     {
-        $response = $this->getResponse(GetCallerIdentityRequest::create($input)->request(), new RequestContext(['operation' => 'GetCallerIdentity']));
+        $input = GetCallerIdentityRequest::create($input);
```
Nyholm, line 84: Same here, The doc block for $input does not specify `@region`.

src/Core/src/Sts/Input/GetCallerIdentityRequest.php
jderusse: good catch, I've to fix it

jderusse: fixed

PR #469, opened 2020-04-05
Prevent conflict with region

jderusse:


src/CodeGenerator/src/Generator/InputGenerator.php
```diff
@@ -97,6 +97,9 @@ public function generate(Operation $operation): ClassName
         $constructorBody = '';
 
         foreach ($shape->getMembers() as $member) {
+            if ('region' === $member->getName()) {
```
Nyholm, line 100: Is it `@region` or `region`?

jderusse:
the MultiRegion hack use `@region`. But here I want to prevent Input that overlap `setRegion` and `getRegion`

src/CodeGenerator/src/Generator/InputGenerator.php
jderusse: the MultiRegion hack use @region. But here I want to prevent Input that overlap setRegion and getRegion

Nyholm:
Thank you

PR #474, opened 2020-04-09
Update http-client-contracts to 1.1.8.

Nyholm:
It was at that version it was documented the httP-code would return int.

Since we have `"symfony/http-client": "^4.4"`, there is impossible to get a version lower that 1.1.8. This is just a "to be technically correct".


jderusse:
good catch! thanks

Nyholm:
Thank you for merging

PR #476, opened 2020-04-11
Remove null in annotation

Nyholm:


src/Core/src/Credentials/CacheProvider.php
```diff
@@ -17,7 +17,7 @@
 final class CacheProvider implements CredentialProvider, ResetInterface
 {
     /**
-     * @var (null|Credentials)[]
+     * @var Credentials[]
```
jderusse, line 20: it's wrong, null could be an expected value

Nyholm: Yes. I realised that now. Thank you

PR #477, opened 2020-04-11
Dont use buggy version of PHP-cs-fixer

Nyholm:
I've reported the bug in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4913

This PR will allow the CI to be green again and make sure #475 is not removing code that should not be removed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "symfony/service-contracts": "^1.0 || ^2.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^2.16",
+        "friendsofphp/php-cs-fixer": "2.16.1",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
         "symfony/service-contracts": "^1.0 || ^2.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^2.16",
+        "friendsofphp/php-cs-fixer": "2.16.1",
```
jderusse, line 19: what's about 
```suggestion
        "friendsofphp/php-cs-fixer": "^2.16,!2.16.2",
```

jderusse: not tested

Nyholm: But maybe 2.16.3 is also buggy?

jderusse:
thank you

PR #482, opened 2020-04-12
Restore is default method

jderusse:
Used here https://github.com/symfony/symfony/pull/35992/commits/82b5e7cf1aa625552d09cbffb50dfc4a767eb6dd

I don't know if it worth such public method in AsyncAws.
A workaround in Symfon would be :neutral_face: 
```diff
        $configuration = $this->sesClient->getConfiguration();
-       if (!$configuration->isDefault('endpoint')) {
+       if ($configuration->get('endpoint') === (new Configuration())->get('endpoint')) {
            $endpoint = parse_url($configuration->get('endpoint'));
            $host = $endpoint['host'].($endpoint['port'] ?? null ? ':'.$endpoint['port'] : '');
        } else {
            $host = $configuration->get('region');
        }

        return sprintf('ses+api://%s@%s', $configuration->get('accessKeyId'), $host);
```

### DIFF
--- a/src/Core/tests/Unit/ConfigurationTest.php
+++ b/src/Core/tests/Unit/ConfigurationTest.php
@@ -30,6 +30,14 @@ class ConfigurationTest extends TestCase
         }
     }
 
+    public function testIsDefault()
+    {
+        $config = Configuration::create(['region' => 'eu-west-3']);
+
+        self::assertTrue($config->isDefault('endpoint'));
+        self::assertFalse($config->isDefault('region'));
+    }
+
     public function provideConfiguration(): iterable
     {
         yield 'simple config' => [['endpoint' => 'foo'], [], ['endpoint' => 'foo']];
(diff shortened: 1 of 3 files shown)


Nyholm:
Im happy to have it. I removed it before because I couldn't see any need for it and we didnt use it internally. I was wrong =)

src/Core/tests/Unit/ConfigurationTest.php
```diff
@@ -30,6 +30,14 @@ public function testCreate($config, $env, $expected)
         }
     }
 
+    public function testIsDefault()
```
Nyholm, line 33: Good. 

With a test, this is official =)

jderusse: good catch. Dones

jderusse:
Changelog entry added.

Nyholm:
I am awk. Feel free to merge and tag

Issue #486, opened 2020-04-15
SSM parameter + env processor


Nemo64:
I'd love to be able to read ssm parameter using this library.

Actually I want to build a symfony bundle that allows me to read secret parameters using EnvProcessors like this:
```php
services:
    some_service:
        arguments: ['%env(ssm:/path/to/parameter)%']
```
This is especially interesting with secret parameters. Maybe even with the secret manager.

- I either build everything with the aws-sdk
- I could contribute here and try to add ssm parameter support (but not the entire ssm api, that's a lot)
- Or I could contribute ssm parameter support and maybe even build ssm env processors into the symfony bundle.

jderusse:
Oh I love this idea!!! definitively something to add in the bundle!

jderusse:
note: this is not necessary to crate an env processor (you'll have some issue with `/` in path).
We can use an EnvVarLoaderInterface for that.

```
services:
    some_service:
        arguments: ['%env(MY_SECRET)%']
   AsyncAws\SymfonyBundle\SsmEnvVarLoader:
       arguments:
         $path: /path/to/parameters/root
```

Nemo64:
My base idea revolved around making me able to configure this completly with environment variables. If you have `%env(resolve:DATABASE_URL)%`, I could define it in my serverless configuration as `DATABASE_URL=mysql://%env(ssm:/app/host)%/` ~ something like that, more likely a secret.

Now to keep that working i could just use the same method `%env(file:)%` is using.
`DATABASE_URL=mysql://%env(ssm:DATABASE_URL_PARAMETER)%/`
`DATABASE_URL_PARAMETER=/app/host`

Another advantage of using an env processor is that it is only executed when the corresponding service is created. Although with an env loader I could probably request them all with 1 api call ~ not too sure.

jderusse:
> something like that, more likely a secret

[Symfony Secrets](https://symfony.com/blog/new-in-symfony-4-4-encrypted-secrets-management) uses `EnvLoaderInterface` because it have several benefits over EnvProcessor:

- in dev mode or CI, you can use the .env file, and use secrets only on production
- all variables are loaded at the same time (avoid 1 call per secret which could be a performance killer when performing 1 HTTP request for each param)
- same variable is reused when used in multiple places
- variables can be dumped at build time (or startup time) and avoid fetching secrets on each request.

> I could define it in my serverless configuration as `DATABASE_URL=mysql://%env(ssm:/app/host)%/`

[...]

PR #489, opened 2020-04-18
Add Ssm Client

jderusse:
Part of #486

### DIFF
--- a/docs/contribute/generate.md
+++ b/docs/contribute/generate.md
@@ -26,7 +26,8 @@ is an example for DynamoDB.
 1. Clone the async-aws/aws repository.
 1. Run `composer install`
 1. Copy `./src/Service/.template` to `./src/Service/DynamoDb`
-1. Rename all occurrences of `Foobar` to `DynamoDb` and `foobar` to `dynamo-db`
+1. Replace all occurrences of `Foobar` to `DynamoDb` and `foobar` to `dynamo-db`
+1. Rename `FoobarClient` into `DynamoDbClient`
 1. Register the service in the `services` section of the `./manifest.json` file.
    1. Look at the official SDK for resources in `./src/data`. ([DynamoDB example](https://github.com/aws/aws-sdk-php/tree/3.133.23/src/data/dynamodb/2012-08-10))
    1. Add a link to `source`, and, if the files exists, links to `documentation`, `example`, `pagination`, `waiter` etc.
--- a/docs/contribute/generate.md
+++ b/docs/contribute/generate.md
@@ -32,8 +32,8 @@ is an example for DynamoDB.
    1. Look at the official SDK for resources in `./src/data`. ([DynamoDB example](https://github.com/aws/aws-sdk-php/tree/3.133.23/src/data/dynamodb/2012-08-10))
    1. Add a link to `source`, and, if the files exists, links to `documentation`, `example`, `pagination`, `waiter` etc.
    1. Leave the `methods` key empty. (`"methods": []`)
+1. Add entry for the client in the `AsyncAws\Symfony\Bundle\DependencyInjection\AwsPackagesProvider::getAllServices` method.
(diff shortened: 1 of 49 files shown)


src/Service/Ssm/tests/Unit/Result/GetParameterResultTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+        $client = new MockHttpClient($response);
+        $result = new GetParameterResult(new Response($client->request('POST', 'http://localhost'), $client));
+
+        self::assertSame('arn:aws:ssm:us-east-2:111122223333:parameter/MyGitHubPassword', $result->getParameter()->getARN());
+        self::assertSame('20200225 190128 800000', $result->getParameter()->getLastModifiedDate()->format('Ymd His u'));
+        self::assertSame('MyGitHubPassword', $result->getParameter()->getName());
+        self::assertSame(ParameterType::SECURE_STRING, $result->getParameter()->getType());
+        self::assertSame('AYA39c3b3042cd2aEXAMPLE/AKIAIOSFODNN7EXAMPLE/fh983hg9awEXAMPLE==', $result->getParameter()->getValue());
+        self::assertSame('3', $result->getParameter()->getVersion());
```
jderusse, line 37: In the AWS .json file this property is a "long" that's why we convert it to string.

Now, The JSON contains response (see above at line 25) is an int... I wonder, if we should deal long an int..

docs/contribute/generate.md
```diff
@@ -26,11 +26,13 @@ is an example for DynamoDB.
...
 1. Run `composer install`
 1. Copy `./src/Service/.template` to `./src/Service/DynamoDb`
-1. Rename all occurrences of `Foobar` to `DynamoDb` and `foobar` to `dynamo-db`
+1. Replace all occurrences of `Foobar` to `DynamoDb` and `foobar` to `dynamo-db`
+1. Rename `FoobarClient` into `DynamoDbClient`
 1. Register the service in the `services` section of the `./manifest.json` file.
    1. Look at the official SDK for resources in `./src/data`. ([DynamoDB example](https://github.com/aws/aws-sdk-php/tree/3.133.23/src/data/dynamodb/2012-08-10))
    1. Add a link to `source`, and, if the files exists, links to `documentation`, `example`, `pagination`, `waiter` etc.
    1. Leave the `methods` key empty. (`"methods": []`)
+1. Add entry for the client in the `AsyncAws\Symfony\Bundle\DependencyInjection\AwsPackagesProvider::getAllServices` method.
```
Nyholm, line 35: I decided not to add it here because there is a test that verifies that. There are a few more things todo, like update the ApiClientFactory and the readme. But I excluded those too.

src/Service/Ssm/tests/Unit/Result/GetParameterResultTest.php
Nyholm: I guess we must use strings since we cannot handle longs in PHP... I think this is correct.

docs/contribute/generate.md
jderusse: Indeed, I didn't ran tests before writing this line.. removed

src/Service/Ssm/composer.json
```diff
@@ -0,0 +1,33 @@
...
+        "ssm"
+    ],
+    "license": "MIT",
+    "require": {
+        "php": "^7.2.5",
+        "ext-json": "*",
+        "async-aws/core": "^0.5"
+    },
+    "require-dev": {
+        "async-aws/core": "dev-master"
```
jderusse, line 16: I'm not sure this is the right way to get #492 ..

Nyholm: No :)

It is easier to see the correct way if we had released 1.0. Of code 1.0
Was released, then the feature you added in #492 would make the next release 1.1.0. We should then update the core branch alias. 

With an updated beach alias, we should just require core ^1.1 here. It will automatically fetch the unreleased version until we actually release core 1.1.0. 

So... now when we are at 0.5.x... either we release a patch version of core now or we can use a branch alias. I think the patch version is simpler.

PR #493, opened 2020-04-21
Fix requirement on SymfonyBundle

jderusse:
Requires `symfony/framework-bundle: ^4.4` because config router=utf-8
Remove mimimum stability=dev from composer.json


Nyholm:
Hm. We can also remove that configuration. I just added it there because it was default.

Nyholm:
Thank you very much!

PR #494, opened 2020-04-21
Add a "link" tool

### LINKED ISSUE #285: How to test PRs?
If someone submit a PR with a patch for a bug, how can that be tested before the PR is merged?

jderusse:
Fixes #285

### DIFF
--- a/link
+++ b/link
@@ -35,8 +35,6 @@ $directories = array_merge(...array_values(array_map(function ($part) {
     return glob(__DIR__.'/src/'.$part.'/*', GLOB_ONLYDIR | GLOB_NOSORT);
 }, $braces)));
 
-$directories[] = __DIR__.'/src/Symfony/Contracts';
-
 foreach ($directories as $dir) {
     if ($filesystem->exists($composer = "$dir/composer.json")) {
         $packages[json_decode(file_get_contents($composer))->name] = $dir;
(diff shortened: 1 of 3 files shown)


link
```diff
@@ -0,0 +1,68 @@
...
+
+$packages = array();
+
+$filesystem = new Filesystem();
+$braces = array('.', 'Integration/*', 'Service');
+$directories = array_merge(...array_values(array_map(function ($part) {
+    return glob(__DIR__.'/src/'.$part.'/*', GLOB_ONLYDIR | GLOB_NOSORT);
+}, $braces)));
+
+$directories[] = __DIR__.'/src/Symfony/Contracts';
```
Nyholm, line 38: Is this needed?

link
```diff
@@ -0,0 +1,68 @@
...
+    if (!$copy && is_link($dir)) {
+        echo "\"$package\" is already a symlink, skipping.".PHP_EOL;
+        continue;
+    }
+
+    if (!isset($packages[$package])) {
+        continue;
+    }
+
+    $sfDir = ('\\' === DIRECTORY_SEPARATOR || $copy) ? $packages[$package] : $filesystem->makePathRelative($packages[$package], dirname(realpath($dir)));
```
Nyholm, line 57: `$sfDir`? Maybe rename it =)

link
jderusse: good catch

link
jderusse: thanks

link
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env php
+<?php
+
+require __DIR__.'/vendor/symfony/filesystem/Exception/ExceptionInterface.php';
+require __DIR__.'/vendor/symfony/filesystem/Exception/IOExceptionInterface.php';
+require __DIR__.'/vendor/symfony/filesystem/Exception/IOException.php';
+require __DIR__.'/vendor/symfony/filesystem/Filesystem.php';
```
stof, line 7: why explicit requires rather than requiring the autoloader ?

jderusse: Dummy copy/paste from Symfony's implementation

stof: The difference is that the Symfony script is usable without running `composer install`, as the require calls are in the package itself, not in `vendor`. But here, you rely on `composer install` having run already anyway (and so the autoloader will be there too).

jderusse: Ah ah.. That's so obvious.. Feel Dumb now..

Fixed in #498

PR #497, opened 2020-04-22
Remove "namespace" from SSM in manifest

Nyholm:
The "namespace" should only be used for non-standard namespaces. 

When I removed this I saw some failing tests complaining about the readme.


jderusse:
Good catch, thanks

PR #502, opened 2020-04-25
[Flysystem 1] Better check if file exists.

Nyholm:
We dont need to check if a directory exists.

### DIFF
--- a/src/Integration/Flysystem/S3/src/S3FilesystemV1.php
+++ b/src/Integration/Flysystem/S3/src/S3FilesystemV1.php
@@ -173,22 +173,12 @@ class S3FilesystemV1 extends AbstractAdapter implements CanOverwriteFiles
     {
         $location = $this->applyPathPrefix($path);
 
-        $result = $this->client->getObject(array_merge($this->options, [
+        $result = $this->client->objectExists(array_merge($this->options, [
             'Bucket' => $this->bucket,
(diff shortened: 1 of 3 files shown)


jderusse:
An other optimization: use `objectExists([''])->isSuccess()` or at least `headObject()` to avoid downloading the content (even if the stream is not consumed, I believe some bytes are downloaded and bufferized)

Nyholm:
Nice. I've updated the PR

src/Integration/Flysystem/S3/src/S3FilesystemV1.php
```diff
@@ -173,22 +173,14 @@ public function has($path)
...
 
-        $result = $this->client->getObject(array_merge($this->options, [
+        $result = $this->client->objectExists(array_merge($this->options, [
             'Bucket' => $this->bucket,
             'Key' => $location,
         ]));
 
-        try {
-            $result->resolve();
+        $result->wait();
```
jderusse, line 181: `wait` will wait until the object exists.
You should use `isSuccess` or `isFailure` to immediately get if the o ject exists or not.

Nyholm: Oops, 
Thank you

Nyholm:
Thank you for the reviews.

PR #503, opened 2020-04-25
[SF] Make sure we override config in a predictable way

Nyholm:


Nyholm:
Thank you for the review

PR #507, opened 2020-04-27
Added logging on HTTP exception

Nyholm:
My issue is that Im using Flysystem, which is using our adapter, which is using S3. 

Im 3 layers down in abstractions and I know S3 throws an exceptions that is caught in Flysystem. I want to log that exception somehow.

### DIFF
--- a/src/Core/src/Exception/Http/HttpExceptionTrait.php
+++ b/src/Core/src/Exception/Http/HttpExceptionTrait.php
@@ -39,7 +39,7 @@ trait HttpExceptionTrait
      */
     private $awsDetail;
 
-    public function __construct(ResponseInterface $response, LoggerInterface $logger)
+    public function __construct(ResponseInterface $response)
     {
         $this->response = $response;
         /** @var int $code */
@@ -68,13 +68,6 @@ trait HttpExceptionTrait
                 // Not XML ¯\_(ツ)_/¯
             }
         }
-
-        $logger->error($message, [
-            'aws_code' => $this->awsCode,
-            'aws_message' => $this->awsMessage,
-            'aws_type' => $this->awsType,
-            'aws_detail' => $this->awsDetail,
-        ]);
 
         $message .= <<<TEXT
 
--- a/src/Core/src/Exception/Http/HttpExceptionTrait.php
+++ b/src/Core/src/Exception/Http/HttpExceptionTrait.php
@@ -2,6 +2,7 @@
 
 namespace AsyncAws\Core\Exception\Http;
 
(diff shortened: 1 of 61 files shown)


Nyholm:
The first commit is more interesting to review, The other just fixed the tests and CS

src/Core/src/Exception/Http/HttpExceptionTrait.php
```diff
@@ -68,6 +69,13 @@ public function __construct(ResponseInterface $response)
             }
         }
 
+        $logger->error($message, [
```
jderusse, line 72: I'm not a big fan of loging in an exception's constructor :/

Nyholm: That is fair.. But where else can I use it?

Nyholm: Oh, In the Response class of course =)

Nyholm:
PR is updated. 

Thank you @jderusse for the review. The PR is way cleaner and better now

src/Core/src/Response.php
```diff
@@ -61,10 +62,16 @@ class Response
...
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    public function __construct(ResponseInterface $response, HttpClientInterface $httpClient, LoggerInterface $logger)
     {
         $this->httpResponse = $response;
         $this->httpClient = $httpClient;
+        $this->logger = $logger;
```
jderusse, line 74: what's about `$ths->logger = $logger ?? new NullLogger()` and remove code from Tests classes

Nyholm: I was considering that, but the only win is that test classes looks a bit cleaner. 

There is no point in doing that check on runtime.

Nyholm:
Thank you for the reviews and suggestions.

temp:
@Nyholm Is it expected to log a 404 as an error (in Response::getResolveStatus()).
I'm using async-aws with flysystem, and on each missing resource for $flysystem->has(...) an error log message is written, which results in a sentry issue here... :-D
I'd say a 404 for an existance-check is a valid answer, not an error.

Nyholm:
Good question. I’ve seen the same thing in my logs. 

Should we log it as “info” instead?

temp:
Yes, would be fine for me.

Nyholm:
Do you mind submitting a PR?

Nyholm:
Only 404 seams like the best way to move forward now.

PR #510, opened 2020-04-29
Fix SES doc example

tyx:
According to https://github.com/async-aws/aws/blob/3d4ff8366a2f278ee6a77b0b43666cef1931ae30/src/Service/Ses/src/ValueObject/Destination.php#L8


jderusse:
Good catch, thanks

PR #512, opened 2020-04-29
Use annotation to ignore phpstan rule

jderusse:
This PR replaces the phpstan config file by annotation to ignore rule.

It also refactor few part of the code to replace ignored rule php code assertion.


src/CodeGenerator/src/Command/GenerateCommand.php
```diff
@@ -61,6 +62,7 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        \assert($output instanceof ConsoleOutputInterface);
```
Nyholm, line 65: Or: 

```php
/** @var ConsoleOutputInterface $output */
```

jderusse: done

PR #516, opened 2020-05-02
Generate CodeDeploy client

Nyholm:
I'll use this to write preHooks for my lambda deployments. This is the only operation I need. 

TODO: 
- [x] Add tests
- [x] Add documentation


Nyholm:
Thank you for the review.

PR #525, opened 2020-05-03
Override theme diff color

jderusse:
This PR override the Darcula's diff color:

Before:
Screenshot from 2020-05-03 18-27-40

After:
Screenshot from 2020-05-03 18-27-30

Both colors come from the current Darcula palette (green is used for string and red for keywords)


Nyholm:
Thank you. I’ve been wanting to change this too.

PR #530, opened 2020-05-04
Fix CS

jderusse:
Fix the remaining tests once new version of PhpStan will be released (see 
https://github.com/phpstan/phpdoc-parser/issues/45#issuecomment-623697630)


src/CodeGenerator/src/Generator/PhpGenerator/ClassFactory.php
```diff
@@ -177,6 +177,7 @@ public function fromPropertyReflection(\ReflectionProperty $from): Property
             ? ClassType::VISIBILITY_PRIVATE
             : ($from->isProtected() ? ClassType::VISIBILITY_PROTECTED : ClassType::VISIBILITY_PUBLIC)
         );
+        /** @phpstan-ignore-next-line */
```
Nyholm, line 180: This line should be removed.

jderusse:
thanks

PR #532, opened 2020-05-05
Adds ECS Credential Provider 

### LINKED ISSUE #531: [WIP] Add ECS Credential Provider
Hey there, 
I'm using bref to deploy a server to lambda, everything works great but when I deploy to Codebuild using a custom ECS image it breaks since bref uses this package and this package doesnt support ECS Credentials.

~~Actually, bref package, uses async-aws/core, I don't know if is okey to submit the pr here or i should do in async-aws/core.~~

I will make a PR that adds support for this, hope that pr gets merged.

See: https://docs.aws.amazon.com/aws-sdk-php/v3/api/class-Aws.Credentials.EcsCredentialProvider.html

Greetings

nickaguilarh:
Fixes #531


Nyholm:
Thank you. 

Could you also add some documentation about this?

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->timeout = getenv(self::ENV_TIMEOUT) ?: $timeout;
+        $this->envUri = getenv(self::ENV_URI) ?: '';
+    }
+
+    public function getCredentials(Configuration $configuration): ?Credentials
+    {
+        // fetch credentials from ecs endpoint
+        try {
+            $response = $this->httpClient->request('GET', self::ENDPOINT . $this->envUri, ['timeout' => $this->timeout]);
+            $result = $this->toArray($response);
```
Nyholm, line 51: Shoudn't we check the status code before we get the response?

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -0,0 +1,108 @@
...
+        return new Credentials(
+            $result['AccessKeyId'],
+            $result['SecretAccessKey'],
+            $result['Token'],
+            new \DateTimeImmutable($result['Expiration'])
+        );
+    }
+
+    /**
+     * Copy of Symfony\Component\HttpClient\Response::toArray without assertion on Content-Type header.
```
Nyholm, line 81: Why? Doesn't the response include the proper Content-Type?

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -0,0 +1,108 @@
...
+
+    private $timeout;
+
+    private $envUri;
+
+    public function __construct(?HttpClientInterface $httpClient = null, ?LoggerInterface $logger = null, float $timeout = 1.0)
+    {
+        $this->logger = $logger ?? new NullLogger();
+        $this->httpClient = $httpClient ?? HttpClient::create();
+        $this->timeout = getenv(self::ENV_TIMEOUT) ?: $timeout;
```
jderusse, line 42: Move those getenv calls to `Configuration` class.
See how `WebIdentityProvider` works.

note: I like this `METADATA_SERVICE_TIMEOUT` parameter, that could be reused in InstanceProvider too

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public function getCredentials(Configuration $configuration): ?Credentials
+    {
+        // fetch credentials from ecs endpoint
+        try {
+            $response = $this->httpClient->request('GET', self::ENDPOINT . $this->envUri, ['timeout' => $this->timeout]);
+            $result = $this->toArray($response);
+
+            if (200 != $response->getStatusCode()) {
```
jderusse, line 53: I don't think we should handle cases were AWS returns unexpected values.

src/Core/src/Credentials/ContainerProvider.php
jderusse: I think this is a copy paste from `InstanceProvider`. 

If AWS return an invalid content type, I sugest to move this in a trait. Otherwise, you can delete it

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->httpClient = $httpClient ?? HttpClient::create();
+        $this->timeout = getenv(self::ENV_TIMEOUT) ?: $timeout;
+        $this->envUri = getenv(self::ENV_URI) ?: '';
+    }
+
+    public function getCredentials(Configuration $configuration): ?Credentials
+    {
+        // fetch credentials from ecs endpoint
+        try {
+            $response = $this->httpClient->request('GET', self::ENDPOINT . $this->envUri, ['timeout' => $this->timeout]);
```
jderusse, line 50: `$this->envUri` can be empty, the AWS SDK contains an early exit that "disable" the provider if the env variable is missing.

https://github.com/aws/aws-sdk-php/blob/af91c2fc467a326e5bcb5665e4e2ad3d84d28be2/src/Credentials/CredentialProvider.php#L90-L92

docs/authentication/ecs-container.md
```diff
@@ -0,0 +1,16 @@
...
+---
+
+# Using ECS Container Authentication Endpoint
+
+When you run code within an ECS image, AsyncAws is able to fetch Credentials from the
+[Role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html) attached to the instance using and internal aws endpoint.
+
+This type of credentials will use the environment value of `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` to generate an url that returns the temporary credentials.
+
+The current IP Address used to fetch the credentials: `169.254.170.2`
```
jderusse, line 12: I don' think this is needed. Maybe you can add a link to the documentation (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) ?

nickaguilarh:
Thanks for the review,
Yeah, it basically is a copy of Instance Provider with some modifications, i kinda did this for a setup on a production server so i had to get it working asap. Now that if fixed and working temporary so I can give a properly look at this. 
I'll make changes request and check if this return and invalid header

src/Core/src/Credentials/ContainerProvider.php
nickaguilarh: So, this is what I get when fetching a request.

> GET /v2/credentials/79ec8795-217b-4bc6-9fee-ba50fb71d87e HTTP/1.1
--
100 | > Host: 169.254.170.2
101 | > User-Agent: curl/7.67.0
102 | > Accept: */*
103 | >
104 | * Mark bundle as not supporting multiuse
105 | < HTTP/1.1 200 OK
106 | < Content-Type: application/json
107 | < X-Rate-Limit-Duration: 1
108 | < X-Rate-Limit-Limit: 40
109 | < X-Rate-Limit-Request-Forwarded-For:
110 | < X-Rate-Limit-Request-Remote-Addr: 169.254.172.2:41774
111 | < Date: Tue, 05 May 2020 18:44:56 GMT
112 | < Content-Length: 1499
113 | <

It's a valid json so I will get rid of it.

nickaguilarh:
> Thanks a lot for this contribution 
> 
> Don't forget to add an entry in the changelog

Should I add it as 1.0.x or 0.5.x?
I see in changelog 0.5.x is the last entry

Nyholm:
This is the core package. Latest version is 1.0.0. 

Add this as 1.1.0

src/Core/src/Credentials/ContainerProvider.php
jderusse: thanks for the test!! `Content-Type: application/json`is what you need.
You can delete this `toArray` method and use `$response->toArray()` instead

nickaguilarh:
So, the commit fixing the cr is made:
I also changed the way how timeout is set, now is trying to get from env and if has a default value of 1.0 as in constructors. 
Let me know if there is something else i need to do.

src/Core/src/Credentials/InstanceProvider.php
```diff
@@ -33,18 +33,18 @@ final class InstanceProvider implements CredentialProvider
 
     private $timeout;
 
-    public function __construct(?HttpClientInterface $httpClient = null, ?LoggerInterface $logger = null, float $timeout = 1.0)
```
Nyholm, line 36: We cannot remove this because it will break BC.  I suggest leaving this class unchanged and we can deal with the BC layer in the next PR.

nickaguilarh: Got it. Is it Ok if I leave the part that fetched the timeout from env but dont use it?
Also, that test is really awesome i just understood more about how is structured this project by having to debug that error :)

Nyholm: Sure. 

For the record: To the next PR that handles the BC layer we need to make the `private $timeout` to support both null and a float. We could remove the constructors third argument and then use `func_get_args()` to see if a third argument was passed to the constructor. If it was we populate our property (and trigger a deprcation), if it wasn't we leave the property to null. 

Then when we need the timeout we use our property and fallback to the environment variable.

src/Core/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+## 1.1.0
+
+### Added
+
+- Support for ECS Credentials Provider
+
+### Changed
+
+- Timeout parameter on EC2 Instance and ECS Credentials Provider now is fetched from environment variable. Otherwise, it is set to 1.0
```
Nyholm, line 11: This line is not needed now since we moved this thing to a next PR.

src/Core/src/Configuration.php
```diff
@@ -56,6 +59,10 @@ final class Configuration
             self::OPTION_WEB_IDENTITY_TOKEN_FILE => 'AWS_WEB_IDENTITY_TOKEN_FILE',
             self::OPTION_ROLE_SESSION_NAME => 'AWS_ROLE_SESSION_NAME',
         ],
+        [
+            self::OPTION_ECS_AUTH_ENDPOINT => 'AWS_CONTAINER_CREDENTIALS_RELATIVE_URI',
+            self::OPTION_METADATA_SERVICE_TIMEOUT => 'AWS_METADATA_SERVICE_TIMEOUT',
```
Nyholm, line 64: Im not sure these two should be in a group. 

@jderusse you have better understanding of this.

jderusse: Agreed with @Nyholm, a "group" means: if, at least, one option is defined by the user, the others env variables will be ignored.
ie, if user setup `accessKeyId`, and `accessKeySecret` but leave `sessionToken` empty, it will not be override by `AWS_SESSION_TOKEN`.

I think we can use `OPTION_METADATA_SERVICE_TIMEOUT` in several place,I think this variable shoud be moved  in a dedicated "group"

src/Core/src/Configuration.php
```diff
@@ -56,6 +59,10 @@ final class Configuration
             self::OPTION_WEB_IDENTITY_TOKEN_FILE => 'AWS_WEB_IDENTITY_TOKEN_FILE',
             self::OPTION_ROLE_SESSION_NAME => 'AWS_ROLE_SESSION_NAME',
         ],
+        [
+            self::OPTION_ECS_AUTH_ENDPOINT => 'AWS_CONTAINER_CREDENTIALS_RELATIVE_URI',
```
jderusse, line 63: Please use a name close to the env variable `OPTION_CONTAINER_CREDENTIALS_RELATIVE_URI`
moreover, it's not an endpoint (no scheme, nor host) but a path

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -0,0 +1,68 @@
...
+    public function __construct(?HttpClientInterface $httpClient = null, ?LoggerInterface $logger = null)
+    {
+        $this->logger = $logger ?? new NullLogger();
+        $this->httpClient = $httpClient ?? HttpClient::create();
+    }
+
+    public function getCredentials(Configuration $configuration): ?Credentials
+    {
+        $timeout = $configuration->get(Configuration::OPTION_METADATA_SERVICE_TIMEOUT);
+        $envUri = $configuration->get(Configuration::OPTION_ECS_AUTH_ENDPOINT);
```
jderusse, line 38: `$relativeUri`or `authRelativeUri`

src/Core/src/Credentials/InstanceProvider.php
jderusse: I suggest to update this class in antoher PR (I can handle it if you want)

src/Core/src/Configuration.php
nickaguilarh: Got it, I wasnt sure about this but now is clear as crystal

src/Core/src/Configuration.php
```diff
@@ -38,6 +40,7 @@ final class Configuration
         self::OPTION_ROLE_ARN => true,
         self::OPTION_WEB_IDENTITY_TOKEN_FILE => true,
         self::OPTION_ROLE_SESSION_NAME => true,
+        self::OPTION_METADATA_SERVICE_TIMEOUT => true,
```
jderusse, line 43: missing `OPTION_CONTAINER_CREDENTIALS_RELATIVE_URI`?

src/Core/src/Credentials/InstanceProvider.php
nickaguilarh: @jderusse that would be nice since i dont fully understant the BC layer

src/Core/src/Configuration.php
nickaguilarh: I wasn't sure about that since that relative uri can't be change. It's different every time an instance is run. That is as far as i know.

jderusse: I got your point, and agreed, people will propably not use it, but for consistency with other options (ie. `SESSION_TOKEN`), I prefer to let users choose the way they want to inject it.

Ie the following code works

```yaml
async_aws:
    config:
        ContainerCredentialsRelativeUri: '%env(CONTAINER_CREDENTIALS_RELATIVE_URI)%'
```

It also let the opportunity for user to share the parameter with sub-system that don't have access to the env variables (docker run, other instane, ...).

nickaguilarh: That makes sense to me. I'll add it

nickaguilarh:
Hey there
Today I was replacing the temporary fix made on our servers and I got a bug intruced on this PR, 
On line 119 in Configuration.php it defines a `?string` return value but the timeout is a float as specified on HttpClientInterface.php line 55. 
How should we addres this? 
Changing the timeout constant to string works just fine.

I just realized this since we can't complete emulate AWS ECS Instances on local machine.

Nyholm:
Hm. Do we expect users to change this timeout? What would be reason for that? Or is changing timeout something just 1% of the users will do?  If so.. maybe go back to passing timeout as a constructor argument is a fine solution?

Nyholm:
I opened https://github.com/async-aws/aws/issues/546

PR #533, opened 2020-05-05
Lock versions for phpstan and php-cs-fixer

Nyholm:
Make sure CI on master is not red because some static tool updated their rules.


Nyholm:
Blocked by #530

Nyholm:
Thank you for the reviews

PR #537, opened 2020-05-05
Added Monolog integration for CloudWatchLogs

guillaumesmo:
Please give me some feedback on this, if I'm on the right path I'll add tests and documentation

Source of the integration: https://github.com/maxbanton/cwh/blob/master/src/Handler/CloudWatch.php
I have removed some functionality that shouldn't be the responsibility of the handler in my opinion (creating the log group, setting the retention period etc)


src/Service/CloudWatchLogs/src/Result/PutLogEventsResponse.php
```diff
@@ -0,0 +1,46 @@
...
+     * The next sequence token.
+     */
+    private $nextSequenceToken;
+
+    /**
+     * The rejected events.
+     */
+    private $rejectedLogEventsInfo;
+
+    public function getnextSequenceToken(): ?string
```
guillaumesmo, line 21: Not sure if the casing is as expected on those methods, but they are autogenerated
guess this can only be fixed in the generator?

jderusse: Indeed, AWS uses lowers case to define properties for this service (while all other services are UpperCamelCase).

I think this is something we should fix in generator. /cc @Nyholm

composer.json
```diff
@@ -36,12 +36,14 @@
...
         "psr-4": {
             "AsyncAws\\CloudFormation\\": "src/Service/CloudFormation/src",
+            "AsyncAws\\CloudWatchLogs\\": "src/Service/CloudWatchLogs/src",
             "AsyncAws\\CodeDeploy\\": "src/Service/CodeDeploy/src",
             "AsyncAws\\DynamoDb\\": "src/Service/DynamoDb/src",
             "AsyncAws\\CodeGenerator\\": "src/CodeGenerator/src",
             "AsyncAws\\Core\\": "src/Core/src",
             "AsyncAws\\Flysystem\\S3\\": "src/Integration/Flysystem/S3/src",
             "AsyncAws\\Lambda\\": "src/Service/Lambda/src",
+            "AsyncAws\\MonologV2\\CloudWatch\\": "src/Integration/MonologV2/CloudWatch/src",
```
jderusse, line 46: We only provide 1 integration per application for all supported version. Use just `Monolog`

src/Integration/MonologV2/CloudWatch/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 Max Leonov
```
Nyholm, line 3: This should be changed.

src/Integration/MonologV2/CloudWatch/src/LogHandler.php
```diff
@@ -0,0 +1,314 @@
...
+
+use AsyncAws\CloudWatchLogs\CloudWatchLogsClient;
+use AsyncAws\CloudWatchLogs\ValueObject\LogStream;
+use AsyncAws\Core\Exception\Http\ClientException;
+use Monolog\Formatter\FormatterInterface;
+use Monolog\Formatter\LineFormatter;
+use Monolog\Handler\AbstractProcessingHandler;
+use Monolog\Logger;
+
+class LogHandler extends AbstractProcessingHandler
```
Nyholm, line 15: CloudwatchHandler

src/Integration/MonologV2/CloudWatch/src/LogHandler.php
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        $this->client = $client;
+        $this->group = $group;
+        $this->stream = $stream;
+        $this->batchSize = $batchSize;
+
+        parent::__construct($level, $bubble);
+
+        $this->savedTime = new \DateTime;
```
Nyholm, line 125: Use immutable

jderusse, line 125: Use `DateTimeImmutable`

src/Integration/MonologV2/CloudWatch/src/LogHandler.php
```diff
@@ -0,0 +1,314 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\MonologV2\CloudWatch;
```
Nyholm, line 5: Could we support both v1 and v2 is monolog? Are the Handers very different? 
I don’t remember them being very different... but I might be wrong.

composer.json
```diff
@@ -36,12 +36,14 @@
     "autoload": {
         "psr-4": {
             "AsyncAws\\CloudFormation\\": "src/Service/CloudFormation/src",
+            "AsyncAws\\CloudWatchLogs\\": "src/Service/CloudWatchLogs/src",
```
Nyholm, line 39: Make sure to add the tests in autoload-dev

src/Service/CloudWatchLogs/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+![](https://github.com/async-aws/logs/workflows/Tests/badge.svg?branch=master)
+![](https://github.com/async-aws/logs/workflows/BC%20Check/badge.svg?branch=master)
+
+An API client for CloudWatchLogs.
+
+## Install
+
+```cli
+composer require async-aws/logs
```
jderusse, line 11: `async-aws/cloud-watch-logs` ? /cc @Nyholm

src/Service/CloudWatchLogs/src/Result/PutLogEventsResponse.php
Nyholm: Hm. Yeah. Maybe. But in a different PR.

src/Service/CloudWatchLogs/tests/Unit/Input/DescribeLogStreamsRequestTest.php
```diff
@@ -0,0 +1,35 @@
...
+        $input = new DescribeLogStreamsRequest([
+            'logGroupName' => 'change me',
+            'logStreamNamePrefix' => 'change me',
+            'orderBy' => 'change me',
+            'descending' => false,
+            'nextToken' => 'change me',
+            'limit' => 1337,
+        ]);
+
+        // see https://docs.aws.amazon.com/cloudwatchlogs/latest/APIReference/API_DescribeLogStreams.html
```
jderusse, line 23: Sometimes, the generator fails to guess the url.
The Correct url is https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_DescribeLogStreams.html

(same for all others tests)

src/Integration/MonologV2/CloudWatch/src/LogHandler.php
guillaumesmo: the v2 handler adds return types to all interface methods, so I don't think it can be the same file?

jderusse: Symfony use this trick:

```php
// CloudWatchHandlerV1.php
class CloudWatchHandlerV1 {
  public function foo(){}
}

// CloudWatchHandlerV2.php
class CloudWatchHandlerV2 {
  public function foo(): void {}
}

// CloudWatchHandler.php
if (interface_exists(InterfaceThatExistsOnlyInV2::class)) {
  class CloudWatchHandler extends CloudWatchHandlerV2{}
} else {
  class CloudWatchHandler extends CloudWatchHandlerV1{}
}

Comment (jderusse) [reply src/Integration/MonologV2/CloudWatch/src/LogHandler.php:5]: But I'm also fine if we only support v2.

Comment (Nyholm) [reply src/Integration/MonologV2/CloudWatch/src/LogHandler.php:5]: If it is only return types, then it is fine: https://3v4l.org/qnojm

Comment (guillaumesmo) [reply src/Integration/MonologV2/CloudWatch/src/LogHandler.php:5]: It seems like I was wrong, although there is an interface the change is also applied on the AbstractProcessingHandler class which should be used as base class, so the trick in https://3v4l.org/qnojm is not applicable

I'll try the Symfony trick

File: src/Service/CloudWatchLogs/tests/Unit/CloudWatchLogsClientTest.php [hunk 10]
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace AsyncAws\CloudWatchLogs\Tests\Unit;
+
+use AsyncAws\CloudWatchLogs\CloudWatchLogsClient;
+use AsyncAws\CloudWatchLogs\Result\PutLogEventsResponse;
+use AsyncAws\CloudWatchLogs\ValueObject\InputLogEvent;
+use AsyncAws\CloudWatchLogs\ValueObject\PutLogEventsRequest;
```
guillaumesmo, line 8: there's another issue with the generator here

this import should be AsyncAws\CloudWatchLogs\Input\PutLogEventsRequest instead

Nyholm: Oh, that is true.. thank you

Nyholm:
Oops. This was not the intention.

src/Service/CloudWatchLogs/src/Result/PutLogEventsResponse.php
Nyholm: Fixed in master. 

You can rebase if you want to. The bot will fix it if we merge this PR without a rebase.

src/Integration/MonologV2/CloudWatch/src/LogHandler.php
Nyholm: Monolog v1 is still massively popular. If we can support both with this small trick then I would be happy.

Nyholm:
I've added the client package: https://packagist.org/packages/async-aws/cloud-watch-logs

src/Integration/MonologV2/CloudWatch/src/LogHandler.php
guillaumesmo: I tested the same class with Monolog v1 and v2 and it seems to work without modification

https://3v4l.org/vhQ3D

Not sure why maxbanton/cwh was tagged 2.0 with the upgrade to Monolog v2, that was probably not required

composer.json
```diff
@@ -10,6 +10,7 @@
         "ext-filter": "*",
         "ext-hash": "*",
         "ext-json": "*",
+        "monolog/monolog": "^1.0 || ^2.0",
```
jderusse, line 13: should be require-dev

src/Integration/Monolog/CloudWatch/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 Max Leonov
```
jderusse, line 3: revert

src/Integration/Monolog/CloudWatch/composer.json
```diff
@@ -0,0 +1,32 @@
...
+    "keywords": [
+        "aws",
+        "Monolog",
+        "CloudWatch",
+        "Logs"
+    ],
+    "license": "MIT",
+    "require": {
+        "php": "^7.2.5",
+        "async-aws/cloud-watch-logs": "0.1.x-dev",
```
jderusse, line 14: `^0.1` should work

src/Integration/Monolog/CloudWatch/LICENSE
guillaumesmo: is the code different enough to not require citing the source?

https://github.com/maxbanton/cwh/blob/master/LICENSE

src/Integration/Monolog/CloudWatch/composer.json
guillaumesmo: async-aws/cloud-watch-logs is not released yet

```
  Problem 1
    - The requested package async-aws/cloud-watch-logs ^0.1 is satisfiable by async-aws/cloud-watch-logs[0.1.x-dev] but these conflict with your requirements or minimum-stability.
```

Nyholm: It works better on CI ;)

There is some rules that allows each package do install dev-master if needed.

src/Integration/Monolog/CloudWatch/composer.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "async-aws/monolog-cloudwatch",
+    "type": "library",
+    "description": "Monolog CloudWatch Logs adapter",
```
Nyholm, line 4:
```suggestion
    "description": "Monolog integration with CloudWatch logs",
```

src/Integration/Monolog/CloudWatch/composer.json
```diff
@@ -0,0 +1,32 @@
...
+    "keywords": [
+        "aws",
+        "Monolog",
```
Nyholm, line 7:
```suggestion
        "aws",
        "aync-aws",
        "Monolog",
```

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
```diff
@@ -0,0 +1,295 @@
...
+
+            $this->buffer[] = $record;
+
+            if (\count($this->buffer) >= $this->batchSize) {
+                $this->flushBuffer();
+            }
+        }
+    }
+
+    private function checkThrottle(): void
```
Nyholm, line 165: Im not sure why this is needed. Why do we sleep 1 second? Why is 1000ms such a magic number... 

I think we should rework this in the next PR.

src/Integration/Monolog/CloudWatch/composer.json
```diff
@@ -0,0 +1,32 @@
...
+        "CloudWatch",
+        "Logs"
+    ],
+    "license": "MIT",
+    "require": {
+        "php": "^7.2.5",
+        "async-aws/cloud-watch-logs": "0.1.x-dev",
+        "monolog/monolog": "^1.1"
```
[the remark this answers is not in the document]
guillaumesmo: there must be an issue with monolog 1.0.0, but tests pass with 1.0.1

src/Integration/Monolog/CloudWatch/composer.json
guillaumesmo: same issue on CI

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
guillaumesmo: 5 RPS comes from the official documentation: https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html
I have copied that in the phpdoc

I agree the throttling mechanism is very basic. The original author allows 5 requests within the same second, instead of waiting 1 sec after the 5th request. In both cases, I would not choose any other value than 1000ms. 

Yet, this doesn't throttle requests coming from different processes, but it's still better than nothing

src/Integration/Monolog/CloudWatch/composer.json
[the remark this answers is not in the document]
guillaumesmo: Added 2.0 as well

src/Integration/Monolog/CloudWatch/composer.json
guillaumesmo: I reverted it to ^0.1-dev, please let me know if I need to change it

Nyholm: We can fix that before we tag a new version

Nyholm:
Good work today!

PR #538, opened 2020-05-05
UpperCamelCase getters and setters

jderusse:
Fixes https://github.com/async-aws/aws/pull/537#discussion_r420435886

### DIFF
--- a/src/CodeGenerator/src/Generator/PaginationGenerator.php
+++ b/src/CodeGenerator/src/Generator/PaginationGenerator.php
@@ -101,7 +101,7 @@ class PaginationGenerator
             $singlePage = empty($pagination->getOutputToken());
             $iteratorBody .= strtr('yield from $page->PROPERTY_ACCESSOR(SINGLE_PAGE_FLAG);
             ', [
-                'PROPERTY_ACCESSOR' => 'get' . $resultKey,
+                'PROPERTY_ACCESSOR' => 'get' . \ucfirst($resultKey),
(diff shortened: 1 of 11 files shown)


Nyholm:
This is fine since php method names are not case sensitive.

jderusse:
Indeed, I wonder if we shouldn't lower the property name for consistency?

Nyholm:
Na. I don’t think we should change the properties. There is no real win and it will make the generators more complicated.

src/CodeGenerator/src/Generator/PaginationGenerator.php
```diff
@@ -114,7 +114,7 @@ private function generateOutputPagination(Operation $operation, Pagination $pagi
             [$returnType, $iteratorType] = $this->typeGenerator->getPhpType($listShape);
             $iteratorTypes[] = $iteratorType;
 
-            $getter = 'get' . $resultKey;
+            $getter = 'get' . \ucfirst($resultKey);
```
guillaumesmo, line 117: this could be moved up, and be used in line 104 as well

guillaumesmo:
I was planning to have a PR for this as well, but I slept on it and it's already done :)

awesome!

src/CodeGenerator/src/Generator/PaginationGenerator.php
jderusse: good catch, thanks

jderusse:
thanks for the review

PR #539, opened 2020-05-06
Consistency in property and Input Key

jderusse:
This is an alternative of #538 that also define a consistency for properties and Input.

It provides a BC Layer and migration path for input key case.

It also provide a kind of "feature flag" per service (defined via the manifest). Use to define which service should contains a BC layer (sts, sqs) or not (code deploy).


Nyholm:
I prefer #538 instead.

PR #540, opened 2020-05-06
Fixed imports of Input classes with Request suffix

guillaumesmo:
See https://github.com/async-aws/aws/pull/537#discussion_r420615052

### DIFF
--- a/src/CodeGenerator/src/Generator/TestGenerator.php
+++ b/src/CodeGenerator/src/Generator/TestGenerator.php
@@ -261,7 +261,7 @@ class TestGenerator
                     return '""';
                 }
 
-                if ('Input' === substr($shape->getName(), -5)) {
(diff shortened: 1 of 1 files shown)


src/CodeGenerator/src/Generator/TestGenerator.php
```diff
@@ -261,7 +261,7 @@ private function getInputCode(PhpNamespace $namespace, Shape $shape, bool $inclu
                     return '""';
                 }
 
-                if ('Input' === substr($shape->getName(), -5)) {
+                if ('Input' === substr($shape->getName(), -5) || 'Request' === substr($shape->getName(), -7)) {
```
Nyholm, line 264: This is statement is still not 100% correct. The generated code is no guarantee that this will be a valid class. 

But Im fine with this for now. It is only test code and it will break loudly when we run the tests.

Nyholm:
Thank you

PR #542, opened 2020-05-06
Handle recursion in testGenerator

jderusse:
This is an alternative to #540 

- fix infinity loop in recursion
- detect input/valueObject based on recursion instead of suffix

### DIFF
--- a/src/CodeGenerator/src/Generator/TestGenerator.php
+++ b/src/CodeGenerator/src/Generator/TestGenerator.php
@@ -252,16 +252,16 @@ class TestGenerator
         $this->fileWriter->write($namespace);
     }
 
-    private function getInputCode(PhpNamespace $namespace, Shape $shape, bool $includeOptionalParameters = true, bool $nested = false): string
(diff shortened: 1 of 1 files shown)


Nyholm:
Im not 100% sure what is happening here to be honest. Could you elaborate a little?

jderusse:
This PR fixes 2 things:

**a reliable way to get the namespace of the shape:**
The current implementation check for the prefix of the Shape (either Input or Request) which seem dubious.
This PR use the `ObjectValue` namespace when dealing with a nested object. And use `Input` when it's the root object.

**better handling of recursion**
When testing `DynamoDb` `DeleteItem` we face an issue where `AttributeValue` contains a property that reference itself, which ended in an infinite loop.
The current implementation ignore nesteed object when the parent is a Map. Which has several drawback:
- does not handle recursion in List or Structure
- break nesteed not-recursive object in map.

This PR break store the visited Shape in array, which solve the 2 previous issues.

jderusse:
Tank you for the review, sorry if my PR description weren't understandable enough :(

PR #553, opened 2020-05-06
Handle enum in Map's Keys

jderusse:
This PR provides missing Enum for Maps Key.

### DIFF
--- a/src/Service/Sqs/src/Enum/QueueAttributeName.php
+++ b/src/Service/Sqs/src/Enum/QueueAttributeName.php
@@ -2,7 +2,7 @@
 
 namespace AsyncAws\Sqs\Enum;
 
-class QueueAttributeName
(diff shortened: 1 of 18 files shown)


src/Service/Sqs/src/Enum/QueueAttributeName.php
```diff
@@ -2,7 +2,7 @@
 
 namespace AsyncAws\Sqs\Enum;
 
-class QueueAttributeName
+final class QueueAttributeName
```
Nyholm, line 5: This is a bug fix. All our enums and value objects are final

jderusse: I don't know why generator update this file now and didn't triggered error before

jderusse:
thanks

PR #558, opened 2020-05-07
Use array<string, X> for Map

### LINKED ISSUE #554: DocBlock for list/map
Currently Map and List are documented as `foo[]` (except for enums cases)

Shouldn't we use `list<foo>` and `array<string, foo>`?

jderusse:
Fix #554 

Note: this PR just update Map. Using List<> for array triggers a lot of psalm errors (because of array_map in Input parsing).. I don't think it requires such effort


Nyholm:
So we are casting array values to stings so we don’t deal with arrays using int keys. Right?

jderusse:
Indeed, this PR casts array **keys** to string, and don't deal with Map using int key right.
But this is not possible case, and checked in several places: ie. https://github.com/async-aws/aws/blob/87f05f3983cb889d191a0174f6503a845b9d4aa7/src/CodeGenerator/src/Generator/ObjectGenerator.php#L213-L215

jderusse:
thanks for the review

PR #563, opened 2020-05-07
chore: add cognito support and cloudwatch logs into Symfony bundle

shouze:


Nyholm:
Looking at https://github.com/async-aws/symfony-bundle/compare/0.2.3...master, I see that we also added CloudWatchLogs. Could you add that too?

shouze:
> Looking at [async-aws/symfony-bundle@0.2.3...master](https://github.com/async-aws/symfony-bundle/compare/0.2.3...master), I see that we also added CloudWatchLogs. Could you add that too?

done!

Nyholm:
Awesome. Thank you

PR #567, opened 2020-05-07
Fix wrong signature in SES

jderusse:


src/Core/src/Signer/SignerV4.php
```diff
@@ -270,6 +271,13 @@ private function convertBodyToStream(Request $request, \DateTimeImmutable $now,
             $contentLength = $body->length();
         }
 
+        if (!$this instanceof SignerV4ForS3) {
+            $request->setBody($body = StringStream::create($body));
+
+            return;
+        }
+        // Code above can be removed in version 2.0
```
jderusse, line 279: The code above can be removed an backported to S3 Signer

Nyholm, line 279: Is code below is only for S3?

Could we add a comment to say that it was moved to the S3 client?

src/Service/S3/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## 1.0.1
```
Nyholm, line 3: Isnt this a new feature of S3?

Nyholm: Dont we also require to update the core version constraint in S3's composer.json

jderusse: switched to `1.1.0`.

The core is backward compatible, no need to bump the version.
People with s3:1.1.0 and core:1.0.0 will call the private `SignerV4::convertBodyToStream`
People with s3:1.0.0 and core: 1.0.1 will call `SignerV4::convertBodyToStream` which still works
People with s3:1.1.0 and core: 1.0.1 will call `SignerV4ForS3::convertBodyToStream`

src/Core/src/Signer/SignerV4.php
jderusse: good idea. Done

src/Core/src/Signer/SignerV4.php
```diff
@@ -101,6 +102,65 @@ protected function buildBodyDigest(Request $request, bool $isPresign): string
...
+            $contentLength = $body->length();
+        }
+
+        if (!$this instanceof SignerV4ForS3) {
+            $request->setBody($body = StringStream::create($body));
+
+            return;
+        }
+        // Code above is dedicated to S3 signature and has been moved to SignerV4ForS3.
+        // It can be removed in Core version 2.0
```
Nyholm, line 127: Sorry for the confusion. 

All the code below is for S3 < 1.1, right?  So the "code above" is for every one. The code below is for old S3 clients. The code below should be removed.

jderusse: I mixed `above` and `below`.

So.
The code "below" works only for S3 (initial bug with SES)
The code "above" early exit for != S3 signer (everyone else). Fixe the bug
The code "below" keep compatibility with S3 < 1.1
(It can not be removed before 2.0 because S3: 1.0 requires Core 1.0)

is it more clear?

Nyholm: Yes, it is better. :+1:

Nyholm:
### Core is currently 1.0.0. 
We should remove the invalid signature code and release 1.0.1. It will fix all clients but S3 ≤ 1.0.0. 

We should add a conflict with s3 < 1.1.0. 

### S3 is currently 1.0.0
We should add a new feature in this library to handle signatures and then release 1.1.0.

We don't need any conflicts because S3 1.0.0 and 1.1.0 will work with core 1.0.0 and 1.0.1. 

-----------

This works perfectly. Please double check/confirm that I'm correct.

jderusse:
I ended to the same conclusion @Nyholm 

PR updated => remove code in code + conflict <1.1

src/Core/src/Signer/SignerV4.php
```diff
@@ -101,6 +96,24 @@ protected function buildBodyDigest(Request $request, bool $isPresign): string
...
+            $contentLength = $body->length();
+        }
+
+        // If content length is unknown, use the rewindable stream to read it once locally in order to get the length
+        if (null === $contentLength) {
+            $request->setBody($body = RewindableStream::create($body));
+            $body->read();
+        }
+
+        $request->setBody($body = StringStream::create($body));
```
Nyholm, line 114: Don’t need the $body variable here

src/Core/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixed
 
+- remove code from SignerV4 dedicated to S3 requests.
```
Nyholm, line 11: Maybe we should explain the bug better. Ie “fixed invalid signatures”. And explain why it is a bug fix.

src/Core/src/Signer/SignerV4.php
```diff
@@ -101,6 +96,24 @@ protected function buildBodyDigest(Request $request, bool $isPresign): string
         return $hash;
     }
 
+    protected function convertBodyToStream(Request $request, \DateTimeImmutable $now, string $credentialString, string $signingKey, string &$signature): void
+    {
+        $body = $request->getBody();
+        if ($request->hasHeader('content-length')) {
+            $contentLength = (int) $request->getHeader('content-length');
+        } else {
+            $contentLength = $body->length();
```
Nyholm, line 105: Why do we need the length here?

src/Service/S3/src/Signer/SignerV4ForS3.php
```diff
@@ -50,4 +58,75 @@ protected function buildBodyDigest(Request $request, bool $isPresign): string
...
+        // If content length is unknown, use the rewindable stream to read it once locally in order to get the length
+        if (null === $contentLength) {
+            $request->setBody($body = RewindableStream::create($body));
+            $body->read();
+            $contentLength = $body->length();
+        }
+
+        // no need to stream small body. It's simple to convert it to string directly
+        if ($contentLength < self::CHUNK_SIZE) {
+            $request->setBody($body = StringStream::create($body));
```
Nyholm, line 80: No need to store the $body variable.

src/Core/src/Signer/SignerV4.php
jderusse: good catch

src/Core/src/Signer/SignerV4.php
jderusse: You're totally right.. This method is now, almost empty..
I've refactored it to remove all those arguments, and remove the byref variable.
Please review

Nyholm: I really like the SigningContext. :+1:

src/Service/S3/src/Signer/SignerV4ForS3.php
jderusse: refactored

src/Service/S3/src/Signer/SignerV4ForS3.php
```diff
@@ -50,4 +59,78 @@ protected function buildBodyDigest(Request $request, bool $isPresign): string
...
```
Nyholm, line 82:
```suggestion
            $request->setBody(StringStream::create($body));
```

src/Core/CHANGELOG.md
Nyholm:
```suggestion
- Fixed invalid chunking of request body. When a large request body was sent, we must not split it up to multiple parts. This causes issues on most clients. It is the preferred way for the S3Client. This version removed the invalid code from SignerV4 to make sure requests are not chunked. 
```

Nyholm: I see that you updated the comment. Maybe use a combination of our suggestions?

jderusse: I've shorten a little bit your sentence (not sure changelog is the place for detailed explanation)

jderusse:
Thank you for the review

PR #569, opened 2020-05-07
Add DynamoDB integration with PHPSessions

### LINKED ISSUE #527: Php session handler
The aws sdk comes with a [session handler](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/service_dynamodb-session-handler.html) that stores session data in dynamodb.
Since the [session handler interface](https://www.php.net/manual/en/class.sessionhandlerinterface.php) is a native php feature, this could probably be a feature of async-aws as well.

The aws/sdk session handler can easily be integrated in symfony as well.
```yaml
framework:
    session:
        handler_id: session.handler.dynamodb

services:
    session.handler.dynamodb:
        class: Aws\DynamoDb\SessionHandler
        factory: [Aws\DynamoDb\SessionHandler, fromClient]
        arguments:
            - '@aws.dynamodb'
            - {table_name: '%env(resolve:SESSION_TABLE)%'}
```

But maybe there is some room for optimisation with the async-aws bundle as well although documenting it is probably enough.

Nyholm:
This will fix #527

The PR is currently in draft. We need to update all instances of `$this->connection`. 

@Nemo64 feel free to continue this work if you want to.

### DIFF
--- a/src/Integration/Aws/DynamoDbSession/.gitattributes
+++ b/src/Integration/Aws/DynamoDbSession/.gitattributes
@@ -1,0 +1,4 @@
+/.github export-ignore
+/tests export-ignore
+/.gitignore export-ignore
+/phpunit.xml.dist export-ignore
--- a/src/Integration/Aws/DynamoDbSession/.github/FUNDING.yml
+++ b/src/Integration/Aws/DynamoDbSession/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+github: [nyholm, jderusse]
--- a/src/Integration/Aws/DynamoDbSession/.github/workflows/.editorconfig
+++ b/src/Integration/Aws/DynamoDbSession/.github/workflows/.editorconfig
@@ -1,0 +1,2 @@
+[*.yml]
+indent_size = 2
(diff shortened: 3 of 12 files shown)


Nemo64:
Ohh wow. I will definitely look into this. Thanks.

So you are ok with copying the official implementation and just adjusting it, right?

Nyholm:
Yes. I think the end result will not be super similar. I also think that there are not many ways one could implement this specific interface using DynamoDB.

I added a reference in the readme

Nemo64:
I started porting the rest of the SessionHandler but I must admit that I won't have the time to properly implement, test and document it the coming weeks.

Here is what I have so far: https://github.com/Nemo64/aws/tree/feature/sessions/src/Integration/Aws/DynamoDbSession/src

If someone else wants to pick it up I'm for that or it'll need to wait a bit.

I have 2 notes so far:
- I omitted the entire garbage collection logic from the original. It isn't needed in modern DynamoDB since there now is a [TTL feature](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html) that should be preferred in any case since it is cheaper and automatic.
- I'm a bit concerned about the DynamoDB implementation in this library. Wrapping every value into an AttributeValue object is probably a performance problem and it also makes it a lot harder to read values as you normally can just use `current($item)` to read the value and now you need to know the type to call the right getter. Not a big deal for the session handler but may be for someone seriously using the DynamoDB as main data storage.

PR #570, opened 2020-05-07
Fix Configuration's timeout cast to string

### LINKED ISSUE #546: [Core] OPTION_METADATA_SERVICE_TIMEOUT is expected to be float
From @nickaguilarh in https://github.com/async-aws/aws/pull/532#issuecomment-624848395

> Hey there
Today I was replacing the temporary fix made on our servers and I got a bug intruced on this PR,
On line 119 in Configuration.php it defines a ?string return value but the timeout is a float as specified on HttpClientInterface.php line 55.
How should we addres this?
Changing the timeout constant to string works just fine when env variable is not set, but I'm not sure if it will work in every case.
>
>I just realized this since we can't complete emulate AWS ECS Instances on local machine.

jderusse:
Fixes #546

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,8 @@
         "symfony/filesystem": "^5.0",
         "symfony/finder": "^4.4 || ^5.0",
         "symfony/http-kernel": "^4.4 || ^5.0",
-        "symfony/phpunit-bridge": "^5.0"
+        "symfony/phpunit-bridge": "^5.0",
+        "symfony/deprecation-contracts": "^2.1"
     },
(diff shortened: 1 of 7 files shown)


jderusse:
Closed in favor of https://github.com/async-aws/aws/pull/571

Nyholm:
Just a public note from what we discussed before. 

The options in `Configuration` is defining config and settings (mostly auth). Using `timeout` would be "behaviour". For this reason and because it is very rare to change the default timeout, I think we should move timeout to a constructor argument.

PR #573, opened 2020-05-07
[Monolog CloudWatch] Rename composer package

Nyholm:
I rename it to be consistent with `async-aws/cloud-watch-logs`


Nyholm:
Thank you

PR #580, opened 2020-05-08
Add a cache layer on top of SSM

### LINKED ISSUE #565: ssm parameter vault caching
I'm testing #490 and now that i'm using it I can say that the solution is better then I expected.

But I think that service needs caching since [pricing is in the cents per 1000 requests](https://aws.amazon.com/systems-manager/pricing/) compared to lambdas [cents per million requests](https://aws.amazon.com/lambda/pricing/) and the parameter loader will requests the parameters every time even if they are not needed.

A cache that stores these parameters in `cache.app` or a custom cache configuration should be fine. The idea of ssm parameters is to ease deployment and not to provide instant updates when changed. Although this should definitely be configurable.

I can try to implement something in the next time if there isn't an issue with my idea.

jderusse:
Fixes #565

### DIFF
--- a/src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
+++ b/src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
@@ -6,7 +6,6 @@ namespace AsyncAws\Symfony\Bundle\DependencyInjection;
 
 use AsyncAws\Symfony\Bundle\Secrets\CachedEnvVarLoader;
 use AsyncAws\Symfony\Bundle\Secrets\SsmVault;
-use Symfony\Component\Cache\Adapter\FilesystemAdapter;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
--- a/src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
+++ b/src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
@@ -4,13 +4,17 @@ declare(strict_types=1);
 
 namespace AsyncAws\Symfony\Bundle\DependencyInjection;
 
+use AsyncAws\Symfony\Bundle\Secrets\CachedEnvVarLoader;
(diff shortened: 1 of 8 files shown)


src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
```diff
@@ -137,6 +140,27 @@ private function registerEnvLoader(ContainerBuilder $container, array $config):
                 $config['secrets']['path'],
                 $config['secrets']['recursive'],
             ]);
+
+        if ($config['secrets']['cache']['enabled']) {
+            if (!\interface_exists(CacheInterface::class)) {
+                throw new InvalidConfigurationException(sprintf('You have enabled "async_aws.secrets.cache" but the "symfony/cache" package is not installed. Try running "composer require symfony/cache"'));
+            }
+
+            if (null === $cacheId = $config['secrets']['cache']['pool']) {
```
Nyholm, line 149: We could set the `pool` to be required. Or we can fallback to cache.system. 

I think this logic could be removed.

src/Integration/Symfony/Bundle/src/DependencyInjection/Configuration.php
```diff
@@ -48,6 +48,13 @@ public function getConfigTreeBuilder()
                         ->scalarNode('path')->info('Path to the parameters.')->defaultNull()->end()
                         ->booleanNode('recursive')->info('Retrieve all parameters within a hierarchy.')->defaultValue(true)->end()
                         ->scalarNode('client')->info('Name of the SSM client. When null, use the default SSM configuration.')->defaultNull()->end()
+                        ->arrayNode('cache')
+                            ->canBeEnabled()
+                            ->children()
+                                ->scalarNode('pool')->info('Identifier of the Symfony Cache Pool.')->defaultNull()->end()
```
Nyholm, line 54: `->default('system.cache')`

src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
jderusse: I hesitate a lot between both solution. This fallback to `cache.system` reduce complexity. 
Fixed thanks

jderusse:
thanks for the review and tips

PR #581, opened 2020-05-08
report AWS deprecation

jderusse:
Report deprecation for:
* operation: when user call `$client->operation()`
* input: when user call getter/setter, and when request is built
* result: when user call getter

note: can not trigger deprecation on constructors because ValueObject are also used by result, and `initialize`  will trigger the deprecation.

note: I didn't use `symfony/deprecation-contracts` because, the signature of the \trigger_deprecation` does not fit with our needs (no specific package nor version.)


jderusse:
thanks

PR #582, opened 2020-05-09
Run generate in parallel

jderusse:
This PR run "generate" in parallel.
This reduce time to generate all operations from 16sec to 6 (on my machine).

It uses pcntl to fork, and communicate between parent and child (report errors, and update progress bar)

It adds complexity, and maybe some compatibility issue with Window.
Do you think it worth it? Maybe by I can providing 2 versions (with/without fork)

### DIFF
--- a/src/CodeGenerator/src/Command/GenerateCommand.php
+++ b/src/CodeGenerator/src/Command/GenerateCommand.php
@@ -65,98 +65,140 @@ class GenerateCommand extends Command
         /** @var ConsoleOutputInterface $output */
         $io = new SymfonyStyle($input, $output);
 
-        $progressService = (new SymfonyStyle($input, $output->section()))->createProgressBar();
-        $progressService->setFormat(' [%bar%] %message%');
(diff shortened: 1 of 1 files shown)


Nyholm:
Nice. It was way faster for me too. 

Would it be possible to add a `--single-process` parameter to avoid issues on Windows?

jderusse:
I just pushed force this PR for a new implementation:
- less complexe: no more msg between parent and child. Just fork 
- both parallel and sequential are available. User without pcntl switch to sequential

Nyholm:
Nice. 

```
# Master
./generate --all  28.43s user 10.24s system 96% cpu 39.975 total
./generate --all --raw  10.76s user 8.75s system 97% cpu 20.111 total

# With this PR
./generate --all  34.31s user 18.93s system 539% cpu 9.870 total
./generate --all --raw  15.86s user 16.75s system 487% cpu 6.693 total
```

That is quite some improvements =)

jderusse:
thanks for the numbers

PR #587, opened 2020-05-09
Adding support for Store::flush()

Nyholm:
We can implement flush by just remove and create a table

### DIFF
--- a/src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
+++ b/src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
@@ -5,7 +5,6 @@ namespace AsyncAws\Illuminate\Cache;
 use AsyncAws\Core\Exception\Http\HttpException;
 use AsyncAws\DynamoDb\DynamoDbClient;
 use AsyncAws\DynamoDb\Enum\KeyType;
-use AsyncAws\DynamoDb\Result\TableNotExistsWaiter;
 use AsyncAws\DynamoDb\ValueObject\KeySchemaElement;
 use Illuminate\Contracts\Cache\Store;
 use Illuminate\Support\Carbon;
--- a/src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
+++ b/src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
@@ -390,7 +390,9 @@ class AsyncAwsDynamoDbStore implements Store
         }
 
         // Wait until table is removed
-        if (!$this->dynamoDb->tableNotExists(['TableName' => $this->table])->wait()->isSuccess()) {
+        $response = $this->dynamoDb->tableNotExists(['TableName' => $this->table]);
+        $response->wait();
+        if (!$response->isSuccess()) {
             throw new \RuntimeException('Could not flush DynamoDb cache. Table could not be deleted.');
         }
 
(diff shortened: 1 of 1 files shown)


src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
```diff
@@ -377,7 +379,35 @@ public function forget($key)
...
+            $this->dynamoDb->deleteTable(['TableName' => $this->table]);
+        } catch (HttpException $e) {
+            if ($e->getCode() !== 404) {
+                // Any error but "table not found"
+                throw new \RuntimeException('Could not flush DynamoDb cache. Table could not be deleted.', $e->getCode(), $e);
+            }
+        }
+
+        // Wait until table is removed
+        if (!$this->dynamoDb->tableNotExists(['TableName'=>$this->table])->isSuccess()) {
```
jderusse, line 393: should call `wait` before. otherwise, you'll get `false`of the table is not yet deleted

src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
```diff
@@ -377,7 +379,35 @@ public function forget($key)
...
+        // Create a new table
+        $this->dynamoDb->createTable([
+            'TableName'=> $this->table,
+            'KeySchema' => [
+                new KeySchemaElement(['AttributeName' => 'key', 'KeyType' => KeyType::HASH]),
+            ],
+        ]);
+
+        // Wait until table is created
+        if (!$this->dynamoDb->tableExists(['TableName'=>$this->table])->isSuccess()) {
```
jderusse, line 406: same here

src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
Nyholm: Thank you

src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
```diff
@@ -377,7 +379,35 @@ public function forget($key)
...
+            }
+        }
+
+        // Wait until table is removed
+        if (!$this->dynamoDb->tableNotExists(['TableName'=>$this->table])->isSuccess()) {
+            throw new \RuntimeException('Could not flush DynamoDb cache. Table could not be deleted.');
+        }
+
+        // Create a new table
+        $this->dynamoDb->createTable([
```
jderusse, line 398: what 's about crossed concurrency. should we use a kind of lock somewhere?

Nyholm: Hm.. But how?

The only shared resource we know about is `DynamoDb`, and Im removing it. =)

FYI: https://github.com/async-aws/aws/pull/588

jderusse: I'm not sure this worse a so complex thing.. 
Let keep it simple. Maybe we should reduce the time between delete and create table (ie. by using `$response->wait(500, 5)` by default, the TableDoesNotExists waiter poll every 20 seconds which seems huge

Nyholm: Thank you. I've updated the PR

PR #594, opened 2020-05-10
Generate EventBridge::putEvents()

Nyholm:
This method is needed for https://github.com/brefphp/symfony-messenger


src/Core/src/AwsClientFactory.php
```diff
@@ -138,6 +139,19 @@ public function dynamoDb(): DynamoDbClient
         return $this->serviceCache[__METHOD__];
     }
 
+    public function eventBridge(): EventBridgeClient
```
jderusse, line 142: Do not forget the changelog

Nyholm: Sure. I've added a changelog entry.

Thanks

PR #595, opened 2020-05-10
Add IAM client

jderusse:
For now just CRUD operation on `user`

https://twitter.com/benjamindavies/status/1258300350018322435

PR #596, opened 2020-05-10
Handle global endpoint

jderusse:
This is extract from https://github.com/async-aws/aws/pull/595 to handle globalen endpoint

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -164,7 +164,7 @@ abstract class AbstractApi
             if (null === $region && !$this->configuration->isDefault('region')) {
                 $region = $this->configuration->get('region');
             }
-            $endpoint = $this->getEndpointPattern($region) . $uri;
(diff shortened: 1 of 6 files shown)


src/Core/src/AbstractApi.php
```diff
@@ -151,11 +158,21 @@ protected function getSignerFactories(): array
      */
     private function getEndpoint(string $uri, array $query, ?string $region): string
     {
+        if (!$this->configuration->isDefault('endpoint')) {
+            $endpoint = $this->configuration->get('endpoint');
+        } else {
+            if (null === $region && !$this->configuration->isDefault('region')) {
+                $region = $this->configuration->get('region');
+            }
+            $endpoint = $this->getEndpointPattern($region) . $uri;
```
Nyholm, line 167: The `$uri` will be added twice, both here and on 175.

jderusse: good catch thanks

PR #597, opened 2020-05-10
Fix table  and remove useless end-column

jderusse:


src/Service/EventBridge/tests/Unit/Input/PutEventsRequestTest.php
```diff
@@ -10,6 +10,7 @@ class PutEventsRequestTest extends TestCase
 {
     public function testRequest(): void
     {
+        \date_default_timezone_set('UTC');
         $input = new PutEventsRequest([
             'Entries' => [new PutEventsRequestEntry([
                 'Time' => new \DateTimeImmutable('2020-05-10 12:14:00'),
```
Nyholm, line 16: We could also set the timezone on the DateTimeImmutable object. Then we dont mess with global settings

jderusse: good idea, thanks

jderusse:
thanks

PR #598, opened 2020-05-10
Get proper attributes from queue

Nyholm:
This is a workaround. It looks like the official definition of the SQS::receiveMessage is wrong... Hm.


Nyholm:
Thank you @viezel for testing

PR #600, opened 2020-05-10
Add a persisting cache layer for HTTP credentials

### LINKED ISSUE #585: Support for custom Credential cache implementations
In the official AWS SDK for PHP, the default credentials chain (\Aws\Credentials\CredentialProvider::defaultProvider) accepts a \Aws\CacheInterface in the "credentials" config key. Also, the library provides implementations of PSR 6 and 16 for that interface (\Aws\PsrCacheAdapter, \Aws\Psr16CacheAdapter)

This is especially useful when using EC2 and ECS metadata credential providers, so the credentials can be stored in an APC cache for example instead of having a HTTP call on each request

I would like to work on such a feature for AsyncAWS
could you provide me some guidelines which I should follow if I want this merged eventually?

I see that \AsyncAws\Core\Credentials\CacheProvider already acts as a memory cache, maybe a good starting point

jderusse:
This fixes #585

Adds a cache layer on top of ContainerProvider and InstanceProvider:

TODO: IniFileProvider and `WebIdentityProvider` uses STS and should be cached too (but credentials fetched .ini file should not be cached)


src/Core/src/Credentials/ChainProvider.php
```diff
@@ -60,4 +67,25 @@ public function reset()
...
         $this->lastSuccessfulProvider = [];
     }
+
+    public static function createDefaultChain(?HttpClientInterface $httpClient = null, ?LoggerInterface $logger = null, $cache = null): CredentialProvider
+    {
+        $httpClient = $httpClient ?? HttpClient::create();
+        $logger = $logger ?? new NullLogger();
+        if (null === $cache) {
+            if (\interface_exists(CacheInterface::class)) {
+                $cache = new FilesystemAdapter('async-aws');
```
Nyholm, line 77: I’m not sure it is a good idea to create cache adapters in the constructor.

src/Core/src/Credentials/ChainProvider.php
```diff
@@ -60,4 +67,25 @@ public function reset()
     {
         $this->lastSuccessfulProvider = [];
     }
+
+    public static function createDefaultChain(?HttpClientInterface $httpClient = null, ?LoggerInterface $logger = null, $cache = null): CredentialProvider
```
Nyholm, line 71: Cool. I like the idea of this function.

src/Core/src/Credentials/ChainProvider.php
```diff
@@ -60,4 +67,25 @@ public function reset()
...
+    {
+        $httpClient = $httpClient ?? HttpClient::create();
+        $logger = $logger ?? new NullLogger();
+        if (null === $cache) {
+            if (\interface_exists(CacheInterface::class)) {
+                $cache = new FilesystemAdapter('async-aws');
+            }
+        } elseif (!$cache instanceof CacheInterface) {
```
Nyholm, line 79: I really think we should allow psr6 too.

src/Core/src/Credentials/PersistingCacheProvider.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->decorated = $decorated;
+        $this->cache = $cache;
+    }
+
+    public function getCredentials(Configuration $configuration): ?Credentials
+    {
+        return $this->cache->get(sha1(\serialize([$configuration, \get_class($this->decorated)])), function (ItemInterface $item) use ($configuration) {
+            $credential = $this->decorated->getCredentials($configuration);
+            if (null === $credential) {
+                $item->expiresAfter(0);
```
Nyholm, line 36: This is also good to remember, right?
Or every user http request will make this request and get nothing back. 

Let’s save it for 5 seconds?

src/Core/src/Credentials/PersistingCacheProvider.php
```diff
@@ -0,0 +1,44 @@
...
+use Symfony\Contracts\Cache\ItemInterface;
+
+/**
+ * Cache the Credential generated by the decorated CredentialProvider in a persisting pool.
+ *
+ * The Credential will be reused until it expires.
+ *
+ * @author Jérémy Derussé <jeremy@derusse.com>
+ */
+final class PersistingCacheProvider implements CredentialProvider
```
Nyholm, line 18: Let’s create a similar class for psr6.

src/Core/src/Credentials/PersistingCacheProvider.php
jderusse: I though like you,but then switch to 0:
In fact HTTP calls always return a credential. There are 2 cases where it return null:
- endpoint is not reachable (ie, the chain use InstanceCredential in a server that is not running in AWS)
- endpoint is temporary unreachable (network issue, when fetching credential, or timeout (remember 1.0 sec))

The first one should not happens in production. InstanceCredential is the last provider in the chain, and reaching it while the application in not runing in AWS means a wrong configuration.
The second one is temporary and should not be cached IMHO

src/Core/src/Credentials/ChainProvider.php
jderusse: thanks

src/Core/src/Credentials/PersistingCacheProvider.php
Nyholm: Okey, You are correct. Thank you

src/Core/src/Credentials/Credentials.php
```diff
@@ -21,16 +23,20 @@ final class Credentials implements CredentialProvider
...
 
+    private $cacheable;
+
     public function __construct(
         string $accessKeyId,
         string $secretKey,
         ?string $sessionToken = null,
-        ?\DateTimeImmutable $expireDate = null
+        ?\DateTimeImmutable $expireDate = null,
+        bool $cacheable = false
```
jderusse, line 33: The purpose of this parameter is to tell if the Credential can be persisted.
Actually it's used once: When credential are fetched from .ini file.

I use it to avoid the following case:
- user call an API
- credential are fetched from .ini and stored in cache
- user update the .ini file
- user call the API
-> cached crednetial are re-used => user don't understand what happens


Do you think it worse it?

src/Core/src/Credentials/Credentials.php
```diff
@@ -21,16 +23,20 @@ final class Credentials implements CredentialProvider
...
     private $expireDate;
 
+    private $cacheable;
+
     public function __construct(
         string $accessKeyId,
         string $secretKey,
         ?string $sessionToken = null,
-        ?\DateTimeImmutable $expireDate = null
+        ?\DateTimeImmutable $expireDate = null,
```
Nyholm, line 32: Could we change the order of these? 
If the `Credentials` are not cacheable, then we wound not have an expiry date.

src/Core/src/Credentials/Credentials.php
Nyholm: Yes. I like it. Good thinking.

src/Core/src/Credentials/IniFileProvider.php
```diff
@@ -64,7 +68,9 @@ private function getCredentialsFromProfile(array $profilesData, string $profile,
             return new Credentials(
                 $profileData[IniFileLoader::KEY_ACCESS_KEY_ID],
                 $profileData[IniFileLoader::KEY_SECRET_ACCESS_KEY],
-                $profileData[IniFileLoader::KEY_SESSION_TOKEN] ?? null
+                $profileData[IniFileLoader::KEY_SESSION_TOKEN] ?? null,
+                null,
+                false // do not cache credentials, the file may be edited by user
```
Nyholm, line 73: Thank you for the comment.

src/Core/src/Credentials/PsrCacheProvider.php
```diff
@@ -0,0 +1,42 @@
...
+
+declare(strict_types=1);
+
+namespace AsyncAws\Core\Credentials;
+
+use AsyncAws\Core\Configuration;
+use Psr\Cache\CacheItemPoolInterface;
+
+/**
+ * Cache the Credential generated by the decorated CredentialProvider with PSR6.
```
Nyholm, line 11:
```suggestion
 * Cache the Credential generated by the decorated CredentialProvider with PSR-6.
```

src/Core/src/Credentials/SymfonyCacheProvider.php
```diff
@@ -0,0 +1,43 @@
...
+declare(strict_types=1);
+
+namespace AsyncAws\Core\Credentials;
+
+use AsyncAws\Core\Configuration;
+use Symfony\Contracts\Cache\CacheInterface;
+use Symfony\Contracts\Cache\ItemInterface;
+
+/**
+ * Cache the Credential generated by the decorated CredentialProvider with Symfony Cache.
```
Nyholm, line 12:
```suggestion
 * Cache the Credential generated by the decorated CredentialProvider with Symfony Cache.
 * Symfony Cache provides stampede protection which is preferred on applications with more than 
 * 1 or 2 requests per second. 
```

src/Core/src/Credentials/Credentials.php
jderusse: thatwould be a BC break. I'm not sure it worth it

src/Core/src/Credentials/SymfonyCacheProvider.php
jderusse: that more clear. thanks

src/Core/src/Credentials/Credentials.php
Nyholm: Oh, Yes, You are correct. 

I thought both of these were new parameters.

src/Core/src/Credentials/ConfigurationProvider.php
```diff
@@ -23,7 +23,9 @@ public function getCredentials(Configuration $configuration): ?Credentials
             return new Credentials(
                 $configuration->get(Configuration::OPTION_ACCESS_KEY_ID),
                 $configuration->get(Configuration::OPTION_SECRET_ACCESS_KEY),
-                $configuration->get(Configuration::OPTION_SESSION_TOKEN)
+                $configuration->get(Configuration::OPTION_SESSION_TOKEN),
+                null,
```
Nyholm, line 27: What expire date will be used here?

Nyholm: Should we even cache this BTW? It is quicker to read from memory that to use any cache

src/Core/src/Credentials/Credentials.php
```diff
@@ -62,4 +68,22 @@ public function getCredentials(Configuration $configuration): ?Credentials
     {
         return $this->isExpired() ? null : $this;
     }
+
+    public static function adjustExpireDate(?\DateTimeImmutable $expireDate, ?\DateTimeImmutable $reference = null): ?\DateTimeImmutable
```
Nyholm, line 72: Should we force this to never return nullable?

src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
```diff
@@ -93,9 +97,32 @@ private function addServiceDefinition(ContainerBuilder $container, string $name,
...
         }
 
+        if ((null === $credentialServiceId = $config['credential_provider']) && \class_exists(SymfonyCacheProvider::class) && \interface_exists(CacheInterface::class)) {
+            $credentialServiceId = 'async_aws.credential';
+            if (!$container->hasDefinition($credentialServiceId)) {
+                $container->Register($credentialServiceId, CredentialProvider::class)
+                    ->setFactory([ChainProvider::class, 'createDefaultChain'])
+                    ->setArguments([$httpClient, $logger]);
+
+                $container->Register('async_aws.credential.cache', SymfonyCacheProvider::class)
```
Nyholm, line 107: Is it really capital R?

x3

src/Core/src/Credentials/ConfigurationProvider.php
jderusse: For AWS when null: never expires.
When using our new CacheProvider, it uses the CacheAdapter default lifetime.

> It is quicker to read from memory that to use any cache

Yes, we may avoid storing theme for nothing. But when decorating the ChainProvider with a cache layer, the cache will always be called for read.

src/Core/src/Credentials/Credentials.php
```diff
@@ -62,4 +68,22 @@ public function getCredentials(Configuration $configuration): ?Credentials
...
+    {
+        if (null === $expireDate) {
+            return $expireDate;
```
Nyholm, line 75: I missed this. Could we change this?
```suggestion
            return null;
```

src/Core/src/Credentials/Credentials.php
Nyholm: No, we cant, we are returning null if `$expireDate` is null.

src/Core/src/Credentials/IniFileProvider.php
```diff
@@ -95,7 +101,7 @@ private function getCredentialsFromRole(array $profilesData, array $profileData,
             return null;
         }
 
-        $stsClient = new StsClient(isset($profilesData[$sourceProfileName][IniFileLoader::KEY_REGION]) ? ['region' => $profilesData[$sourceProfileName][IniFileLoader::KEY_REGION]] : [], $sourceCredentials);
+        $stsClient = new StsClient(isset($profilesData[$sourceProfileName][IniFileLoader::KEY_REGION]) ? ['region' => $profilesData[$sourceProfileName][IniFileLoader::KEY_REGION]] : [], $sourceCredentials, $this->httpClient);
```
Nyholm, line 104: Could we make this into 3 lines?

src/Core/src/Credentials/Credentials.php
jderusse: yes you're right.

src/Core/src/Credentials/PsrCacheProvider.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public function getCredentials(Configuration $configuration): ?Credentials
+    {
+        $item = $this->cache->getItem('AsyncAws.Credentials.' . sha1(\serialize([$configuration, \get_class($this->decorated)])));
+        if (!$item->isHit()) {
+            $item->set($credential = $this->decorated->getCredentials($configuration));
+
+            if (null !== $credential && $credential->isCacheable()) {
+                $item->expiresAt($credential->getExpireDate());
```
Nyholm, line 36: What if `$credential->getExpireDate()` returns null?

src/Core/src/Credentials/SymfonyCacheProvider.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->cache = $cache;
+    }
+
+    public function getCredentials(Configuration $configuration): ?Credentials
+    {
+        return $this->cache->get('AsyncAws.Credentials.' . sha1(\serialize([$configuration, \get_class($this->decorated)])), function (ItemInterface $item) use ($configuration) {
+            $credential = $this->decorated->getCredentials($configuration);
+
+            if (null !== $credential && $credential->isCacheable()) {
+                $item->expiresAt($credential->getExpireDate());
```
Nyholm, line 38: Same here.

src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
jderusse: wrote to much golang today

src/Core/src/Credentials/Credentials.php
jderusse: can be deleted since `adjustExpireDate` expect a `\DateTimeImmutable`

src/Core/src/Credentials/PsrCacheProvider.php
jderusse: `$item->expiresAt(null)` is valid and use the default pool TTL

src/Core/src/Credentials/IniFileProvider.php
jderusse: done

src/Core/src/Credentials/PsrCacheProvider.php
Nyholm: But if the default cache for the pool is 10 years, then this will break the application eventually.

jderusse: but if the credential does not expires, then why couldn't we cache it for 10 years?

Note: All credentials fetched from HTTP have an expireDate. Maybe we should cache only credential with expiration?

Nyholm: That is definitely more safe.

jderusse: done. And it also removes the `cacheable` propery in credential

jderusse: thanks for the hint

Nyholm:
Thank you for this. 
I'll merge when CI is green

PR #602, opened 2020-05-11
Use correct filename in docs for Illuminate Cache

hmazter:
Fix the filename for the config instructions for Illuminate Cache


Nyholm:
Oops, Thank you for the fix

PR #604, opened 2020-05-11
Fix minimal requirement for clients with global endpoint

jderusse:
Metho `getEndpointPattern` will be available in core:1.2


Nyholm:
Thank you.

PR #605, opened 2020-05-12
Refactor endpoint generation

jderusse:
The current implementation of IAM is ~wrong: AWS does not provide regional endpoint.
After digging in AWS SDK metadata, I found that all endpoint are listed in an endpoints.json file

And this file contains also advance metadata that are leveraged in this PR to:
- handle more region (-fips, -gov, -iso, -china)
- handle signature dedicated to region (S3 for eu-west-1)
- handle region not available for the given service
- fix the issue when configuring a region with a global client (ie. `new IamClient(['region' => 'eu-west-1'])`)

I think we should merge this PR before releasing a new version of Core, Iam or S3 (that use the removed `getEndpointPattern` method)

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -146,12 +146,15 @@ abstract class AbstractApi
     /**
      * This method is a BC layer for client that does not require core:^1.2.
      *
-     * @return array{endpoint: string, signRegion: string, signService: string, signVersions: list<string>}
+     * @return array{endpoint: string, signRegion: string, signService: string, signVersions: string[]}
      */
     protected function getEndpointMetadata(string $region): array
     {
+        /** @var string $endpoint */
+        $endpoint = $this->configuration->get('endpoint');
+
         return [
-            'endpoint' => $this->configuration->get('endpoint'),
+            'endpoint' => $endpoint,
             'signRegion' => $region,
             'signService' => $this->getSignatureScopeName(),
             'signVersions' => [$this->getSignatureVersion()],
@@ -166,7 +169,9 @@ abstract class AbstractApi
      */
     private function getEndpoint(string $uri, array $query, ?string $region): string
     {
-        $metadata = $this->getEndpointMetadata($region = $this->configuration->get('region'));
+        /** @var string $region */
+        $region = $region ?? $this->configuration->get('region');
+        $metadata = $this->getEndpointMetadata($region);
         if (!$this->configuration->isDefault('endpoint')) {
             $endpoint = $this->configuration->get('endpoint');
         } else {
@@ -189,8 +194,8 @@ abstract class AbstractApi
 
     private function getSigner(?string $region)
     {
-        $region = $region ?? $this->configuration->get(Configuration::OPTION_REGION);
         /** @var string $region */
+        $region = $region ?? $this->configuration->get(Configuration::OPTION_REGION);
         if (!isset($this->signers[$region])) {
             $metadata = $this->getEndpointMetadata($region);
             $factories = $this->getSignerFactories();
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -144,14 +144,22 @@ abstract class AbstractApi
     }
 
     /**
-     * This method is a BC layer for client that does not require core:^1.2.
+     * Returns the AWS endpoint metadata for the given region.
+     * When user did not provide a region, the client have to either return a global endpoint or fallback to
+     * the Configuration::DEFAULT_REGION constant.
+     *
+     * This implementation is a BC layer for client that does not require core:^1.2.
+     *
+     * @param ?string $region region provided by the user (without fallback to a default region)
      *
      * @return array{endpoint: string, signRegion: string, signService: string, signVersions: string[]}
      */
-    protected function getEndpointMetadata(string $region): array
+    protected function getEndpointMetadata(?string $region): array
     {
         /** @var string $endpoint */
         $endpoint = $this->configuration->get('endpoint');
+        /** @var string $region */
+        $region = $region ?? $this->configuration->get('region');
 
         return [
             'endpoint' => $endpoint,
@@ -162,15 +170,16 @@ abstract class AbstractApi
     }
 
     /**
-     * Fallback function for getting the endpoint. This could be overridden by any APIClient.
+     * Build the endpoint full uri.
      *
-     * @param string $uri   or path
-     * @param array  $query parameters that should go in the query string
+     * @param string  $uri    or path
+     * @param array   $query  parameters that should go in the query string
+     * @param ?string $region region provided by the user in the `@region` parameter of the Input
      */
     private function getEndpoint(string $uri, array $query, ?string $region): string
     {
         /** @var string $region */
-        $region = $region ?? $this->configuration->get('region');
+        $region = $region ?? $this->configuration->isDefault('region') ? null : $this->configuration->get('region');
         $metadata = $this->getEndpointMetadata($region);
         if (!$this->configuration->isDefault('endpoint')) {
             $endpoint = $this->configuration->get('endpoint');
@@ -180,8 +189,8 @@ abstract class AbstractApi
 
         /** @psalm-suppress PossiblyNullArgument */
         $endpoint = strtr($endpoint, [
-            '%region%' => $region,
-            '%service%' => $this->getServiceCode(),
+            '%region%' => $region ?? $this->configuration->get('region'),
+            '%service%' => $this->getServiceCode(), // if people provides a custom endpoint 'http://%service%.localhost/
         ]);
 
         $endpoint .= $uri;
@@ -192,10 +201,13 @@ abstract class AbstractApi
         return $endpoint . (false === \strpos($endpoint, '?') ? '?' : '&') . http_build_query($query);
     }
 
+    /**
+     * @param ?string $region region provided by the user in the `@region` parameter of the Input
+     */
     private function getSigner(?string $region)
     {
         /** @var string $region */
-        $region = $region ?? $this->configuration->get(Configuration::OPTION_REGION);
+        $region = $region ?? $this->configuration->isDefault('region') ? null : $this->configuration->get('region');
         if (!isset($this->signers[$region])) {
             $metadata = $this->getEndpointMetadata($region);
             $factories = $this->getSignerFactories();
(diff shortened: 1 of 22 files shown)


src/Core/src/AbstractApi.php
```diff
@@ -144,12 +144,21 @@ protected function getSignerFactories(): array
     }
 
     /**
-     * @psalm-suppress InvalidNullableReturnType
+     * This method is a BC layer for client that does not require core:^1.2.
```
stof, line 147: should we deprecate it ?

src/Service/CognitoIdentityProvider/src/CognitoIdentityProviderClient.php
```diff
@@ -220,6 +220,61 @@ public function verifySoftwareToken($input): VerifySoftwareTokenResponse
...
+                    'endpoint' => 'https://cognito-idp-fips.us-west-2.amazonaws.com',
+                    'signRegion' => 'us-west-2',
+                    'signService' => 'cognito-idp',
+                    'signVersions' => [
+                        0 => 'v4',
+                    ],
+                ];
+        }
+
+        throw new \InvalidArgumentException(sprintf('The region "%s" is not supported by "CognitoIdentityProvider".', $region));
```
stof, line 275: should it be a dedicated exception rather than a generic InvalidArgumentException ?

src/Service/S3/src/S3Client.php
```diff
@@ -651,9 +651,188 @@ public function uploadPart($input): UploadPartOutput
...
+                        0 => 's3v4',
+                    ],
+                ];
+            case 'us-west-1':
+                return [
+                    'endpoint' => 'https://s3.us-west-1.amazonaws.com',
+                    'signRegion' => 'us-west-1',
+                    'signService' => 's3',
+                    'signVersions' => [
+                        0 => 's3',
```
stof, line 811: Do we actually need to cover the non-v4 signature support for some regions ? Only v4 is implemented in the project anyway (which is supported everywhere)

src/Service/CognitoIdentityProvider/src/CognitoIdentityProviderClient.php
stof: I suggest adding an `UnsupportedRegionException`.

src/Core/src/AbstractApi.php
Nyholm: New clients will implement this. Like `EventBridgeClient::getEndpointMetadata()`. The method should be declared abstract, but it has been added here to allow old versions of clients to still work as expected. 

So, no, we should not deprecate it. Is that right @jderusse?

src/Service/S3/src/S3Client.php
jderusse: `s3v4` is also supported in S3Client. From my last test V4 does not works on S3 endpoints.

I hesitate to filter this list to just keep the best version we knwo. But this means, people won't be able to add their own implementation (by extenting a Client for Instance).

WDYT

src/Core/src/AbstractApi.php
```diff
@@ -144,12 +144,21 @@ protected function getSignerFactories(): array
...
-        /** @psalm-suppress NullableReturnStatement */
-        return $this->configuration->get('endpoint');
+        /** @var string $endpoint */
+        $endpoint = $this->configuration->get('endpoint');
+
+        return [
+            'endpoint' => $endpoint,
+            'signRegion' => $region,
+            'signService' => $this->getSignatureScopeName(),
+            'signVersions' => [$this->getSignatureVersion()],
```
Nyholm, line 160: We should deprecate these functions in SQS, S3 and SES. And then configure the generator to not include them in other clients. But that should be done in another PR.

src/Core/src/AbstractApi.php
jderusse: Yes. I think that once every clients will requires `core:^1.2` should can make it abstract.
Maybe we can trigger a deprecation telling `this method should not be called, update your client`

note: With this PR, we don't need the 3 others abstract methods (`getSignatureScopeName`, `getSignatureVersion` and `getServiceCode`)

src/Core/src/Sts/StsClient.php
```diff
@@ -89,9 +89,121 @@ public function getCallerIdentity($input = []): GetCallerIdentityResponse
...
+            case 'eu-west-2':
+            case 'eu-west-3':
+            case 'me-south-1':
+            case 'sa-east-1':
+            case 'us-east-1':
+            case 'us-east-2':
+            case 'us-west-1':
+            case 'us-west-2':
+                return [
+                    'endpoint' => 'https://sts.%region%.amazonaws.com',
```
bendavies, line 116: reading https://docs.aws.amazon.com/IAM/latest/UserGuide/programming.html#IAMEndpoints, it seems safer to use the global endpoint?

bendavies: meh, I guess it's better to respect what the user requests.

src/Core/src/AbstractApi.php
jderusse: agreed

src/Service/CognitoIdentityProvider/src/CognitoIdentityProviderClient.php
jderusse: thanks for the suggestion

src/Core/src/Sts/StsClient.php
jderusse: Agreed, if the user choose a region, it should be used.

But because, configuration have a default value for region, user will always have a default region. I wonder, if `us-east-1` is enabled by default for STS.

If not, the fix would be to allow null region in this method, and take `null` as `global`. 
WDYT

src/Core/src/AbstractApi.php
Nyholm: > Yes. I think that once every clients will requires core:^1.2 should can make it abstract.

No, we can't. We always need to consider that other people may have created their own clients. 

> Maybe we can trigger a deprecation telling this method should not be called, update your client

Yes! That sounds like a good idea.

src/Core/src/Sts/StsClient.php
bendavies: oh yes, i didn't notice you made the region non nullable. 
I think it has to remain nullable to use the global endpoint, yes.

bendavies: or at least provide some way of using the global endpoint, like you say

bendavies: config would have to be like this:

```yaml
async_aws:
    config:
        region: us-east-1

    clients:
		sts:
            region: ~
		iam:
            region: ~
```

jderusse: IMO no need to reset STS's region to null. You can not activate a service for a region while keeping another disabled.
If you setup `region: us-east-1` in the default config, it means, you have activated the region, and STS will be activated to.

The single edge-case I see is using `new StsClient([])` when `us-east-1` is not enabled. This can occurs when using `IniFileProvider` with a profile pointing to a Role

note: IAM is not impacted, as it does not have regional endpoint, and all region will fallback to global endpoint.

jderusse:
I think the the method should allow null region. I'll add this to that PR later

src/Core/src/Sts/StsClient.php
jderusse: PR updated:
* `new IamClient()` => global
* `new IamClient(['eu-west-1'])` => global
* `new IamClient(['us-east-1'])` => global
* `new StsClient()` => global
* `new StsClient(['eu-west-1'])` => eu-west-1
* `new StsClient(['us-east-1'])` => us-east-1
* `new SqsClient()` => us-east-1
* `new StsClient(['eu-west-1'])` => eu-west-1
* `new StsClient(['us-east-1'])` => us-east-1

jderusse:
I'm good with this. Just remaining comment from @stof  about signing versions

src/Service/S3/src/S3Client.php
stof: Well, I know that the s3 client must used the s3-specific signature for things to work (so `s3v4` rather than `v4`). But here, the array also lists `s3`, which is the s3-specific variant of **v2** in the official SDK.

src/Core/src/AbstractApi.php
stof: Well, then we should indeed deprecate the implementation in core, so that they get a warning telling them that their client must be updated to implement this method.

Making it abstract will be possible only in 2.0 as that's a BC break (but the deprecation warning will warn them before).

src/Core/src/Sts/StsClient.php
```diff
@@ -89,9 +89,121 @@ public function getCallerIdentity($input = []): GetCallerIdentityResponse
...
+                    'endpoint' => 'https://sts-fips.us-west-2.amazonaws.com',
+                    'signRegion' => 'us-west-2',
+                    'signService' => 'sts',
+                    'signVersions' => [
+                        0 => 'v4',
+                    ],
+                ];
+        }
+
+        return [
```
stof, line 199: Should this global endpoint be used also for unsupported regions or only for `null` ?

jderusse: hmm indeed, some endpoint may have regional + global endpoint, in that case we have a flag telling if the endpoint is regionnalized and could be used to check this case

jderusse: refactored to check if the region is supported. 

this also fix a potential issue where region with typo would have been use the global endpoint.

src/Service/S3/src/S3Client.php
jderusse: Added filter, that returns only the versions supported by the client.

note: `s3` is still returned in the list, because the client currently provides a (wrong) implementation of `s3`. In fact this is a `s3v4`  , but aliased `s3`.

This will be fixed when deprecating the `getSignatureVersion`, which will return `s3v4` in a separate PR.

src/Core/src/Sts/StsClient.php
```diff
@@ -92,6 +93,15 @@ public function getCallerIdentity($input = []): GetCallerIdentityResponse
     protected function getEndpointMetadata(?string $region): array
     {
         switch ($region) {
+            case null:
```
stof: PHP 7.2+ optimizes switch statements when all cases are strings or integers (see https://derickrethans.nl/php7.2-switch.html). So I suggest keeping the case of `null` before the switch, so that the region switch is a jump table rather than a big if/elseif.

Also, due to using loose comparison, this actually allows the empty string to use the global endpoint.

jderusse: Thanks for the hint, you should reload the page, I fixed it yesterday

src/Core/src/Sts/StsClient.php
```diff
@@ -92,16 +92,17 @@ public function getCallerIdentity($input = []): GetCallerIdentityResponse
...
+        if (null === $region) {
+            return [
+                'endpoint' => 'https://sts.amazonaws.com',
+                'signRegion' => 'us-east-1',
+                'signService' => 'sts',
+                'signVersions' => [
+                    0 => 'v4',
+                ],
+            ];
+        }
```
stof, line 104: Should we generated an empty line after this `if` ?

Nyholm:
Thank you for this PR and for all reviews.

PR #610, opened 2020-05-12
[Laravel] Storage::url() and Storage::temporaryUrl()

### LINKED ISSUE #603: [Laravel/s3] Calling Storage::url() fails
With the storage configured with `'driver' => 'async-aws-s3'`.
Calling this:
```php
$publicUrl = Storage::url($filename);
```

Which is documented here: https://laravel.com/docs/7.x/filesystem#file-urls
Fails with the following error:

```json
{
    "message": "This driver does not support retrieving URLs.",
    "exception": "RuntimeException",
    "file": "/application/vendor/laravel/framework/src/Illuminate/Filesystem/FilesystemAdapter.php",
    "line": 440,
    "trace": [
        {
            "file": "/application/app/Services/ProfileImageService.php",
            "line": 18,
            "function": "url",
            "class": "Illuminate\\Filesystem\\FilesystemAdapter",
            "type": "->"
        },
        ...
    ]
}
```

The Flystem-adapter used internally is `AsyncAws\Flysystem\S3\S3FilesystemV1`.

Laravel checks if the adapter has a method `getUrl` or is an instance of some specified adapters and calls special methods for those.
But since `S3FilesystemV1` does not have a `getUrl` or is listed in the special cases it will fail. Or am I missing something?
https://github.com/laravel/framework/blob/7.x/src/Illuminate/Filesystem/FilesystemAdapter.php#L421

Nyholm:
This will fix #603

### DIFF
--- a/src/Integration/Flysystem/S3/src/S3FilesystemV1.php
+++ b/src/Integration/Flysystem/S3/src/S3FilesystemV1.php
@@ -84,6 +84,19 @@ class S3FilesystemV1 extends AbstractAdapter implements CanOverwriteFiles
         $this->options = $options;
     }
 
+    public function getClient(): S3Client
+    {
+        return $this->client;
+    }
+
+    /**
+     * Get the S3Client bucket.
+     */
+    public function getBucket(): string
+    {
+        return $this->bucket;
+    }
+
     /**
      * {@inheritdoc}
      */
--- a/src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemAdapter.php
+++ b/src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemAdapter.php
@@ -7,7 +7,6 @@ namespace AsyncAws\Illuminate\Filesystem;
 use AsyncAws\Flysystem\S3\S3FilesystemV1;
 use AsyncAws\S3\Input\GetObjectRequest;
 use Illuminate\Filesystem\FilesystemAdapter;
-use League\Flysystem\Cached\CachedAdapter;
 
 /**
  * A small class that override the url() function.
--- a/src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemAdapter.php
+++ b/src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemAdapter.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Illuminate\Filesystem;
+
+use AsyncAws\Flysystem\S3\S3FilesystemV1;
+use AsyncAws\S3\Input\GetObjectRequest;
+use Illuminate\Filesystem\FilesystemAdapter;
+use League\Flysystem\Cached\CachedAdapter;
+
+/**
+ * A small class that override the url() function.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class AsyncAwsFilesystemAdapter extends FilesystemAdapter
+{
+    /**
+     * Return a presigned URL to the S3 object.
+     *
+     * @param string             $path
+     * @param \DateTimeInterface $expiration
+     *
+     * @return string
+     *
+     * @throws \RuntimeException
+     */
+    public function temporaryUrl($path, $expiration, array $options = [])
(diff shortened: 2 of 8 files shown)


src/Integration/Flysystem/S3/src/S3FilesystemV1.php
```diff
@@ -84,6 +84,19 @@ public function __construct(S3Client $client, string $bucket, string $prefix = '
...
 
+    public function getClient(): S3Client
+    {
+        return $this->client;
+    }
+
+    /**
+     * Get the S3Client bucket.
+     */
+    public function getBucket(): string
```
Nyholm, line 95: The same method exists in Flysystem's S3 adapter.

hmazter:
I will test this and check it out in more depth tomorrow.

But the first thing I'm noticing is that the URL is time-limited which the vanilla Laravel URL is not. `Storage::url()` just returns the plain URL to the object, if the object has a public ACL it is viewable at that URL (for example an image uploaded with public ACL, which is what I use that for)

For signed URLs with expiration `Storage::temporaryUrl()` is used, https://github.com/laravel/framework/blob/61e00dac57ec137c52d6cd1e76de9c374dc6fae6/src/Illuminate/Filesystem/FilesystemAdapter.php#L543

Nyholm:
Oh. Yeah. I’ve implemented the temporary url. :/

Nyholm:
Blocked by https://github.com/async-aws/aws/pull/605

Nyholm:
I am happy with this PR now.

src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemAdapter.php
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @throws \RuntimeException
+     */
+    public function url($path)
+    {
+        $url = $this->temporaryUrl($path, new \DateTimeImmutable('+10minutes'));
+
+        // remove all query parameters.
+        if (false === $pos = strpos($url, '?')) {
+            throw new \RuntimeException('Expected presigned URL to include a quiery string');
```
jderusse, line 63: query

src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemAdapter.php
```diff
@@ -0,0 +1,87 @@
...
+    private function getFlysystemAdapter(): S3FilesystemV1
+    {
+        if (!method_exists($this->driver, 'getAdapter')) {
+            throw new \RuntimeException(sprintf('Could not call "getAdapter" of class "%s"', \get_class($this->driver)));
+        }
+
+        $adapter = $this->driver->getAdapter();
+
+        if ($adapter instanceof CachedAdapter) {
+            $adapter = $adapter->getAdapter();
```
jderusse, line 78: Is it possible to have several layers of cache? ie. `MemoryCache(FileCache(ConcreteAdapter))`

Nyholm: Yes, it is. Good catch.. I copied the code from Laravel. I didnt think much

hmazter:
It tested this PR and it seems to work. 👍 
Noticed one thing though. The URL returned is a different format then the one returned "natively" by Laravel.

Url returned with this PR:
`https://s3.eu-north-1.amazonaws.com/{bucket-name}/{path}`

Url returned with native Laravel Flysystem:
`https://bucket-name.s3.eu-north-1.amazonaws.com/{path}`

Object URL in S3 Console:
`https://bucket-name.s3.eu-north-1.amazonaws.com/{path}`

It works, just didn't have the new format in our CSP image-src 😄

Nyholm:
The one with the bucket name is "virtual hosted style". I knew one was "old" and one was "new". 

Looking at the docs now: https://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAPI.html

It seams like buckets created after 2020-09-30 will only support the virutal host style. I'll better write a ticket so we update. 

Thank you for pointing this out

Nyholm:
https://github.com/async-aws/aws/issues/612

PR #613, opened 2020-05-13
Deprecate client's abstract methods

jderusse:
Following #605

### DIFF
--- a/src/CodeGenerator/src/Generator/ClientGenerator.php
+++ b/src/CodeGenerator/src/Generator/ClientGenerator.php
@@ -48,12 +48,6 @@ class ClientGenerator
 
         $classes = $namespace->getClasses();
         $class = $classes[\array_key_first($classes)];
-        if (null !== $prefix = $definition->getEndpointPrefix()) {
-            $class->addMethod('getServiceCode')
-                ->setReturnType('string')
-                ->setVisibility(ClassType::VISIBILITY_PROTECTED)
-                ->setBody("return '$prefix';");
-        }
 
         $supportedVersions = eval(sprintf('class A%s extends %s {
             public function __construct() {}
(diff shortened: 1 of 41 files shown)


src/Core/src/Sts/StsClient.php
```diff
@@ -213,16 +193,22 @@ protected function getEndpointMetadata(?string $region): array
 
     protected function getServiceCode(): string
     {
+        @trigger_error('Using the client with an old version of Core is deprecated. Run "composer update async-aws/core".', \E_USER_DEPRECATED);
```
jderusse, line 196: Shouldn't we use `core:^1.2` instead and removes those methods?

src/Service/CloudFormation/src/CloudFormationClient.php
```diff
@@ -176,16 +158,22 @@ protected function getEndpointMetadata(?string $region): array
 
     protected function getServiceCode(): string
```
Nyholm, line 159: This function can be removed in all clients but the one tagged 1.0

src/Core/src/AbstractApi.php
```diff
@@ -83,11 +84,29 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
+    protected function getServiceCode(): string
+    {
+        throw new LogicException(\sprintf('The method "%s" should not be called. The Client "%s" must implement the "%s" method.', __FUNCTION__, \get_class($this), 'getEndpointMetadata'));
+    }
 
-    abstract protected function getSignatureVersion(): string;
+    /**
+     * @deprecated
+     */
+    protected function getSignatureVersion(): string
```
Nyholm, line 98: We can just remove these and update the version constraints. Right?

jderusse: yes, in all but 1.0 clients (because they may have been extends right?)

Nyholm:
So after thinking of this for a while:

We cannot remove protected methods. In the AbstractApi we need to implement them and trigger deprecation. 

I’m all clients but SQS, s3 and ses, we can remove the override of those functions and require core 1.2. 

In SQS, s3 and ses, we must keep the override of these functions. 

I’m not sure if the abstractApi should throw exception or just trigger deprecation... hm. I think throwing is okey because all children has implemented these functions (they are currently abstract).

src/CodeGenerator/src/Generator/ClientGenerator.php
```diff
@@ -67,12 +61,17 @@ public function getVersion() {
...
+            return strtr('        return [
+                "endpoint" => "ENDPOINT",
+                "signRegion" => REGION,
+                "signService" => SIGN_SERVICE,
+                "signVersions" => SIGN_VERSIONS,
+            ];' . "\n", [
+                'ENDPOINT' => strtr($config['endpoint'], ['%region%' => '$region']),
+                'REGION' => isset($config['signRegion']) ? \var_export($config['signRegion'], true) : '$region',
+                'SIGN_SERVICE' => var_export($config['signService'], true),
+                'SIGN_VERSIONS' => \json_encode($signatureVersions),
```
Nyholm, line 73: Haha. This is an ugly trick. I thought about it on #605, but I wasn’t sure it was pretty enough. :)

I’m okey with it :)

src/Service/CloudFormation/src/CloudFormationClient.php
jderusse: thanks, done

src/CodeGenerator/src/Generator/ClientGenerator.php
jderusse: hey hey

jderusse:
We ended to the same conclusion @Nyholm .
- remove method for clients <1
- deprecate methods for client >= 1
- convert abstract method to method that report the change.

I choose to throw an exception instead of triggering error because nobody have a reason to call those method:
- people that implement a client with core < 1.2 had overridden the method
- people that implement a client with core >= 1.2 should implement the new getEndpointMetadata and don't have to use those methods

PR #616, opened 2020-05-15
Allows special char in S3 Bucket and Key

### LINKED ISSUE #614: [S3] keys cut of after hash sign / pound character ("#")
I encountered this while migrating some code over from the official AWS SDK. 

    /** @var AsyncAws\S3\S3Client $s3Client */
    $s3Client->putObject([
        'Bucket' => 'demo-bucket',
        'Key' => 'key-with#pound.txt',
        'Body' => 'this is just a test',
        'ContentType' => 'text/plain',
    ]);

Expected Result: The Object `s3://demo-bucket/key-with#pound.txt` is created.

Actual Result: The Object `s3://demo-bucket/key-with` is created.

Tested with v1.1 and dev-master. Symfony HttpClient v5.0.8.

It works as expected (see above) with the official AWS SDK. I know the [official docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html) list the 'Pound' character ("#") under "Characters to Avoid", but maybe it's still worth fixing. 

It's not a big issue for me personally, I can simply rename the troublesome objects.

jderusse:
This PR fixes #614 

2 issues:
- When building the URI, the Input's property should be url_encoded.
- The S3 signer should not encode the path in canonicalisation (see https://github.com/aws/aws-sdk-php/blob/master/src/Signature/S3SignatureV4.php#L60-L66)

### DIFF
--- a/src/CodeGenerator/src/Generator/InputGenerator.php
+++ b/src/CodeGenerator/src/Generator/InputGenerator.php
@@ -320,11 +320,11 @@ class InputGenerator
 
         $requestUri = null;
         $body['uri'] = $body['uri'] ?? '';
-        $uriStringCode = '"'.$operation->getHttpRequestUri().'"';
+        $uriStringCode = '"' . $operation->getHttpRequestUri() . '"';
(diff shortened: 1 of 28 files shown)


Nyholm:
Could you add some tests to this PR so we avoid regression?

PR #621, opened 2020-05-15
Change configuration of CloudWatchLogsHandler

guillaumesmo:
I think it's a bad practice to have inline config arguments in the CloudWatchLogsHandler constructor

this will lead to a very long argument list (see https://github.com/maxbanton/cwh/blob/master/src/Handler/CloudWatch.php#L121) and the need for users to hard-code defaults in instance creation

this PR proposes a signature change with a `$config` array with a BC layer until version 1 is released

this will also greatly simplify the bundle factory, since it will not have to take care of handling defaults

### DIFF
--- a/src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
+++ b/src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
@@ -98,23 +98,9 @@ class CloudWatchLogsHandler extends AbstractProcessingHandler
     public function __construct(
         CloudWatchLogsClient $client,
         $config,
-        /*$stream,
-        $batchSize = 10000,*/
         $level = Logger::DEBUG,
         $bubble = true
     ) {
-        if (!\is_array($config)) {
-            @\trigger_error('Creating CloudWatchLogsHandler with inline config arguments is deprecated', \E_USER_WARNING);
-            $arguments = \func_get_args();
-            $config = [
-                'group' => $arguments[1],
-                'stream' => $arguments[2],
-                'batchSize' => $arguments[3] ?? null,
-            ];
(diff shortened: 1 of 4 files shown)


src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
```diff
@@ -90,36 +80,49 @@ class CloudWatchLogsHandler extends AbstractProcessingHandler
...
+        $config /*,
         $stream,
         $batchSize = 10000,
         $level = Logger::DEBUG,
-        $bubble = true
+        $bubble = true*/
     ) {
-        if ($batchSize > 10000) {
+        if (!\is_array($config)) {
+            @\trigger_error('Creating CloudWatchLogsHandler with inline config arguments is deprecated', \E_USER_WARNING);
```
jderusse, line 105: no need to define a BC layer, the package has not yet been tagged 1.0

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
```diff
@@ -90,34 +80,35 @@ class CloudWatchLogsHandler extends AbstractProcessingHandler
...
      *  The ':' (colon) and '*' (asterisk) characters are not allowed.
-     * @param string $stream
-     * @param int    $batchSize
-     * @param int    $level
-     * @param bool   $bubble
+     * @param array{
+     *   batchSize?: int,
+     *   bubble?: bool,
+     *   group: string,
+     *   level?: int,
```
Nyholm, line 90: We should not have level and bubble here.

guillaumesmo: indeed, fixed!

thank you

PR #623, opened 2020-05-16
Suppress one more "BC warning" about TestCase

Nyholm:
The latest [Roave BC check](https://github.com/Roave/BackwardCompatibilityCheck) will find one more row saying: 

> [BC] SKIPPED: Roave\BetterReflection\Reflection\ReflectionClass "PHPUnit\Framework\TestCase" could not be found in the located source

I tested with the same core package and two different versions of Roave BC Check.


Nyholm:
No, I dont think so. Since the roave tool will checkout last tag, then run composer install. Step 2 is to checkout HEAD and run composer install. 

I may get away with the same trick as in #624. I'll test that.

Nyholm:
No, that trick didn't work. It will still fail to locale TestCase on the latest tag.

jderusse:
thanks

PR #593, opened 2020-05-10
Added broken link checker plugin

Nyholm:
Im not sure how or if this plugin works. But this might be helpful.

PR #628, opened 2020-05-16
Remove broken link checker

Nyholm:
I reverted #593 and here is why: 

1. The "build" on Netlify failed with errors. I didt get any email or other notification. 
2. The reason the build failed was because link `https://%service%.%region%.amazonaws.com` is not valid 

However.. it did find a bug. (also in this PR). 

Screenshot 2020-05-16 at 14 55 34


Nyholm:
I think @guillaumesmo helped proving my point that the broken link checker is not needed. =)

PR #630, opened 2020-05-17
Provides Virtual Hosting of Buckets

### LINKED ISSUE #612: [S3] Use virtual hosted style urls
This might have been fixed in #605. 

We are currently using "path-style request". For buckets created after 2020-09-30, this will not work. See https://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAPI.html

jderusse:
Fixes #612 

In order to let user disable the feature (required when using fakeS3/localstack/minio) I've added a new option in Client's `Configuration`. 
But instead of adding the option dedicated to S3 in the Core/Configuration, I allowed extending the Configuration
Even if I like having the Client using it dedicated Configuration, I'm not 100% happy with the changes (remove final, convert private to protected, ...). WDYT?.


src/Core/src/Configuration.php
```diff
@@ -12,7 +12,7 @@
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  * @author Jérémy Derussé <jeremy@derusse.com>
  */
-final class Configuration
+class Configuration
```
Nyholm, line 15: Hm.. Na. I think I was happy with this as final.

src/Service/S3/src/S3Client.php
```diff
@@ -47,9 +49,27 @@
...
+    /**
+     * @param S3Configuration|Configuration|array $configuration
+     */
+    public function __construct(
+        $configuration = [],
+        ?CredentialProvider $credentialProvider = null,
+        ?HttpClientInterface $httpClient = null,
+        ?LoggerInterface $logger = null
+    ) {
+        if (\is_array($configuration)) {
```
Nyholm, line 66: But what if I use a configuration object. Then I have to choose between S3Configuration and Configuration.

src/Core/src/Configuration.php
jderusse: Reverted

src/Service/S3/src/S3Client.php
jderusse: it's not an issue: If user provide a configuration, then it just don't benefit from S3 options.
The `instanceof` is check when client need to fetch "specific S3 configuration".

anyway this code is reverted

jderusse:
S3 error can be ignored, (composer.json requires `^1.2`)

Nyholm:
I wonder if the output URL falls under the BC promise.  Ie, is it a BC break to go from path based URL to virtual URLs? If yes, then we need to make virtual host opt-in.

Are there any pros/cons with doing `virtualHost: true` or `pathBased: false?`

jderusse:
Given, bucket created after September will not be compatible with path style, I think this cannot be option.

I prefer Path style for consistency with AWS Sdk.

Nyholm:
> Given, bucket created after September will not be compatible with path style

Correct. 

> I think this cannot be option.

I think it can. All the buckets created today will always be able to use path style. 

> I prefer Path style for consistency with AWS Sdk.

Are you sure the offical SDK is using path style? I thought they were using virtual host?

jderusse:
> I think it can. All the buckets created today will always be able to use path style.

people using the current version of S3 with buckets created in September, will hit a bug if they use the default configuration.

I prefer dealing with "BC break" for 1% of users than dealing with bug report (or additional painfill configuration) for 99%

> Are you sure the offical SDK is using path style? 
I was talking about the Config Name. Yesterday php and golang use opt-out `use-path-style-endpoont` options)

docs/clients/s3.md
```diff
@@ -67,3 +67,14 @@ This information is required by AWS, and cannot be guessed by AsyncAws.
...
+## Virtual Hosted-Style Requests
+
+When calling AWS endpoints, Async Aws uses [Virtual Hosted-Style Requests](https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html):
+The bucket name is part of the endpoint's host. To disable this behaviors, use
+the parameter `s3PathStyleEndpoint` when initializing the client.
+
+```php
+use AsyncAws\S3\S3Client;
+
+$s3 = new S3Client(['pathStyleEndpoint' => false]);
```
shyim, line 79: Is it now "s3PathStyleEndpoint" or "pathStyleEndpoint"? 🤔

jderusse: good catch!!! thanks

src/Core/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Add a `PsrCacheProvider` and `SymfonyCacheProvider` to persists crendentials in a cache pool
 - Add a `Credential::adjustExpireDate` method for adjusting the time according to the time difference with AWS clock
 - Support for global and regional endpoints
+- Add a `Configuration::optionExists` to let 3rd party checking if an option is available (needed by library supporting several version of core)
```
bendavies, line 12:
```suggestion
- Add a `Configuration::optionExists` to allow third parties to check if an option is available (needed by libraries supporting several version of core)
```

src/Service/S3/src/S3Client.php
```diff
@@ -652,6 +652,27 @@ public function uploadPart($input): UploadPartOutput
...
+        $bucketLen < 3 || $bucketLen > 63
+        || filter_var($bucket, \FILTER_VALIDATE_IP) // Cannot look like an IP address
+        || !preg_match('/^[a-z0-9]([a-z0-9\-]*[a-z0-9])?$/', $bucket) // Bucket cannot have dot (because of TLS)
+        || filter_var(\parse_url($configuration->get('endpoint'), \PHP_URL_HOST), \FILTER_VALIDATE_IP) // Custom endpoint cannot look like an IP address @phpstan-ignore-line
+        || \filter_var($configuration->get('s3PathStyleEndpoint'), \FILTER_VALIDATE_BOOLEAN)
+        ) {
+            return parent::getEndpoint($uri, $query, $region);
+        }
+
+        return \preg_replace(';https?://;', '$0' . $uriParts[1] . '.', parent::getEndpoint('/' . ($uriParts[2] ?? ''), $query, $region));
```
bendavies, line 673: using `;` as a pattern delimiter looks really odd to me. I've never seen that. How about `|` which seems to be a lot more common>

docs/configuration.md
```diff
@@ -94,3 +94,10 @@ An identifier for the assumed role session
 The relative path that is used to fetch credentials inside and ECS instance.
 See [IAM Roles for Tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) for more information.
 
+### s3PathStyleEndpoint
```
bendavies, line 97: `pathStyleEndpoint` seems to make more sense since....we are already in the context of an `S3Client`

bendavies: hmm, i see this is `global` config. but i stand by my point that this looks super strange:
```php
return new S3Client(['s3PathStyleEndpoint' => true]);
```

src/Service/S3/src/S3Client.php
```diff
@@ -652,6 +652,27 @@ public function uploadPart($input): UploadPartOutput
...
     }
 
+    protected function getEndpoint(string $uri, array $query, ?string $region): string
+    {
+        $uriParts = \explode('/', $uri, 3);
+        $bucket = $uriParts[1] ?? '';
+        $bucketLen = \strlen($bucket);
+        $configuration = $this->getConfiguration();
+
+        /** @psalm-suppress PossiblyNullArgument */
```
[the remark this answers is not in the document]
jderusse: beause `$configuration->get('endpoint')` can return `?string`, but will never return`null` because the `endpoint` property has a default value.

Let me thinking that we have a coupling between `S3` and `Core`, and `S3` knows how behave the internal implementation of `Core/Configuration`.... 
Shouldn't we provide a `getString` method ? /cc @Nyholm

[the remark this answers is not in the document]
jderusse: I don't use `\assert(is_string($bucket));` because I don't want executing code at runtime to hide a false positive.

docs/clients/s3.md
```diff
@@ -67,3 +67,14 @@ This information is required by AWS, and cannot be guessed by AsyncAws.
 If `ContentLength` is absent, AsyncAws will read the output before sending the
 request which could have a performance impact.
 
+## Virtual Hosted-Style Requests
+
+When calling AWS endpoints, Async Aws uses [Virtual Hosted-Style Requests](https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html):
+The bucket name is part of the endpoint's host. To disable this behaviors, use
```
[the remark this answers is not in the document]
jderusse: thanks

src/Core/CHANGELOG.md
jderusse: thanks

src/Service/S3/src/S3Client.php
jderusse: thanks for the suggestion

docs/configuration.md
jderusse: Good point... Other clients use `PathStyleEndpoint`.
updated

docs/configuration.md
```diff
@@ -94,3 +94,10 @@ An identifier for the assumed role session
 The relative path that is used to fetch credentials inside and ECS instance.
 See [IAM Roles for Tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) for more information.
 
+### pathStyleEndpoint
```
Nyholm, line 97: Should we name this to `s3PathStyleEndpoint`?

src/Service/S3/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Added
+
+- Uses Virtual Hosting of Buckets when calling AWS API
```
Nyholm, line 7:
```suggestion
- Changed from "path  style" endpoints (https://amazon.com/bucket) to "host style" endpoints (https://bucket.amazon.com). To keep the old behavior, use `s3PathStyleEndpoint:true`.
```

src/Service/S3/src/S3Client.php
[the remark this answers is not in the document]
Nyholm: Im fine with the current code. The 6 options that have a default value will never be null when using `beause $configuration->get()`. That is something that is very hard for psalm to know, that is why the annotation is fine.

docs/configuration.md
jderusse: this comment conviced me not https://github.com/async-aws/aws/pull/630#discussion_r427631052

At least, I can create a upper section `## S3 Specific configuration`

src/Service/S3/CHANGELOG.md
jderusse: thanks

docs/configuration.md
Nyholm: Fair. 
I was considering the scenario where we want to introduce an option for path style endpoint for Glacier? If someone wanted to use path style for Glacier but not S3 they couldn't do that.. 

But Im am wrong. Or. Im wrong most of the time. One could: 

```php
$glacier = new GlacierClient(['pathStyleEndpoint' => true]);
$s3 = new S3Client(['pathStyleEndpoint' => true]);
```

However, One cannot use the factory...

```php
$factory = new AwsFactory(['pathStyleEndpoint' => true]);
```

But I dont really care that much about the factory =). Using `pathStyleEndpoint` is good.

jderusse: I ended to the same conclusion. the same parameter can be used by several client ,it'll still work.

Regarding the factory, we have the same issue with all options: If someone want to use a region for S3 and another region for Glacier.

Nyholm: Yeah. You are correct

Nyholm:
CI on master fails because of https://bugs.php.net/bug.php?id=79616

I vote for merging this PR.

jderusse:
Thanks you all for the review

src/Service/S3/src/S3Client.php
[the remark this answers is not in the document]
bendavies: @Nyholm it's easy to tell psalm this. Can be done when https://github.com/vimeo/psalm/issues/3419 is fixed.

PR #631, opened 2020-05-17
Make sure passing "null" to the configuration does not have any effect

Nyholm:
This PR does not change anything*. It just adds a test so we dont change this behavior in the future. 

*We do change an edge case. For values without default you could have 
```php
$conf = Configuration::create();
$conf->get('accessKeyId') === null;
// Assert "null" is the default value

$conf = Configuration::create(['accessKeyId'=>null]);
$conf->isDefault('accessKeyId') === false; // Strange...
$conf->has('accessKeyId') === false;
$conf->get('accessKeyId') === null;
```

With this PR, we will acknowledge that `null` is default of `accessKeyId`. 

-------------

This PR allows us to write Symfony config like this. Because `default::ENV_VAR` will return `null` if `ENV_VAR` is not defined. 

```yaml
async_aws:
    config:
        region: '%env(default::AWS_REGION)%'
        accessKeyId: '%env(default::AWS_ACCESS_KEY_ID)%'
        accessKeySecret: '%env(default::AWS_SECRET_ACCESS_KEY)%'
        sessionToken: '%env(default::AWS_SESSION_TOKEN)%'
```

### DIFF
--- a/src/Core/src/Configuration.php
+++ b/src/Core/src/Configuration.php
@@ -103,7 +103,12 @@ final class Configuration
         }
 
         $configuration = new Configuration();
-        $configuration->userData = \array_fill_keys(\array_keys($options), true);
+        $configuration->userData = [];
+        foreach ($options as $key => $value) {
+            if (null !== $value) {
+                $configuration->userData[$key] = true;
+            }
+        }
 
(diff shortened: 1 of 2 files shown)


Nyholm:
Thank you =)

Issue #632, opened 2020-05-19
Using env var from DotEnv symfony component


tyx:
Hi !

Currently, using env var `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` loaded from `.env` with DotEnv component seems not working.

I have false result in this getenv call : https://github.com/async-aws/aws/blob/master/src/Core/src/Configuration.php#L96

But I had the correct result in `$_ENV` or `$_SERVER` at the same line.

shouze:
Yes this is a big classic. DotEnv component don't use getenv nor setenv because those methodes are not thread safe.

tyx:
I'm ready to make a PR if needed !

nicolas-grekas:
Please send a PR, no code should use getenv ever :)

tyx:
Could I just replace `getenv` by `$_SERVER` ?

nicolas-grekas:
Looks like you could yes.

jderusse:
@nicolas-grekas do you have some pointers about `getenv` vs `$_SERVER` vs `$_ENV`? why `getenv` is discouraged? why using `$_SERVER` over `$_ENV`?

nicolas-grekas:
getenv/putenv are not thread safe. In practice, this means that if a script calls putenv, an concurrent request can have getenv return null.
This happens trivially on Windows where thread safety is important.

shouze:
@jderusse putenv/getenv are [not thread safe](https://github.com/symfony/symfony/blob/87c9ab4cbe89e90236a453718575e2b93fe5f88b/src/Symfony/Component/Dotenv/Dotenv.php#L194). see also [this one](https://www.php.net/manual/fr/function.getenv.php#119680).

About using `$_SERVER` over `$_ENV` is probably [because of variables order](https://www.php.net/manual/en/ini.core.php#ini.variables-order) no?
![image](https://user-images.githubusercontent.com/54712/82316566-47ca3100-99cd-11ea-8b82-ef9b92c3dba5.png)

nicolas-grekas:
`$_ENV` is not populated by default on e.g Ubuntu. This makes `$_SERVER` the most reliable source for reading env values. Note that ppl should then be sure to never read an env that starts with `HTTP_` since these can be trivially confused with an HTTP header (see https://httpoxy.org/).

shouze:
I've made a PR, let me know if it's ok like that guys :wink:

PR #633, opened 2020-05-19
fix: thread safety regarding env vars

shouze:
Fix: #632

### DIFF
--- a/src/Core/src/Configuration.php
+++ b/src/Core/src/Configuration.php
@@ -93,8 +93,8 @@ final class Configuration
             foreach ($fallbackGroup as $option => $envVariableNames) {
                 $envVariableNames = (array) $envVariableNames;
                 foreach ($envVariableNames as $envVariableName) {
-                    if (false !== $value = \getenv($envVariableName)) {
-                        $options[$option] = $value;
+                    if (isset($_SERVER[$envVariableName])) {
+                        $options[$option] = $_SERVER[$envVariableName];
 
                         break;
                     }
--- a/src/Core/src/Credentials/IniFileLoader.php
+++ b/src/Core/src/Credentials/IniFileLoader.php
@@ -65,13 +65,13 @@ final class IniFileLoader
     private function getHomeDir(): string
     {
         // On Linux/Unix-like systems, use the HOME environment variable
-        if (false !== $homeDir = \getenv('HOME')) {
+        if (\is_string($homeDir = $_SERVER['HOME'] ?? null)) {
             return $homeDir;
         }
 
         // Get the HOMEDRIVE and HOMEPATH values for Windows hosts
-        $homeDrive = \getenv('HOMEDRIVE');
-        $homePath = \getenv('HOMEPATH');
+        $homeDrive = $_SERVER['HOMEDRIVE'] ?? null;
+        $homePath = $_SERVER['HOMEPATH'] ?? null;
 
         return ($homeDrive && $homePath) ? $homeDrive . $homePath : '/';
     }
--- a/generate
+++ b/generate
@@ -1,27 +1,27 @@
 #!/usr/bin/env php
 <?php
 
-$generatorDir = __DIR__.'/src/CodeGenerator';
-if (!file_exists($generatorDir.'/generate.php')) {
+$generatorDir = __DIR__ . '/src/CodeGenerator';
+if (!file_exists($generatorDir . '/generate.php')) {
     echo "Unable to find the `generate.php` script in `/src/CodeGenerator/`.\n";
     exit(1);
 }
 
-if (!file_exists(__DIR__.'/vendor/autoload.php')) {
+if (!file_exists(__DIR__ . '/vendor/autoload.php')) {
     echo "Run `composer install` before you run the `generate` script.\n";
     exit(1);
 }
 
-require __DIR__.'/vendor/autoload.php';
+require __DIR__ . '/vendor/autoload.php';
 
-if (!getenv('ASYNC_AWS_GENERATE_MANIFEST')) {
-    putenv('ASYNC_AWS_GENERATE_MANIFEST='.__DIR__.'/manifest.json');
+if (!isset($_SERVER['ASYNC_AWS_GENERATE_MANIFEST'])) {
+    $_SERVER['ASYNC_AWS_GENERATE_MANIFEST'] = __DIR__ . '/manifest.json';
 }
-if (!getenv('ASYNC_AWS_GENERATE_SRC')) {
-    putenv('ASYNC_AWS_GENERATE_SRC='.__DIR__.'/src');
+if (!isset($_SERVER['ASYNC_AWS_GENERATE_SRC'])) {
+    $_SERVER['ASYNC_AWS_GENERATE_SRC'] = __DIR__ . '/src';
 }
-if (!getenv('SYMFONY_PATCH_TYPE_DECLARATIONS')) {
-    putenv('SYMFONY_PATCH_TYPE_DECLARATIONS=deprecations=1');
+if (!isset($_SERVER['SYMFONY_PATCH_TYPE_DECLARATIONS'])) {
+    $_SERVER['SYMFONY_PATCH_TYPE_DECLARATIONS'] = 'deprecations=1';
 }
 
-require $generatorDir.'/generate.php';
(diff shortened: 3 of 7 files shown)


src/Core/src/Credentials/IniFileLoader.php
```diff
@@ -65,13 +65,13 @@ public function loadProfiles(array $filepaths): array
     private function getHomeDir(): string
     {
         // On Linux/Unix-like systems, use the HOME environment variable
-        if (false !== $homeDir = \getenv('HOME')) {
+        if (false !== $homeDir = $_SERVER['HOME']) {
```
nicolas-grekas, line 68: `if (\is_string($homeDir = $_SERVER['HOME'] ?? null)) {`

src/Core/src/Credentials/IniFileLoader.php
```diff
@@ -65,13 +65,13 @@ public function loadProfiles(array $filepaths): array
...
             return $homeDir;
         }
 
         // Get the HOMEDRIVE and HOMEPATH values for Windows hosts
-        $homeDrive = \getenv('HOMEDRIVE');
-        $homePath = \getenv('HOMEPATH');
+        $homeDrive = $_SERVER['HOMEDRIVE'];
```
nicolas-grekas, line 73: `?? null` (same below)

src/Core/src/Configuration.php
```diff
@@ -93,8 +93,8 @@ public static function create(array $options)
             foreach ($fallbackGroup as $option => $envVariableNames) {
                 $envVariableNames = (array) $envVariableNames;
                 foreach ($envVariableNames as $envVariableName) {
-                    if (false !== $value = \getenv($envVariableName)) {
-                        $options[$option] = $value;
+                    if (\array_key_exists($envVariableName, $_SERVER)) {
```
nicolas-grekas, line 96: `isset()` is enough - `null` is not a legit value anyway

Nyholm:
Will Lambda/EC2 populate the `$_SERVER` variable with the environment variables?

chalasr:
@Nyholm Yes, no issue here (belongs to the PHP runtime). See e.g. https://bref.sh/docs/runtimes/http.html#context-access

Nyholm:
That is exactly what Im worried about. This is perfectly fine to run on Lambda (bref) and Ec2: 

```php
$s3 = new S3Client();
$result = $s3->getObject(['Bucket'=>'my-private-bucket', 'Key'=>'my-file.jpg');
```

The authentication credentials is in the environment variables. 

However, after merging this PR, will the same code be able to run? Im not sure that bref is setting `$_SERVER['AWS_ACCESS_KEY_ID ']` and `$_SERVER['AWS_SECRET_ACCESS_KEY ']`. If they are not, this will PR will break their code in production runtime. 

Same thing with EC2. Is `$_SERVER['AWS_ACCESS_KEY_ID ']` available when you running CLI commands?

jderusse:
I don't know how bref works, but env variables are populated into `$_SERVER`. This PR will not be an issue for CLI command on EC2.

shouze:
@Nyholm yes for sure it will be populated into bref.

FYI, I've squashed my 2 commits into a single one I think we're ready for a merge.

Nyholm:
You are all correct. I had to test myself just to be sure. 

Thank you for the PR and thank you for all the reviews.

mnapoli:
This is so weird, this broke in my local environment (running Symfony on my dev machine with `php -S 127.0.0.1:8000 -t public`).

After debugging for an hour a failed presigned S3 URL, I realized that:

```php
getenv('HOME'); // works
$_SERVER['HOME']; // undefined
```

which broke the authentication of the SDK.

Now I remember I had to explicitly enable the environment variables in `$_SERVER` in Bref (https://github.com/brefphp/bref/blob/master/runtime/layers/fpm/php.ini#L32). I'm surprised to see that PHP's installation on MacOS with Brew does not enable it:

```ini
variables_order = "GPCS"
```

Maybe the problem is with my laptop (in which case I don't know why), but is it?

- are environment variables populated for you (`php -i | grep variables_order`) on your laptop?
- in general, if PHP recommends `GPCS` as the good value for production and development, should we still use this?

As you can probably guess I'm full of questions and have no idea if there's a problem at all here :/

shouze:
> * php -i | grep variables_order

I'm using brew under Linux and here's what I have:
```shell
❯ php -i | grep variables_order
variables_order => GPCS => GPCS
```
So far no env vars loaded too. Does launching php test server work this way?
```shell
php -c variables_order=EGPCS -S 127.0.0.1:8000 -t
``

**Comment by mnapoli** [comment]
Yes it should work (`php -c variables_order=EGPCS -i | grep variables_order` does work), but does that mean that we should revert the PR? Because it might be broken for more people then?

**Comment by shouze** [comment]
> Yes it should work (`php -c variables_order=EGPCS -i | grep variables_order` does work), but does that mean that we should revert the PR? Because it might be broken for more people then?

I'm not sure it's a good idea to revert now. Thread safety if far more important for production than being able to use php test server. I understand the frustraction/bad DX of course so it would be nice to find a way to improve DX without compromising production safety. I have no idea at the moment but I think it was important to get agree on that first.

**Comment by jderusse** [comment]
what's about fallback to gentenv when value in `$_SERVER` is null?

or use an internal `env` method that check the value of ini setting to return either `$_server` or `getenv`.

**Comment by mnapoli** [comment]
> Thread safety if far more important for production than being able to use php test server.

I'm not sure it limits to "php test server". It applies to PHP installed with brew on Mac and Linux: the PHP cli, php-fpm and `php -S`. It impacts a lot of people.

Furthermore, this is the consequence of _the default PHP config_. So it's perfectly possible it's affecting production environments too.

Actually, I had a look at the `php.ini` shipped by PHP itself:

development version: https://github.com/php/php-src/blob/56135654a0f2751068cd27317f7dab9f483ff456/php.ini-development#L637-L648
production version: https://github.com/php/php-src/blob/56135654a0f2751068cd27317f7dab9f483ff456/php.ini-production#L637-L648

In both cases, the default value is `GPCS`, and the recommended values are `GPCS`, meaning this PR does not work with these settings.

**Comment by Nyholm** [comment]
Im on OSX with PHP installed from brew: 

```
php -i | grep variables_order
variables_order => GPCS => GPCS
```

My coworker in on Ubuntu 20.04.1:

```
php -i | grep variables_order
variables_order => GPCS => GPCS
```

We both use the "stock" configuration for this value. 

----------

To clarify: 
1. Are you saying it doesn't work on `GPCS`? 
2. Does it work with `EGPCS`?

**Comment by acsfer** [comment]
@mnapoli we are calling `$_SERVER` here, not `$_ENV`, so **GPCS** defaults should satisfy this.

See also https://github.com/symfony/symfony/issues/26451#issuecomment-373153726

**Comment by mnapoli** [comment]
Wait no, I did more testing!

>    Are you saying it doesn't work on GPCS?

It doesn't work with GPCS.

>    Does it work with EGPCS?

It doesn't work with EGPCS either, I read the config too fast. `E` is for `$_ENV`, not "populating environment variables" in `$_SERVER`.

However, after testing some more, including with the official PHP docker image, here is what I see:

- GPCS or EGPCS doesn't matter
- `$_SERVER['xxx']` works in CLI
- `$_SERVER['xxx']` **doesn't work with PHP's built-in server**

So you were right @shouze, the scope seems to be more limited than I thought.

Here is how you can reproduce (works also without docker):

```
docker run --rm -it -p 8000:8000 --env TEST=abc -v $PWD:/app -w /app php -S 0.0.0.0:8000
```

Create a `index.php` file with:

```php
<?php

echo getenv('TEST');
echo $_SERVER['TEST'];
```

Only the first one will be populated. I'm not sure why this happening, could it be a bug in the built-in server?

acsfer:
See the comment on my post ;)

Nyholm:
I did some testing:  (All with "stock" php config)

| Server | `getenv` | `$_SERVER` |
| ------ | --------- | ------------ |
| php -S | Yes | No |
| php -S (with docker) | Yes | No |
| Symfony webserver | Yes | Yes |
| php-fpm Nginx | Yes | Yes |
| php-fpm on Lambda with Bref | Yes | Yes |

It might be something wrong with the built-in PHP server.

jderusse:
#769 should fix this issue.
/cc @mnapoli

PR #634, opened 2020-05-20
Ignore expires header for signing requests

pableu:
When uploading to S3 with an `Expires` header, it throws an Exception: 

```
AsyncAws\Core\Exception\Http\ClientException: HTTP 403 returned for "https://s3.eu-central-1.amazonaws.com/my-bucket/my-key.txt".
        Code:    SignatureDoesNotMatch
        Message: The request signature we calculated does not match the signature you provided. Check your key and signing method.
        ...
```

This PR fixes that. But I have to be honest, I don't know if it breaks other stuff.

I haven't found an exact documentation which headers should go into `CanonicalHeaders`, so if you know more or have a better way to fix this, feel free to take over or replace this PR by a better one :)

### DIFF
--- a/src/Core/src/Signer/SignerV4.php
+++ b/src/Core/src/Signer/SignerV4.php
@@ -24,6 +24,7 @@ class SignerV4 implements Signer
         'content-type' => true,
         'content-length' => true,
         'expect' => true,
+        'expires' => true,
         'max-forwards' => true,
         'pragma' => true,
         'range' => true,


pableu:
I think there might be a better solution for this. The [original S3 spec](https://awsdocs.s3.amazonaws.com/S3/20060301/s3-api-20060301.pdf) also recommends to include all relevant headers: 

>  in order to prevent data tampering, you should consider including all the headers in the signature calculation.

Maybe the encoding of formatting of the expires header is somehow broken? 

My (failing) code is somewhat like this: 

```
$s3Client->putObject(
    [
        'Bucket'      => static::get_bucket(),
        'Key'         => $key,
        'Body'        => $body,
        'ContentType' => $mime_content_type,
        'Expires' => new DateTimeImmutable('+24 hours')
    ]
);
```

jderusse:
We have an issue in the Signer, when canonicalizing `expires:Thu, 21 May 20 20:54:15 +0200`, the signer sorted the content of `expires` and converted it to `expires:21 May 20 20:54:15 +0200,Thu`

I've opened #635. Could you please test?

pableu:
Sounds good, thank you.

I don't have access to a computer until Sunday, I'll gladly test it then if it's still desired.

jderusse:
thanks you

pableu:
I upgraded to 1.2.0 and it works as expected. Thanks you for the fix in #635.

PR #637, opened 2020-05-21
[SimpleS3] Dont use presignUrl, use parent::getEndpoint()

Nyholm:
Blocked by #630

### DIFF
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace AsyncAws\SimpleS3;
 
-use AsyncAws\S3\Input\GetObjectRequest;
 use AsyncAws\S3\S3Client;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -17,20 +17,7 @@ class SimpleS3Client extends S3Client
...
-        $url = $this->presign($input, new \DateTimeImmutable('+10minutes'));
-
-        // remove all query parameters.
-        if (false === $pos = strpos($url, '?')) {
-            // If the client didn't have any credentials, there will not be any query string
-            return $url;
-        }
-
-        return substr($url, 0, $pos);
+        return $this->getEndpoint(sprintf('/%s/%s', $bucket, $key), [], null);
```
jderusse, line 20: key and buck should be `urlencod`

jderusse: hmm. and `/` in key should be conserved. The right uri is:
`$uri = '/' . rawurlencode($bucket) . '/' . str_replace('%2F', '/', rawurlencode($key));`

Nyholm: Yeah, You are correct. THank you

Nyholm:
Yeah, Indeed. 

Thank you for the review

Issue #566, opened 2020-05-07
[DynamoDb] Generate batchWriteItem() and batchGetItem()


Nyholm:
This might need some more logic in the code generator. 

It is needed for #551

PR #645, opened 2020-05-21
implement rds-data service

Nemo64:
Since the session handler was already done and is working great I thought I try something different. 😉

I implemented the rds data service.
I would like to use it in my [dbal-rds-data](https://github.com/Nemo64/dbal-rds-data/tree/feature/async-aws) library... in fact I already tried it and it seems to work fine in a symfony-doctrine-orm project.

A few notes:
- I really disliked the idea of having every single value wrapped in an object. With the rds data api it isn't unlikely to have hundreds of thousands of values. And it's also a bit more difficult to read the values with value objects... and also the code generator had problems with a recursive schema so I decided to adjust the generator to allow to exclude types from being converted to value objects which can be defined in `manifest.json`. These values will just be passed though.
- the `BatchExecuteStatementRequest` had array of arrays of objects in the schema for parameterSets. I had to implement that in the generator and I haven't checked if all metadata for it is correct yet... and it might not be perfectly thought though so... yeah, would be nice if someone could take a look that knows the generator better. If not than I would just remove that method from the current api since it isn't really necessary and I'm not using that method.
- I had an unexpected string instead of an int in `ExecuteStatementResponseTest.php:72` with `numberOfRecordsUpdated`. I haven't yet searched why yet.
- I haven't implemented the `ExecuteSql` api call since it is deprecated.
- I haven't looked at the integration test yet
- The generator did not populate the endpoints. I also haven't yet looked why and just manually did that.

So since there are more points than I expected, I'm now in the awkward position of having an unfinished surprise feature and having to ask if you actually want to have it.

So: am I on the right track? Is this an api you would merge? I would try to address some of the points I had this weekend then.


Nyholm:
Thank you for this PR. I see you've done some great work on a really hard API =) 
After a few hours working on generators for dynamodb it is normal to get headaches or to be really confused. 

> I really disliked the idea of having every single value wrapped in an object

Without knowing too much. I am not a big fan of move away from the specification of the API. I might have to revisit this later. 

> the BatchExecuteStatementRequest had array of arrays of objects in the schema for parameterSets.

Yes, that is indeed tricky. It is related to #566. We just need to make sure to test it properly. 

> I haven't implemented the ExecuteSql api call since it is deprecated

Good

> The generator did not populate the endpoints

The code generator does not generate all tests automatically Im afraid. There are some manual work too. 

> am I on the right track? 

Yes, you are on the right track. It is just very complicated. I wish I could offer you some advise. But it is just to look at the API specification, the docs, the generated output and the code generator to figure out what is happening where. And then write a bunch of manual tests on the generated output to make sure you are correct. 

> Is this an api you would merge? 

Yes, We merge any generate client that there is a need for.

src/Service/RDSDataService/composer.json
```diff
@@ -1,13 +1,14 @@
 {
-    "name": "async-aws/lambda",
+    "name": "async-aws/rds-data-service",
```
Nyholm, line 2: It seams like you removed a bunch of files from Lambda.

Nemo64: I noticed 😅
I copied some boilerplate form the lambda branch and at some time got confused in the branches. I'll modify the commit when i make this pull request final.

Nemo64:
I think I covered everything. Great tests btw!

I'll still check how big the overhead is if I actually use objects but not today. I'll also need to solve the recursion problem there.

My favourite implementation would be If I could completely remove the `Field` type. Then the api would be a lot cleaner and it would also probably safe some memory by not having to store that many arrays or objects.
```diff
$client = new RDSDataServiceClient();
$response = $client->executeStatement([
    'database' => 'my_database',
    'resourceArn' => 'arn:resource',
    'secretArt' => 'arn:secret',
    'sql' => 'SELECT name FROM users WHERE id = :id',
    'parameters' => [
-       ['name' => 'id', 'value' => ['longValue' => 5]],
+       ['name' => 'id', 'value' => 5],
    ]
]);

foreach ($response->getRecords() as $record) {
-   echo "name: " . $record[0]['stringValue'] . PHP_EOL;
+   echo "name: " . $record[0] . PHP_EOL;
}
```
But that would deviate from the real api and would need a smart integration into the generator. I don't know what do you think? Every implementation using it will probably have some [converter](https://github.com/Nemo64/dbal-rds-data/blob/master/src/RdsDataConverter.php) to abstract that away so doing that in the client might fit the "better developer experience" quiet well.

Nyholm:
I'll review this shortly. 

Could I bother you with one more thing? Could you remove the "array" stuff that you made to simplify the value objects?

It will be easier for me to review and to merge the PR. It will also be easier to review the next PR when you add the "array" stuff back. 

Im sure you are correct that using "array" makes most sense. But Im not ignorant with DynamoDb and I dont want this great PR to sit around until we figured everything out. .. So let's do it one step at the time. 

Thank you

Nemo64:
The problem is that the generator is not able to create the value objects because they are recursive. I assume this is because of an infinity deep metadata definition but I don't know yet.

I need to fix that anyways to even test how big the overhead actually is or If I just performed premature performance optimisation.

But sure, I'll try to find time to figure that out.

Nyholm:
> The problem is that the generator is not able to create the value objects because they are recursive.

But that is solveable. The current DynamoDb client supporting recursive value objects.

jderusse:
I can have a look if you need help

I think the difference with dynamoDb is: the circular dependency is in the request, and the array -> toObject conversion needs adjustments.

Nemo64:
@jderusse That would be awesome!

I did some digging just to see if I didn't overlook some minor detail and my initial thought about it being metadata is wrong I think.

The problematic part is generating the `ArrayOfArray` type in for `ExecuteStatementResponse` and `BatchExecuteStatementResponse`.

The request has this solved by just calling `ArrayValue::create($input["arrayValue"])` which in turn can call `array_map([ArrayValue::class, "create"], $input["arrayValues"] ?? [])` so the recursion occurs during runtime.
But during the response, the `RestJsonParser` tries to map all json values to the right object which then of course is infinite. If you change the manifest to `"arrays": ["ArrayOfArray"]` it can generate the code and it becomes fairly clear what the problem is.

I see 2 solutions:
1. Every ValueObject needs something like a `populateResult` itself so that not all logic is in the Result object. Those methods could then call each other recursively. But that would also introduce more public methods but it would work similar to the Request parsing.
2. The Result object could have a private method for every type it generates. Those could then call each other recursively.

Both solutions would change the generated code of existing Services.

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
```diff
@@ -98,7 +98,7 @@ public function generateRequestBuilder(StructureShape $shape): array
 
     protected function dumpArrayBoolean(string $output, string $input, Shape $shape): string
     {
-        return strtr('$payloadOUTPUT = INPUT ? "true" : "false";', [
+        return strtr('$payloadOUTPUT = (bool) INPUT;', [
```
jderusse, line 101: That were wrong (and nevere used)

src/CodeGenerator/src/Generator/RequestSerializer/JsonRpcSerializer.php
```diff
@@ -30,12 +30,4 @@ public function getHeaders(Operation $operation): string
             'TARGET' => sprintf('%s.%s', $operation->getService()->getTargetPrefix(), $operation->getName()),
         ]);
     }
-
-    protected function dumpArrayBoolean(string $output, string $input, Shape $shape): string
```
jderusse, line 34: no need to extends RestJson anymore (previous code fixed)

src/Service/RDSDataService/tests/Unit/Input/ExecuteStatementRequestTest.php
```diff
@@ -0,0 +1,68 @@
...
+            POST /Execute HTTP/1.0
+            Content-Type: application/json
+
+            {
+               "continueAfterTimeout": false,
+               "database": "my_database",
+               "includeResultMetadata": true,
+               "parameters": [
+                  {"name": "array", "value": {"arrayValue": {"stringValues": ["1", "2"]}}},
+                  {"name": "blob", "value": {"blobValue": "MDEyMzQ1Njc4OQ=="}},
```
jderusse, line 49: Previous code that use array, didn't base64 blobl value.

Nemo64:
A quick status update:

I fixed some missing strtr values and update the documentation to reflect the new object syntax.
I also updated the function generation to not generate `use (&$fn)` statements when not necessary. It isn't the most beautiful implementation (I use a regex to look for the `$fn` variable) but it makes phpstan happy.
Sort of, phpstan seems to not understand out of order array definitions. I reported that https://github.com/phpstan/phpstan/issues/3358 but to solve it, I tried to generate a phpstan-ignore-line into all function calls. That didn't work, I think I need to look further into that.

I also tested it in a project and it works. No performance testing yet though, I'll do that later if I revisit the array implementation again.

While testing it I noticed something strange. The [Api Documentation for the Field](https://docs.aws.amazon.com/rdsdataservice/latest/APIReference/API_Field.html) shows a property `structValue`. This was not generated. And while digging further, I noticed that that [field is actually not in the official sdk](https://github.com/aws/aws-sdk-php/blob/master/src/data/rds-data/2018-08-01/api-2.json#L329). I went to the java sdk to cross check and noticed that the field does not appear there either. I assume it's a documentation error on the aws part but still interesting.

So I need to solve the phpstan complaint. I look into that later.

jderusse:
Thanks for this PR.

- for PhpStan, use a the `.phpstan.neon` file to exclude the pattern (We try to avoid tool's comment in generated code)
- about `structValue`, I foud a reference in the SDK but in the operation `ExecuteSql`

There is a cleaner way to detect if $fn is used:
- when the function is not yet defined I flagged `$this->functions[$keyName] = true;` to avoid recursion.
- add a `if ($this->functions[$keyName] === true) {$this->functions[$keyName]=false}`
- when dumping the function check if `$this->functions[$keyName]` is `true` or `false` to define if you should inject the $fn;
```php
if (!isset($this->functions[$keyName])) {
    // prevent recursion
    $this->functions[$keyName] = true;

    $body = '$fn[FUNCTION_KEY] = function(array $json)INJECT_FN: array {
        // ....
    };';
    

    $this->functions[$keyName] = strtr($body, [
        'FUNCTION_KEY' => \var_export($keyName, true),
        'LIST_ACCESSOR' => $this->parseElement('$item', $shapeMember->getShape(), $listAccessorRequired),
        'INPUT_PROPERTY' => $shape->isFlattened() ? '$json' : '$json' . ($shapeMember->getLocationName() ? '->' . $shapeMember->getLocationName() : ''),
        'INJECT_FN' => $this->functions[$keyName] === false ? ' use (&$fn)' : '', // <== this should beevaluated AFTER this->parseElement 2 lines above
    ]);
} elseif($this->functions[$keyName] === true) {
    $this->functions[$keyName] = false;
}

return strtr($required ? '$fn[FUNCTION_KEY](INPUT)' : 'empty(INPUT) ? [] : $fn[FUNCTION_KEY](INPUT)', [
    'INPUT' => $input,
    'FUNCTION_KEY' => \var_export($keyName, true),
]);
```


A last thing. I think we should rename `RDSDataService` to `RdsDataService` for consistency.

Nemo64:
With the name, I'd tend to agree.

However: the service is called `RDSDataService` in the documentation and all official api clients. I don't like that capitalization either but I also want it to be consistent. In the end it's your project and I'm fine either way but I slightly prefer consistency with the official sdk and documentation.

I'll go though the other points later.

jderusse:
Having `RDS` alongside `Sqs`,  `Sns`, `Asy'cAws` sounds weird to me.
The official SDK is totally inconsistent (see CamelCase property in services: `Sqs::tags` vs `Sqs::QueueUrl` for instance), we should not take it as reference for naming our services IMHO.

Nemo64:
I think I resolved everything.

- I added a few exceptions to the code style check and removed the logic to avoid these code style problems I added earlier.
- I also completely removed the use statement logic and just added a code style exception there too. I think that is a far better solution than to add more complexity to the generator to figure out if it is needed or not.
- I changed the service name to `RdsDataService`. Fun fact: This is technically not even a breaking change because php is not case sensitive although autoloading would cause problems... Doesn't matter since nothing is merged yet.

But now that symfony 5.1 is released, the tests have a deprecation warning.
I could already isolate it... so the tests are now blocked by https://github.com/SymfonyTest/symfony-bundle-test/pull/29.

Adding this service now took longer than figuring out how dbal drivers work 😅

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
Nyholm: Could you revert this please?

src/Service/CloudWatchLogs/src/Result/DescribeLogStreamsResponse.php
```diff
@@ -123,7 +123,8 @@ protected function populateResult(Response $response): void
             }
 
             return $items;
-        })($data['logStreams']);
+        };
+        $this->logStreams = empty($data['logStreams']) ? [] : $fn['list-LogStreams']($data['logStreams']);
```
Nyholm, line 127: Could you help me to understand why this changed? What is the purpose? Or was it a side-effect of something?

Nemo64: The $fn construct has the purpose of handling recursion. I this file it does not change anything. But in src/Service/RdsDataService/src/Result/BatchExecuteStatementResponse.php it there are function calls within functions.

Nyholm: Thank you for the explanation.

src/Service/RdsDataService/src/Result/BatchExecuteStatementResponse.php
```diff
@@ -0,0 +1,124 @@
...
+            foreach ($json as $item) {
+                $a = isset($item) ? (string) $item : null;
+                if (null !== $a) {
+                    $items[] = $a;
+                }
+            }
+
+            return $items;
+        };
+        $this->updateResults = empty($data['updateResults']) ? [] : $fn['list-UpdateResults']($data['updateResults']);
```
Nyholm, line 122: Cool. This is impressive. 

Good work.

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
```diff
@@ -169,18 +170,27 @@ private function dumpArrayStructure(string $output, string $input, StructureShap
...
+            return strtr('
+                $indexCOUNTER = -1;
+                foreach (INPUT as $listValueCOUNTER) {
+                    $indexCOUNTER++;
+                    MEMBER_CODE
+                }
+            ',
+                [
+                    'INPUT' => $input,
+                    'COUNTER' => $cleanCounter ?: '',
```
Nyholm, line 188: Isn't `$cleanCounter` always defined?
```suggestion
                    'COUNTER' => $cleanCounter,
```

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
```diff
@@ -169,18 +170,27 @@ private function dumpArrayStructure(string $output, string $input, StructureShap
...
-                'MEMBER_CODE' => $memberCode = $this->dumpArrayElement(sprintf('%s[$index]', $output), '$listValue', $contextProperty, $memberShape, true),
-            ]);
+        try {
+            ++$counter;
+            $cleanCounter = $counter ?: '';
+
+            return strtr('
+                $indexCOUNTER = -1;
+                foreach (INPUT as $listValueCOUNTER) {
+                    $indexCOUNTER++;
```
Nyholm, line 182: Only in `BatchExecuteStatementRequest` this will be `$index1`. Do we really need COUNTER?

Nyholm: Oh, so we have `$index` and `$index1` if we got a second one.. Hm.. Okey. This if fine then.

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
Nyholm: I see that you want to remove 0. That is why you are doing this.

psalm.baseline
```diff
@@ -35,6 +35,48 @@
       <code>scalarNode</code>
     </PossiblyUndefinedMethod>
   </file>
+  <file src="src/Service/CognitoIdentityProvider/src/Result/AdminCreateUserResponse.php">
```
jderusse, line 38: hmmm.. I'm not a huge fan of ignoring ~ everything

Nyholm: Lets fix that before 0.1.0

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
jderusse: why? I didn't find any documentation about this behavior (nor in AWS SDK code)

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
jderusse: yes, with recursion, we have several counters

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
jderusse: yes. avoid `$index0` (and reduce this PR diff)

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
Nyholm: Oh, We use boolean like this on other places. 

my bad.

Nyholm:
Thank you. 
Let’s continue working on this client with smaller and easy to manage PRs :)

PR #652, opened 2020-05-23
Suppress "Swift_Transport" BC errors

Nyholm:
After this, then finally all packages will have a green BC check =)


Nyholm:
Thank you

PR #653, opened 2020-05-24
[SimpleS3] Added has(), download() and upload()

### LINKED ISSUE #337: Add support for uploading big files on S3

Nyholm:
This will fix #337

### DIFF
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -4,7 +4,12 @@ declare(strict_types=1);
 
 namespace AsyncAws\SimpleS3;
 
+use AsyncAws\Core\Stream\FixedSizeStream;
+use AsyncAws\Core\Stream\ResultStream;
+use AsyncAws\Core\Stream\StreamFactory;
 use AsyncAws\S3\S3Client;
+use AsyncAws\S3\ValueObject\CompletedMultipartUpload;
+use AsyncAws\S3\ValueObject\CompletedPart;
 
 /**
  * A simplified S3 client that hides some of the complexity of working with S3.
@@ -21,8 +26,141 @@ class SimpleS3Client extends S3Client
         return $this->getEndpoint($uri, [], null);
     }
 
-    /*
-     * TODO Add get() and put() which supports large files
-     * TODO add has()
+    public function download(string $bucket, string $key): ResultStream
+    {
+        return $this->getObject(['Bucket' => $bucket, 'Key' => $key])->getBody();
+    }
+
+    public function has(string $bucket, string $key): bool
+    {
+        return $this->objectExists(['Bucket' => $bucket, 'Key' => $key])->isSuccess();
+    }
+
+    /**
+     * @param string|resource|callable|iterable $object
+     * @param array{
+     *   ACL?: \AsyncAws\S3\Enum\ObjectCannedACL::*,
+     *   CacheControl?: string,
+     *   ContentLength?: string,
+     *   ContentType?: string,
+     *   Metadata?: array<string, string>,
+     *   PartSize?: int,
+     * } $options
      */
+    public function upload(string $bucket, string $key, $object, array $options = []): void
+    {
+        // Send upload requests with size 32 MB
+        $megabyte = 1024 * 1024;
+        $stream = $this->getStream($object, 1 * $megabyte); // split the stream in 1MB chunk
+
+        if (!empty($options['ContentLength'])) {
+            $contentLength = (int) $options['ContentLength'];
+        } else {
+            $contentLength = $stream->length();
+        }
+
+        // The maximum number of parts is 10.000. Use 64Mb sized parts unless the content length is really large.
+        $partSize = $options['PartSize'] ?? max(64, null === $contentLength ? 0 : ceil($contentLength / (10000 * 1024 * 1024)));
+        unset($options['PartSize']);
+
+        // If file is less than 64Mb, use normal upload
+        if (null !== $contentLength && $contentLength < 64 * $megabyte) {
+            $this->doSmallFileUpload($options, $bucket, $key, $object);
+
+            return;
+        }
+
+        $parts = [];
+        $uploadId = '';
+        $partNumber = 1;
+        $chunkIndex = 0;
+        $buffer = \fopen('php://temp', 'rw+');
+        foreach ($stream as $chunk) {
+            // Read chunk to resource
+            fwrite($buffer, $chunk);
+            if (++$chunkIndex < $partSize) {
+                // Continue reading chunk into memory
+                continue;
+            }
+
+            // Verify first part
+            if (1 === $partNumber) {
+                /** @var string $uploadId */
+                $uploadId = $this->createMultipartUpload(array_merge($options, ['Bucket' => $bucket, 'Key' => $key]))->getUploadId();
+            }
+
+            // Start upload
+            $parts[] = $this->doMultipartUpload($bucket, $key, $uploadId, $partNumber, $buffer);
+            ++$partNumber;
+            $buffer = \fopen('php://temp', 'rw+');
+            $chunkIndex = 0;
+        }
+
+        if ($chunkIndex > 0) {
+            if (empty($uploadId)) {
+                /*
+                 * The first and only part is too small to upload using MultipartUpload.
+                 * AWS has a limit of minimum 5MB.
+                 *
+                 * Lets use a normal upload.
+                 */
+                $this->doSmallFileUpload($options, $bucket, $key, $buffer);
+
+                return;
+            }
+
+            // upload last chunk
+            $parts[] = $this->doMultipartUpload($bucket, $key, $uploadId, $partNumber, $buffer);
+        }
+
+        if (empty($parts)) {
+            // The upload did not contain any data.
+            $this->doSmallFileUpload($options, $bucket, $key, '');
+
+            return;
+        }
+
+        $this->completeMultipartUpload([
+            'Bucket' => $bucket,
+            'Key' => $key,
+            'UploadId' => $uploadId,
+            'MultipartUpload' => new CompletedMultipartUpload(['Parts' => $parts]),
+        ]);
+    }
+
+    private function getStream($object, int $chunkSize): FixedSizeStream
+    {
+        return FixedSizeStream::create(
+            StreamFactory::create($object, $chunkSize),
+            $chunkSize
+        );
+    }
+
+    private function doMultipartUpload(string $bucket, string $key, string $uploadId, int $partNumber, $buffer): CompletedPart
+    {
+        try {
+            $response = $this->uploadPart([
+                'Bucket' => $bucket,
+                'Key' => $key,
+                'UploadId' => $uploadId,
+                'PartNumber' => $partNumber,
+                'Body' => $buffer,
+            ]);
+
+            return new CompletedPart(['ETag' => $response->getETag(), 'PartNumber' => $partNumber]);
+        } catch (\Throwable $e) {
+            $this->abortMultipartUpload(['Bucket' => $bucket, 'Key' => $key, 'UploadId' => $uploadId]);
+
+            throw $e;
+        }
+    }
+
+    private function doSmallFileUpload(array $options, string $bucket, string $key, $object)
+    {
+        $this->putObject(array_merge($options, [
+            'Bucket' => $bucket,
+            'Key' => $key,
+            'Body' => $object,
+        ]));
+    }
 }
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -49,9 +49,9 @@ class SimpleS3Client extends S3Client
      */
     public function upload(string $bucket, string $key, $object, array $options = []): void
     {
-        // Send upload requests with size 32MB
         $megabyte = 1024 * 1024;
-        $stream = $this->getStream($object, 1 * $megabyte); // split the stream in 1MB chunk
+        // Split the stream in 1MB chunk
+        $stream = $this->getStream($object, 1 * $megabyte);
 
         if (!empty($options['ContentLength'])) {
             $contentLength = (int) $options['ContentLength'];
@@ -88,13 +88,13 @@ class SimpleS3Client extends S3Client
                 continue;
             }
 
-            // Verify first part
+            // We have a good chunk of data to upload. If this is the first part, then get uploadId.
             if (1 === $partNumber) {
                 /** @var string $uploadId */
                 $uploadId = $this->createMultipartUpload(array_merge($options, ['Bucket' => $bucket, 'Key' => $key]))->getUploadId();
             }
 
-            // Start upload
+            // Start uploading the part.
             $parts[] = $this->doMultipartUpload($bucket, $key, $uploadId, $partNumber, $buffer);
             ++$partNumber;
             $buffer = \fopen('php://temp', 'rw+');
@@ -119,7 +119,7 @@ class SimpleS3Client extends S3Client
         }
 
         if (empty($parts)) {
-            // The upload did not contain any data.
+            // The upload did not contain any data. Let's create an empty file
             $this->doSmallFileUpload($options, $bucket, $key, '');
 
             return;
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -49,7 +49,7 @@ class SimpleS3Client extends S3Client
      */
     public function upload(string $bucket, string $key, $object, array $options = []): void
     {
-        // Send upload requests with size 32 MB
+        // Send upload requests with size 32MB
         $megabyte = 1024 * 1024;
         $stream = $this->getStream($object, 1 * $megabyte); // split the stream in 1MB chunk
 
@@ -59,11 +59,16 @@ class SimpleS3Client extends S3Client
             $contentLength = $stream->length();
         }
 
-        // The maximum number of parts is 10.000. Use 64Mb sized parts unless the content length is really large.
-        $partSize = $options['PartSize'] ?? max(64, null === $contentLength ? 0 : ceil($contentLength / (10000 * 1024 * 1024)));
+        /*
+         * The maximum number of parts is 10.000. The partSize must be a power of 2.
+         * We default this to 64MB per part. That means that we only support to upload
+         * files smaller than 64 * 10 000 = 640GB. If you are uploading larger files,
+         * please set PartSize to a higher number, like 128, 256 or 512. (Max 4096).
+         */
+        $partSize = $options['PartSize'] ?? 64;
         unset($options['PartSize']);
 
-        // If file is less than 64Mb, use normal upload
+        // If file is less than 64MB, use normal upload
         if (null !== $contentLength && $contentLength < 64 * $megabyte) {
             $this->doSmallFileUpload($options, $bucket, $key, $object);
 
--- a/src/Core/src/Stream/StreamFactory.php
+++ b/src/Core/src/Stream/StreamFactory.php
@@ -11,19 +11,19 @@ use AsyncAws\Core\Exception\InvalidArgument;
  */
 class StreamFactory
 {
-    public static function create($content): RequestStream
+    public static function create($content, int $preferredChunkSize = 64 * 1024): RequestStream
     {
         if (null === $content || \is_string($content)) {
             return StringStream::create($content ?? '');
         }
         if (\is_callable($content)) {
-            return CallableStream::create($content);
+            return CallableStream::create($content, $preferredChunkSize);
         }
         if (\is_iterable($content)) {
             return IterableStream::create($content);
         }
         if (\is_resource($content)) {
-            return ResourceStream::create($content);
+            return ResourceStream::create($content, $preferredChunkSize);
         }
 
         throw new InvalidArgument(sprintf('Unexpected content type "%s".', \is_object($content) ? \get_class($content) : \gettype($content)));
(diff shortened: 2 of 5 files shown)


src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +26,64 @@ public function getUrl(string $bucket, string $key): string
...
+
+            return;
+        }
+
+        if ($isResource) {
+            $stream = ResourceStream::create($object, $chunkSize);
+        } elseif (\is_callable($object)) {
+            $stream = CallableStream::create($object, $chunkSize);
+        } elseif (\is_iterable($object)) {
+            $stream = IterableStream::create($object);
```
Nyholm, line 66: Is it okey to use the streams here?

They are really helpful =)

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +28,69 @@ public function getUrl(string $bucket, string $key): string
...
+        } else {
+            throw new InvalidArgument('Third argument of SimpleS3Client::upload() must be string, resource, callable or an interable.');
+        }
+
+        $uploadId = $this->createMultipartUpload(['Bucket' => $bucket, 'Key' => $key])->getUploadId();
+        $parts = [];
+        $partNumber = 1;
+        foreach ($stream as $i => $chunk) {
```
jderusse, line 76: chunk must be greater than 5Mb. beware of callable. I suggest to wrap stream in `FixedSizeStream` to deal with that

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +28,69 @@ public function getUrl(string $bucket, string $key): string
...
+                'Bucket' => $bucket,
+                'Key' => $key,
+                'UploadId' => $uploadId,
+                'PartNumber' => $partNumber,
+                'Body' => $chunk,
+            ]);
+            $parts[] = new CompletedPart(['ETag' => $response->getETag(), 'PartNumber' => $partNumber]);
+            ++$partNumber;
+        }
+        echo \count($parts);
```
jderusse, line 87: debug?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +28,69 @@ public function getUrl(string $bucket, string $key): string
...
+            $stream = ResourceStream::create($object, $chunkSize);
+        } elseif (\is_callable($object)) {
+            $stream = CallableStream::create($object, $chunkSize);
+        } elseif (\is_iterable($object)) {
+            $stream = IterableStream::create($object);
+        } else {
+            throw new InvalidArgument('Third argument of SimpleS3Client::upload() must be string, resource, callable or an interable.');
+        }
+
+        $uploadId = $this->createMultipartUpload(['Bucket' => $bucket, 'Key' => $key])->getUploadId();
```
jderusse, line 73: should we delete this if upload failed on a part? or is it handled by AWS ?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +28,69 @@ public function getUrl(string $bucket, string $key): string
...
+
+    /**
+     * @param string|resource|callable|iterable $object
      */
+    public function upload(string $bucket, string $key, $object): void
+    {
+        // Send upload requests with size 64 MB
+        $chunkSize = 64 * 1024 * 1024;
+        $directUpload = \is_string($object);
+        if (!$directUpload && \is_resource($object)) {
```
jderusse, line 49: what's about creating stream first and use the `Stream::size` method then ?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
jderusse: Yeah they are really helpful in that context !

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +28,69 @@ public function getUrl(string $bucket, string $key): string
...
+            }
+        }
+
+        if ($directUpload) {
+            $this->putObject(['Bucket' => $bucket, 'Key' => $key, 'Body' => $object]);
+
+            return;
+        }
+
+        if (\is_resource($object)) {
```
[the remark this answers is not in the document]
Nyholm: I want to be able to define the chunk size. The default chunk size is 64Kb. AWS says that each part (chunk) must be at least 5Mb.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: =)

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: Maybe. I'll check

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: Because if the $object is an iterable or a callable, then we reading things into memory. This is for objects measured in gigabytes.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: Hm.. maybe.

Nyholm: Using FixedSizeStream will remove the users possibility to defined the chunks themselves. 

I prefer not to do that.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: Im now catching any errors and aborting the upload.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
jderusse: hmm, no it shouldn't. if yes it's a bug.
the stream return a size when available, otherwise return null

Nyholm: You are correct. Thank you

jderusse: note: streams have been designed to not store thing in memory (even hash method)

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +28,95 @@ public function getUrl(string $bucket, string $key): string
...
+     *   ACL?: \AsyncAws\S3\Enum\ObjectCannedACL::*,
+     *   CacheControl?: string,
+     *   ContentType?: string,
+     *   Metadata?: array<string, string>,
+     * } $options
      */
+    public function upload(string $bucket, string $key, $object, array $options = []): void
+    {
+        // Send upload requests with size 64 MB
+        $chunkSize = 64 * 1024 * 1024;
```
jderusse, line 53: hmmm we are gonna put 64M of string in memory. isn't it too much ? 

should we create a temporary file, read few MB of chunks, and upload the file when reaching 64Mb? (or more)

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: The code became massively simpler when Im using the FixedSizeStream. 👍

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: Im not sure if this is too much. The smaller this number is, the more requests are we doing. 

This is a "simple S3". If you want full control and optimize memory or requests, then you should use the `S3Client`. 

What is a good number for this?

Nyholm: I use 32Mb.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +30,97 @@ public function getUrl(string $bucket, string $key): string
...
+            'Bucket' => $bucket,
+            'Key' => $key,
+            'UploadId' => $uploadId,
+            'MultipartUpload' => new CompletedMultipartUpload(['Parts' => $parts]),
+        ]);
+    }
+
+    private function getStream($object, int $chunkSize): FixedSizeStream
+    {
+        if (null === $object || \is_string($object)) {
```
jderusse, line 112: Stream factory?

Nyholm: There are 2 differences between this function and the `StreamFactory`. 

1. This function sets the ChunkSize. 
1. This function returns a fixed stream

jderusse: the inner of this function can be simplified by using StreamFactory

Nyholm: I dont see how if we are going to respect the `$chunkSize`

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +30,104 @@ public function getUrl(string $bucket, string $key): string
...
+            $stream = ResourceStream::create($object, $chunkSize);
+        } elseif (\is_callable($object)) {
+            $stream = CallableStream::create($object, $chunkSize);
+        } elseif (\is_iterable($object)) {
+            $stream = IterableStream::create($object);
+        } else {
+            throw new InvalidArgument('Third argument of SimpleS3Client::upload() must be string, resource, callable or an interable.');
+        }
+
+        return FixedSizeStream::create($stream, $chunkSize);
```
jderusse, line 131:
```suggestion
        return new FixedSizeStream(StreamFactory::create($object), $chunkSize));
```

Nyholm: Yes, But you are missing this line: 

`$stream = CallableStream::create($object, $chunkSize);`

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +30,104 @@ public function getUrl(string $bucket, string $key): string
...
+        return $this->objectExists(['Bucket' => $bucket, 'Key' => $key])->isSuccess();
+    }
+
+    /**
+     * @param string|resource|callable|iterable $object
+     * @param array{
+     *   ACL?: \AsyncAws\S3\Enum\ObjectCannedACL::*,
+     *   CacheControl?: string,
+     *   ContentType?: string,
+     *   Metadata?: array<string, string>,
```
jderusse, line 49: We could add the `content-length` option and use it to get the `$contentLength` when the stream is a callable

Nyholm: That would just be a tiny optimisation. 

I also dont expect people that using a callable to know the exact length of the result.

Nyholm:
I could optimise this to send all uploads at once.. 

I should do that.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +30,104 @@ public function getUrl(string $bucket, string $key): string
...
+     * } $options
      */
+    public function upload(string $bucket, string $key, $object, array $options = []): void
+    {
+        // Send upload requests with size 32 MB
+        $megabyte = 1024 * 1024;
+        $stream = $this->getStream($object, 32 * $megabyte);
```
jderusse, line 56: what's about...
```php
$partSize = max(64, ceil($contentLength / (10000 * 1024 * 1024))); // number of chunk per part
$stream = $this->getStream($object, 1 * 1024 * 1024); // split the stream in 1MB chunk

$buffer = \fopen('php://temp', 'rw+');
$chunkIndex = 0;
foreach ($stream as$chunk) {
  if (null === $buffer) {
  }
  fwrite($buffer, $chunk);
  if (++$chunkIndex >= $partSize) {
   $response = $this->uploadPart(['Body' => $buffer]);
   $buffer = \fopen('php://temp', 'rw+');
   $chunkIndex = 0;
  }
}
```

jderusse: this will handle all cases (froms few MB to several TB) with less than 2Mb memory footprint.
Whithout too much complexity

Nyholm: Yes, that would take less memory. I'll implement something like it.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
jderusse: given the stream is wrapped in a `FixedStream` it will not be an issue. 
Maybe a performance issue, because the caller will try chunking the stream too much for nothing. 

My concernes is about dupplicate code (that does 99% the same thing) and maintanability (if we add a new Stream implementation, we'll have to update this code snippet at several places..

Maybe we should add a `int $preferedSize` in `Factory::create` to solve this

jderusse:
> I could optimise this to send all uploads at once..

hmm.. i'm not sure: 
- It mean that uloaded part will be stored in memory (unless you use temporary files)
- uploading hundred of part at the same time could be an issue (unless HttpClient manage a kind of pool to avoid that)
- the network bandwitch is the bottleneck, i'm not sure parallelizing it provide a huge benefit

Nyholm:
I've implemented the algoritm suggested. 

```
Upload 100Mb

## Baseline: 
Memory at start:    796 kB
Memory at end:    2,133 kB
Time consumed:   17,581 ms
Peak memory:     67,679 kB
Real memory:     69,640 kB


## With memory buffer
## With memory buffer
Memory at start:    797 kB
Memory at end:    2,134 kB
Time consumed:   10,070 ms
Peak memory:      5,216 kB
Real memory:      8,192 kB
```

Nyholm:
I will not upload then all simultaneously... I got the following error: 
> The list of parts was not in ascending order. Parts must be ordered by part number.

Anyhow. Thank you for the memory improvements.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: Added

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +26,130 @@ public function getUrl(string $bucket, string $key): string
...
+     *   Metadata?: array<string, string>,
+     * } $options
      */
+    public function upload(string $bucket, string $key, $object, array $options = []): void
+    {
+        // Send upload requests with size 32 MB
+        $megabyte = 1024 * 1024;
+
+        $contentLength = $options['ContentLength'] ?? null;
+        $chunksInPart = max(64, ceil($contentLength / (10000 * 1024 * 1024))); // number of chunk per part
```
Nyholm, line 55: Could you help me to understand this line?

Why `10000`?
This seams like a strange way to define how big the parts should be...

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: Added

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +26,137 @@ public function getUrl(string $bucket, string $key): string
...
+        if (isset($options['ContentLength'])) {
+            $ceil = ceil((int) $options['ContentLength'] / (10000 * 1024 * 1024));
+        }
+
+        $chunksInPart = max(64, $ceil);
+        $stream = $this->getStream($object, 1 * $megabyte); // split the stream in 1MB chunk
+
+        // If file is less than 64Mb, use normal upload
+        $contentLength = $options['ContentLength'] ?? $stream->length();
+        if (null !== $contentLength && $contentLength < 64 * $megabyte) {
```
jderusse, line 63: does it works if contentLength is a string?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +26,137 @@ public function getUrl(string $bucket, string $key): string
...
+     *   Metadata?: array<string, string>,
+     * } $options
      */
+    public function upload(string $bucket, string $key, $object, array $options = []): void
+    {
+        // Send upload requests with size 32 MB
+        $megabyte = 1024 * 1024;
+        $ceil = 0;
+        if (isset($options['ContentLength'])) {
+            $ceil = ceil((int) $options['ContentLength'] / (10000 * 1024 * 1024));
```
jderusse, line 55: what if people provide a stream resource without providing this option?
What's about using the stream defined at line 59 to get the size ?

`$size = $stream->size() ?? $option['contentLength'] ?? 0`

jderusse, line 55: hmmm maybe 10000 is a little but too much...

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
jderusse: you're right... 10000 is the maximum number of parts...
but that's too much..  we should find a better formula !

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: I’m not even sure how this works. :/

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: I've updated this. I added a configuration option. 

I think this is good.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: I respect ContentLength if it exists. 

```
        if (!empty($options['ContentLength'])) {
            $contentLength = (int) $options['ContentLength'];
        } else {
            $contentLength = $stream->length();
        }
```

Nyholm: I understand now. 

It increases the part size if the file is around 1TB to make sure we are using less than 10.000 parts.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: Updated.

Nyholm:
Im done with this PR now.

stof:
I'm wondering whether SimpleS3Client should extend the S3Client or decorate it...

Nyholm:
Good question. It will not be possible to implement getUrl if we use decoration. But yeah. It feels a bit strange.

jderusse:
I thought the same thing. 
Another advantage of composition is for people using the AsyncAwsBundle, it's easier to declare the SimpleS3 service and injecting an existing S3Client, than creating a SimpleS3Client with the proper configuration.

stof:
@Nyholm Well, in the laravel filesystem integration, you managed to implement it with decoration.

Nyholm:
Yeah. The thing I did in the Laravel filesystem felt like a hack too. 
That is removed now.

Nyholm:
Hm. If the SimpleS3Client would decorate an S3Client, Then for a given service that need to check if a bucket exists and then upload a file. 

That could be implemented with that service would need to use `S3Client` and `SimpleS3Client`. 

Or alternatively, it would create a SimpleS3Client in the constructor. 

The last option would be to provide a `SimpleS3Client::getS3Client()`

All 3 solutions would need the "presign URL hack" in the `getUrl()` method. 

--------

Anyhow. I take it as the API if okey and we should revisit the inheritance/composition question before tagging a stable release. 

I suggest to merge this PR in the current state.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +26,146 @@ public function getUrl(string $bucket, string $key): string
...
+     *   ContentType?: string,
+     *   Metadata?: array<string, string>,
+     *   PartSize?: int,
+     * } $options
      */
+    public function upload(string $bucket, string $key, $object, array $options = []): void
+    {
+        $megabyte = 1024 * 1024;
+        // Split the stream in 1MB chunk
+        $stream = $this->getStream($object, 1 * $megabyte);
```
bendavies, line 54: why `1 *`?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -21,8 +26,146 @@ public function getUrl(string $bucket, string $key): string
...
+
+        if (!empty($options['ContentLength'])) {
+            $contentLength = (int) $options['ContentLength'];
+        } else {
+            $contentLength = $stream->length();
```
bendavies, line 59: could be
```php
$contentLength = (int)($options['ContentLength'] ?? $stream->length());
```

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: I just wanted to show it was 1 megabyte.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
Nyholm: No, after your line, then $contentLength will be an int. 

`$stream->length()` returns `?int`

PR #654, opened 2020-05-26
Fix SES signature

jderusse:
When `endpoints.json` does not provide a ScopedService, we fallback to the `api2.json` metadata

### DIFF
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## NOT RELEASED
 
-### Fixed
-
-- Fix invalid signature in SES client because of wrong Scoped Service.
-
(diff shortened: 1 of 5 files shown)


src/Core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
```
jderusse, line 3: I think this bug is serious enougth to tag a `1.2.1` release as soon as merged. 
WDYT

Nyholm: Yes!

jderusse: that were the wrong CHANGELOG
Updated SES's CHANGELOG

PR #656, opened 2020-05-26
Get more context with error logs

Nyholm:
This will help debugging. 

We are also following PSR-3 by using the `exception` key.


Nyholm:
Thank you

PR #661, opened 2020-05-28
Implement CloudFront

### LINKED ISSUE #526: [Feature Request] - Cloudfront  API
- [ ] cache invalidation

Nyholm:
This will fix #526


src/Service/CloudFront/composer.json
```diff
@@ -0,0 +1,33 @@
...
+        "aws",
+        "amazon",
+        "sdk",
+        "async-aws",
+        "cloud-front"
+    ],
+    "license": "MIT",
+    "require": {
+        "php": "^7.2.5",
+        "ext-json": "*",
```
jderusse, line 15: it looks like it don't use json, but xml

src/Service/CloudFront/tests/Unit/Input/CreateInvalidationRequestTest.php
```diff
@@ -0,0 +1,45 @@
...
+            'InvalidationBatch' => new InvalidationBatch([
+                'Paths' => new Paths([
+                    'Quantity' => 1337,
+                    'Items' => ['all/images'],
+                ]),
+                'CallerReference' => 'abc',
+            ]),
+        ]);
+
+        // see https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_CreateInvalidation2019_03_26.html
```
jderusse, line 26: URL does not exists

src/Service/CloudFront/composer.json
Nyholm: Thank yuo

src/Service/CloudFront/tests/Unit/Input/CreateInvalidationRequestTest.php
Nyholm: It came from the generated code.
I removed it.

src/Service/CloudFront/src/CloudFrontClient.php
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @see https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-cloudfront-2019-03-26.html#createinvalidation2019_03_26
+     *
+     * @param array{
+     *   DistributionId: string,
+     *   InvalidationBatch: \AsyncAws\CloudFront\ValueObject\InvalidationBatch|array,
+     *   @region?: string,
+     * }|CreateInvalidationRequest $input
+     */
+    public function createInvalidation2019_03_26($input): CreateInvalidationResult
```
Nyholm, line 24: Hm.. Should we improve on this name?

src/Service/CloudFront/src/ValueObject/Paths.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @param array{
+     *   Quantity: int,
+     *   Items?: null|string[],
+     * } $input
+     */
+    public function __construct(array $input)
+    {
+        $this->Quantity = $input['Quantity'] ?? null;
```
Nyholm, line 27: Is this correct? Quantity is required.

Nyholm: I think this was okey at the time we had setters at the value objects.

Nyholm: I think this is okey. You get a nice error message when this value object is being converted to a request.

src/Service/CloudFront/src/CloudFrontClient.php
Nyholm: Hm. Maybe not. This is the name they decided.

Nyholm: But they are ugly... Hm... 

I need some input =)

src/Service/CloudFront/src/ValueObject/Paths.php
jderusse: yes. people are free to construct their object with constructor or setter.

the required field are checked when building the request

src/Service/CloudFront/src/CloudFrontClient.php
Nyholm: Let's discuss this in next PR.

jderusse: Looking at `api2.json`, the model is:

```
    "CreateInvalidation":{
      "name":"CreateInvalidation2019_03_26",
      "http":{
        "method":"POST",
        "requestUri":"/2019-03-26/distribution/{DistributionId}/invalidation",
        "responseCode":201
      },
      "input":{"shape":"CreateInvalidationRequest"},
      "output":{"shape":"CreateInvalidationResult"},
```

meaning that the operation is "keyed" `CreateInvalidation` but for unknown reason it name is `CreateInvalidation2019_03_26`.

The official AwsSdk uses `CreateInvalidation`. And I trully think we should do the same.
by ignoring the `"name"` attribute.

Nyholm: I fixed this in https://github.com/async-aws/aws/pull/675

I think "name" is the name of the operation. I use the "key" as name of the method.

PR #662, opened 2020-05-28
Fix dynamob DB create table example

tyx:


jderusse:
Good catch, thank you

PR #668, opened 2020-06-03
Differentiate empty list from no list

### LINKED ISSUE #667: [DynamobDB] Deal with empty List
Hello !

Currently it seems impossible to send an empty list.

Here is my operation :
```
 $this->dynamoDbClient->updateItem([
                'TableName' => 'reservations_tenant',
                'Key' => [
                    'user_id' => ['S' => 'myUser'],
                ],
                'UpdateExpression' => 'SET reservation_ids = list_append(if_not_exists(reservation_ids, :emptyList), :newReservationId)',
                'ExpressionAttributeValues' => [
                    ':newReservationId' => ['L' => [['S' => $tenant->reservationId()]]],
                    ':emptyList' => ['L' => []],
                ],
            ])
```

But it sends : 
```
-body: AsyncAws\Core\Stream\StringStream {#2154
           -content: "{"TableName":"reservations_tenant","Key":{"user_id":{"S":"lala"}},"UpdateExpression":"SET reservation_ids = list_append(if_not_exists(reservation_ids, :emptyList), :newReservationId)","ExpressionAttributeValues":{":newReservationId":{"L":[{"S":"TJFXHZI-4"}]},":emptyList":[]}}"
           -lengthCache: null
         }
```

Where it should send:
`"emptyList": ["L" => []]`

Without i get from AWS:
```
SerializationException
Message: Unrecognized collection type class com.amazonaws.dynamodb.v20120810.AttributeValue
```

I am missing something ?

Thanks

tyx:
Try to fix #667

### DIFF
--- a/src/CodeGenerator/src/Generator/InputGenerator.php
+++ b/src/CodeGenerator/src/Generator/InputGenerator.php
@@ -105,33 +105,36 @@ class InputGenerator
             foreach ($memberClassNames as $memberClassName) {
                 $namespace->addUse($memberClassName->getFqdn());
             }
-            $nullable = true;
+            $getterSetterNullable = true;
+
             if ($memberShape instanceof StructureShape) {
                 $memberClassName = $this->objectGenerator->generate($memberShape);
                 $constructorBody .= strtr('$this->NAME = isset($input["NAME"]) ? CLASS::create($input["NAME"]) : null;' . "\n", ['NAME' => $member->getName(), 'CLASS' => $memberClassName->getName()]);
             } elseif ($memberShape instanceof ListShape) {
                 $listMemberShape = $memberShape->getMember()->getShape();
-                $nullable = false;
+                $getterSetterNullable = false;
 
                 if ($listMemberShape instanceof StructureShape) {
+                    $getterSetterNullable = false;
                     $memberClassName = $this->objectGenerator->generate($listMemberShape);
-                    $constructorBody .= strtr('$this->NAME = array_map([CLASS::class, "create"], $input["NAME"] ?? []);' . "\n", ['NAME' => $member->getName(), 'CLASS' => $memberClassName->getName()]);
+                    $constructorBody .= strtr('$this->NAME = isset($input["NAME"]) ? array_map([CLASS::class, "create"], $input["NAME"]) : null;' . "\n", ['NAME' => $member->getName(), 'CLASS' => $memberClassName->getName()]);
                 } elseif ($listMemberShape instanceof ListShape) {
+                    $getterSetterNullable = false;
                     $listMemberShapelevel2 = $listMemberShape->getMember()->getShape();
                     if ($listMemberShapelevel2 instanceof StructureShape) {
                         $memberClassName = $this->objectGenerator->generate($listMemberShapelevel2);
-                        $constructorBody .= strtr('$this->NAME = array_map(static function(array $array) {
+                        $constructorBody .= strtr('$this->NAME = isset($input["NAME"]) ? array_map(static function(array $array) {
                             return array_map([CLASS::class, "create"], $array);
-                        }, $input["NAME"] ?? []);' . "\n", ['NAME' => $member->getName(), 'CLASS' => $memberClassName->getName()]);
+                        }, $input["NAME"]) : null;' . "\n", ['NAME' => $member->getName(), 'CLASS' => $memberClassName->getName()]);
                     } elseif ($listMemberShapelevel2 instanceof ListShape || $listMemberShapelevel2 instanceof MapShape) {
                         throw new \RuntimeException('Recursive ListShape are not yet implemented');
                     } else {
-                        $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? [];' . "\n", ['NAME' => $member->getName()]);
+                        $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? null;' . "\n", ['NAME' => $member->getName()]);
                     }
                 } elseif ($listMemberShape instanceof MapShape) {
                     throw new \RuntimeException('Recursive ListShape are not yet implemented');
                 } else {
-                    $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? [];' . "\n", ['NAME' => $member->getName()]);
+                    $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? null;' . "\n", ['NAME' => $member->getName()]);
                 }
             } elseif ($memberShape instanceof MapShape) {
                 $mapKeyShape = $memberShape->getKey()->getShape();
@@ -139,8 +142,7 @@ class InputGenerator
                     $this->enumGenerator->generate($mapKeyShape);
                 }
                 $mapValueShape = $memberShape->getValue()->getShape();
-                $nullable = false;
-
+                $getterSetterNullable = false;
                 // Is this a list of objects?
                 if ($mapValueShape instanceof StructureShape) {
                     $memberClassName = $this->objectGenerator->generate($mapValueShape);
@@ -175,7 +177,7 @@ class InputGenerator
                     throw new \RuntimeException('Recursive MapShape are not yet implemented');
                 } else {
                     // It is a scalar, like a string
-                    $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? [];' . "\n", ['NAME' => $member->getName()]);
+                    $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? null;' . "\n", ['NAME' => $member->getName()]);
                 }
             } elseif ($member->isStreaming()) {
                 $parameterType = 'string|resource|callable|iterable';
@@ -199,12 +201,13 @@ class InputGenerator
             if ($member->isRequired()) {
                 $property->addComment('@required');
             }
+
             // the "\n" helps php-cs-fixer to with potential wildcard in parameterType
-            $property->addComment("\n@var " . ($nullable ? 'null|' : '') . $parameterType);
+            $property->addComment("\n@var null|$parameterType");
 
             $getter = $class->addMethod('get' . \ucfirst($member->getName()))
                 ->setReturnType($returnType)
-                ->setReturnNullable($nullable);
+                ->setReturnNullable($getterSetterNullable);
             $setter = $class->addMethod('set' . \ucfirst($member->getName()))
                 ->setReturnType('self');
 
@@ -214,7 +217,11 @@ class InputGenerator
                 $setter->addComment('@deprecated');
                 $deprecation = strtr('@trigger_error(\sprintf(\'The property "NAME" of "%s" is deprecated by AWS.\', __CLASS__), E_USER_DEPRECATED);', ['NAME' => $member->getName()]);
             }
-            $getter->setBody($deprecation . strtr('return $this->NAME;', ['NAME' => $member->getName()]));
+            if ($getterSetterNullable) {
+                $getter->setBody($deprecation . strtr('return $this->NAME;', ['NAME' => $member->getName()]));
+            } else {
+                $getter->setBody($deprecation . strtr('return $this->NAME ?? [];', ['NAME' => $member->getName()]));
+            }
             $setter->setBody($deprecation . strtr('
                     $this->NAME = $value;
                     return $this;
@@ -222,12 +229,12 @@ class InputGenerator
                 'NAME' => $member->getName(),
             ]));
             $setter
-                ->addParameter('value')->setType($returnType)->setNullable($nullable)
+                ->addParameter('value')->setType($returnType)->setNullable($getterSetterNullable)
             ;
 
             if ($returnType !== $parameterType) {
-                $setter->addComment('@param ' . $parameterType . ($nullable ? '|null' : '') . ' $value');
-                $getter->addComment('@return ' . $parameterType . ($nullable ? '|null' : ''));
+                $setter->addComment('@param ' . $parameterType . ($getterSetterNullable ? '|null' : '') . ' $value');
+                $getter->addComment('@return ' . $parameterType . ($getterSetterNullable ? '|null' : ''));
             }
         }
 
@@ -344,8 +351,10 @@ class InputGenerator
                 throw new \InvalidArgumentException('Headers does not yet support Enum in value');
             }
 
-            $bodyCode = strtr('foreach (VALUE as $key => $value) {
-                $headers["LOCATION$key"] = $value;
+            $bodyCode = strtr('if (null !== VALUE) {
+                foreach (VALUE as $key => $value) {
+                    $headers["LOCATION$key"] = $value;
+                }
             }', [
                 'PROPERTY' => $member->getName(),
                 'LOCATION' => $member->getLocationName() ?? $member->getName(),
--- a/src/CodeGenerator/src/Generator/ObjectGenerator.php
+++ b/src/CodeGenerator/src/Generator/ObjectGenerator.php
@@ -163,18 +163,18 @@ class ObjectGenerator
                 // Check if this is a list of objects
                 if ($listMemberShape instanceof StructureShape) {
                     $objectClass = $this->generate($listMemberShape);
-                    $constructorBody .= strtr('$this->NAME = array_map([CLASS::class, "create"], $input["NAME"] ?? []);' . "\n", ['NAME' => $member->getName(), 'CLASS' => $objectClass->getName()]);
+                    $constructorBody .= strtr('$this->NAME = isset($input["NAME"]) ? array_map([CLASS::class, "create"], $input["NAME"]) : null;' . "\n", ['NAME' => $member->getName(), 'CLASS' => $objectClass->getName()]);
                 } else {
-                    $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? [];' . "\n", ['NAME' => $member->getName()]);
+                    $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? null;' . "\n", ['NAME' => $member->getName()]);
                 }
             } elseif ($memberShape instanceof MapShape) {
                 $mapValueShape = $memberShape->getValue()->getShape();
 
                 if ($mapValueShape instanceof StructureShape) {
                     $objectClass = $this->generate($mapValueShape);
-                    $constructorBody .= strtr('$this->NAME = array_map([CLASS::class, "create"], $input["NAME"] ?? []);' . "\n", ['NAME' => $member->getName(), 'CLASS' => $objectClass->getName()]);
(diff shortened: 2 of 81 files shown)


Nyholm:
Thank you. 

I think #667 is an issue we should solve. However, note that the clients are generated. Which means that we cannot edit this code directly. We need to modify the code generator. 

https://async-aws.com/contribute/

tyx:
Yes of course ! I just wanted to show you quickly my suggestion before diving in code generator 😅

src/Service/DynamoDb/src/ValueObject/AttributeValue.php
```diff
@@ -77,7 +77,7 @@ public function __construct(array $input)
         $this->NS = $input['NS'] ?? [];
         $this->BS = $input['BS'] ?? [];
         $this->M = array_map([AttributeValue::class, 'create'], $input['M'] ?? []);
-        $this->L = array_map([AttributeValue::class, 'create'], $input['L'] ?? []);
+        $this->L = array_key_exists('L', $input) ? array_map([AttributeValue::class, 'create'], $input['L']) : null;
```
jderusse, line 80: should be isset or you should handle null value

src/Service/DynamoDb/src/ValueObject/AttributeValue.php
```diff
@@ -110,7 +110,7 @@ public function getBS(): array
      */
     public function getL(): array
     {
-        return $this->L;
+        return $this->L ?? [];
```
jderusse, line 113: return type should be ?array as other optional option

Nyholm:
> but that will be a BC break for getters.

But what getters are we talking about? I believe `$s3->listFiles()` should be an empty array an not null. But in this scenario with DynamoDb it should be null. Maybe it is a way to define/differentiate those. 

Im fine with BC breaks in DynamoDB btw.

jderusse:
I'm talking about all getters on Input and ValueObjects:

If we change it this in generator, it will impact all services.

```diff
 $i = new SendEmailRequest([]);
-var_dump($i->ReplyToAddresses()): // array()
+var_dump($i->ReplyToAddresses()): // null
```

tyx:
Hi !

I updated the PR with CodeGenerator changes.

I tried to use the generated code on my project and it works for my usecase.

I just failled to change the getter return type to `array|null` if one of you could help !

Let me know if anything could be improved

src/Service/CloudFormation/src/ValueObject/RollbackConfiguration.php
```diff
@@ -23,7 +23,7 @@ final class RollbackConfiguration
      */
     public function __construct(array $input)
     {
-        $this->RollbackTriggers = array_map([RollbackTrigger::class, 'create'], $input['RollbackTriggers'] ?? []);
+        $this->RollbackTriggers = isset($input['RollbackTriggers']) ? array_map([RollbackTrigger::class, 'create'], $input['RollbackTriggers']) : null;
```
Nyholm, line 26: Hm.. 

```php
$x = new RollbackConfiguration();
$x->getRollbackTriggers(); // will throw type error
```

Nyholm:
Am I correct when I say that it is only DynamoDb that need to make a difference between null and empty list?

If so, could we add a special feature flag in manifest.json?

tyx:
I should admit I really don't know.

jderusse:
> Am I correct when I say that it is only DynamoDb that need to make a difference between null and empty list?

I don't think so. I guess we are lucky because in other services AWS (or we) converts null into empty array.

If we want keeping BC, we may fallback to empty array in getter like @tyx did in the first implementation

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
```diff
@@ -171,18 +171,32 @@ private function dumpArrayList(string $output, string $input, string $contextPro
         $memberShape = $shape->getMember()->getShape();
 
         return strtr('
-            $index = -1;
-            foreach (INPUT as $listValue) {
-                $index++;
-                MEMBER_CODE
+            if (null !== INPUT) {
```
jderusse, line 174: This not needed (or not the right implementation). None of other `dump*` method performs such check because, this should have been done before.

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
```diff
@@ -171,18 +171,32 @@ private function dumpArrayList(string $output, string $input, string $contextPro
...
+                $index = -1;
+                INIT_CODE
```
jderusse, line 176: hmm.  This is alway `$payload = [];` right? no need for an other extra method

src/Service/CloudFormation/src/ValueObject/RollbackConfiguration.php
jderusse: indeed. We should either change the return typehint of getter (BC break) 
Or fallback to empty array

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
jderusse: Got it, you should remove at line 62
```diff
-if ($shape instanceof ListShape || $shape instanceof MapShape) {
+if ($shape instanceof MapShape) {
   $body = 'MEMBER_CODE';
   $inputElement = '$this->' . $member->getName();
 }

File: src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php [hunk 6]
```diff
@@ -171,18 +171,32 @@ private function dumpArrayList(string $output, string $input, string $contextPro
...
                 'INPUT' => $input,
+                'INIT_CODE' => $this->dumpArrayInit($output, $memberShape),
                 'MEMBER_CODE' => $memberCode = $this->dumpArrayElement(sprintf('%s[$index]', $output), '$listValue', $contextProperty, $memberShape, true),
             ]);
     }
 
+    private function dumpArrayInit(string $output, Shape $shape)
+    {
+        switch (true) {
+            case $shape instanceof StructureShape && 'AttributeValue' === $shape->getName():
```
jderusse, line 193: remove specific code `AttributeValue`

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
tyx: I created the other method to ensure not doing this init everywhere as it break many tests.

jderusse: what if the test were wrong before?..

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
tyx: Same here. I restricted to reduce side effect and broken test.

I will revert it to let you have a look on broken test?

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
tyx: Many are correct, for example. in DynamoDB, only List could be sent empty. Not Set (like SS, NS or BS).

tyx:
> > Am I correct when I say that it is only DynamoDb that need to make a difference between null and empty list?
> 
> I don't think so. I guess we are lucky because in other services AWS (or we) converts null into empty array.

And currently, empty array are never sent. And I have no idea what it would mean to send all of them. At least in DynamoDB, only `List` could be sent empty, not `Set` (I don't know for `Map`)

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
tyx: We could introduce a specific Shape like SetShape and keep ListShape for real list ? To allow making things differently

jderusse:
> And currently, empty array are never sent

Thats should still be the case, unless people creates objects with explicit empty array, the arrays should be initialized with null, not beeing sent.

tyx:
Ok you're right. I tried to not change everything, but I guess I have no choice.

Here is the PR updated and tests are green (in local).

Still the getter issue I need to update

tyx:
Ok, here it is !

I found how to handle null value.

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
```diff
@@ -172,17 +169,29 @@ private function dumpArrayList(string $output, string $input, string $contextPro
...
+                'INIT_CODE' => $this->dumpArrayInit($output, $memberShape),
                 'MEMBER_CODE' => $memberCode = $this->dumpArrayElement(sprintf('%s[$index]', $output), '$listValue', $contextProperty, $memberShape, true),
             ]);
     }
 
+    private function dumpArrayInit(string $output, Shape $shape)
+    {
+        switch (true) {
+            case $shape instanceof StructureShape:
+                return strtr('$payloadOUTPUT = [];', [
```
jderusse, line 189: I style don't get why we initialize the array when children is a structure, and we don't do it for map, list, scalar...

src/CodeGenerator/src/Generator/InputGenerator.php
```diff
@@ -111,23 +111,21 @@ public function generate(Operation $operation): ClassName
...
-                    $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? [];' . "\n", ['NAME' => $member->getName()]);
+                    $constructorBody .= strtr('$this->NAME = $input["NAME"] ?? null;' . "\n", ['NAME' => $member->getName()]);
                 }
             } elseif ($memberShape instanceof MapShape) {
                 $mapKeyShape = $memberShape->getKey()->getShape();
                 if (!empty($mapKeyShape->getEnum())) {
                     $this->enumGenerator->generate($mapKeyShape);
                 }
                 $mapValueShape = $memberShape->getValue()->getShape();
-                $nullable = false;
```
jderusse, line 130: I wonder if it make sens to also deal with Map

src/Core/src/Sts/Input/AssumeRoleRequest.php
```diff
@@ -163,9 +163,9 @@ public function getPolicy(): ?string
     }
 
     /**
-     * @return PolicyDescriptorType[]
+     * @return PolicyDescriptorType[]|null
      */
-    public function getPolicyArns(): array
+    public function getPolicyArns(): ?array
```
jderusse, line 168: sadly that's a BC break :(

src/Service/CloudWatchLogs/src/Input/PutLogEventsRequest.php
```diff
@@ -162,9 +162,13 @@ private function requestBody(): array
             throw new InvalidArgument(sprintf('Missing parameter "logStreamName" for "%s". The value cannot be null.', __CLASS__));
         }
         $payload['logStreamName'] = $v;
+        if (null === $v = $this->logEvents) {
+            throw new InvalidArgument(sprintf('Missing parameter "logEvents" for "%s". The value cannot be null.', __CLASS__));
```
jderusse, line 166: Is it a BC break?.. The field is required, but previously AsyncAws replaced null by empty array.

src/Service/DynamoDb/tests/Integration/DynamoDbClientTest.php
```diff
@@ -262,6 +262,28 @@ public function testUpdateItem(): void
         self::assertEquals(200, $result->info()['status']);
     }
 
+    public function testUpdateItemWithList(): void
```
jderusse, line 265: thanks for this!!

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
tyx: After latest change, it could be useless now, I will give it a try !

src/Core/src/Sts/Input/AssumeRoleRequest.php
tyx: Yes for now, I don't implement the first solution to avoid BC break. But it stills possible

src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
tyx: Ok good catch, I just removed this stuff

Nyholm:
I think the easiest way to move forward is to add feature flags in the manifest json. 

So for each API we can toggle if we should do `?array` or not. So for our 4 APIs that are tagged with 1.0, we keep the old behavior, but we can update DynamoDb (and others if we want) with this new behavior.

tyx:
Like this ?
```json
{
    "DynamoDb": {
            "features": {
                "empty_array": true
            }
    }
}
```

src/Core/src/Sts/Input/AssumeRoleRequest.php
```diff
@@ -186,9 +186,9 @@ public function getSerialNumber(): ?string
     }
 
     /**
-     * @return Tag[]
+     * @return Tag[]|null
      */
-    public function getTags(): array
+    public function getTags(): ?array
```
jderusse, line 191: At first I thought that `?array` was the right way to handle getters. It's consistent with all other getter.

But in that precise example, I think that'll reduce DX. leading to
```
if ($request->getTags() !== null) {
  foreach ($request->getTags() as $tag) {}
}

// or
foreach ($request->getTags() ?? [] as $tag) {}
```

I think we can safetly always fallback to `[]` , and change the signature of this getter to `array`.
If one day we change our mind, we'll be able to add a deprecation notice and change the behavior.

In DynamoDb, is it important to distinguished null from empty array (the original bug)?
- if yes, we should go with feature flag as suggested by @Nyholm (only on really required properties)
- if no, we just fallback to ` ?? []` everywhere

jderusse:
More something like this
```
{
    "DynamoDb": {
            "features": {
                "empty_array": [
                   'AttributeValue.NS', // ShapeName "." MemberName
                ]
            }
    }
}
```

src/Core/src/Sts/Input/AssumeRoleRequest.php
tyx: > In DynamoDb, is it important to distinguished null from empty array (the original bug)?

from what I've seen of it, it's not a problem. Getter can continue to return empty array.

Nyholm:
Yeah. Or `s/empty_array/nullable_array/`

src/Core/src/Sts/Input/AssumeRoleRequest.php
jderusse: great, feature flag behavior is not (yet) needed. internal call distinguish null from empty, but external signature always return array.

what do you think @Nyholm

Nyholm: Lets try to use the fallback and we will see =)

tyx: Just updated the PR with the fallback to array.

I still have some fix needed on 3 S3 operations but main modification is here.

tyx:
phpstan failure seems unrelated

src/CodeGenerator/src/Generator/InputGenerator.php
```diff
@@ -182,12 +185,13 @@ public function generate(Operation $operation): ClassName
             if ($member->isRequired()) {
                 $property->addComment('@required');
             }
+
             // the "\n" helps php-cs-fixer to with potential wildcard in parameterType
-            $property->addComment("\n@var " . ($nullable ? 'null|' : '') . $parameterType);
+            $property->addComment("\n@var " . ($propertyNullable ? 'null|' : '') . $parameterType);
```
jderusse, line 190: looks like `$propertyNullable` is always true

src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -254,13 +253,21 @@ private function addProperties(StructureShape $shape, ClassType $class, PhpNames
...
+            } else {
+                $method->setBody($deprecation . strtr('
+                        return $this->NAME ?? [];
+                    ', [
+                    'NAME' => $member->getName(),
+                ]));
+            }
 
-            $nullable = $nullable ?? !$member->isRequired();
+            $nullable = ($nullable ?? !$member->isRequired()) && $getterSetterNullable;
```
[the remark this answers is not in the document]
tyx: we still have one nullable use : https://github.com/async-aws/aws/blob/master/src/CodeGenerator/src/Generator/ObjectGenerator.php#L245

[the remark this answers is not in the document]
tyx: in fact we still need to distinguish List/Map nullable (to use Null Coalesce in getter) from current nullable (to indicate null in return phpdoc).

Or I miss something ? (monday morning so it could be possible :D)

.github/workflows/static.yml
```diff
@@ -20,7 +20,9 @@ jobs:
           composer update --no-interaction --prefer-dist --optimize-autoloader
 
       - name: PHPStan
-        uses: OskarStark/phpstan-ga@0.12.23
+        uses: OskarStark/phpstan-ga@latest
```
shouze, line 23: :crossed_fingers:

shouze: related to https://github.com/OskarStark/phpstan-ga/pull/29

src/CodeGenerator/src/Generator/ObjectGenerator.php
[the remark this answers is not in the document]
tyx: To make is easier to read I could set nullable and getterSetterNullable in List/Map block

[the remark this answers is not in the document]
tyx: I updated with my latest proposal

tyx:
rebased to get phpstan fix ;)

tyx:
I'm done here !

Let me know if anything need to be changed.

src/CodeGenerator/src/Generator/RequestSerializer/QuerySerializer.php
```diff
@@ -68,9 +68,6 @@ public function generateRequestBuilder(StructureShape $shape): array
                 }
                 MEMBER_CODE';
                 $inputElement = '$v';
-            } elseif ($shape instanceof ListShape || $shape instanceof MapShape) {
-                $body = 'MEMBER_CODE';
-                $inputElement = '$this->' . $member->getName();
```
Nyholm, line 73: Can you tell me why this was removed?

tyx: Because as List and Map could be nullable now, we need to go through else statement just below. (to check !== null)

Nyholm: Ah, Thank you

PR #670, opened 2020-06-04
Fix metadata not beeing sent to AWS

### LINKED ISSUE #669: Metadata is not added to pre-signed PutObjectRequest
Doesn't seem to support the addition of Metadata.

This is my code below but the result doesn't add the correct headers to the pre-signed request.
```
        $input = new PutObjectRequest([
            'Bucket' => $this->configuration->getExerciseVideoBucket()
            'Key' => $remoteFilename,
            'Metadata' => [
                "x-amz-meta-filename: $filename"
            ]
        ]);

       $url = $this->s3->presign($input, new \DateTimeImmutable('+20 min'));

```

jderusse:
fixes #669

### DIFF
--- a/src/Service/S3/src/Result/GetObjectOutput.php
+++ b/src/Service/S3/src/Result/GetObjectOutput.php
@@ -454,7 +454,7 @@ class GetObjectOutput extends Result
         $this->Metadata = [];
         foreach ($headers as $name => $value) {
             if ('x-amz-meta-' === substr($name, 0, 11)) {
-                $this->Metadata[$name] = $value[0];
+                $this->Metadata[substr($name, 11)] = $value[0];
             }
         }
 
--- a/src/Service/S3/tests/Unit/Result/GetObjectOutputTest.php
+++ b/src/Service/S3/tests/Unit/Result/GetObjectOutputTest.php
@@ -57,7 +57,8 @@ class GetObjectOutputTest extends TestCase
 
         $metadata = $result->getMetadata();
         self::assertCount(1, $metadata);
-        self::assertArrayHasKey('x-amz-meta-tobias', $metadata);
-        self::assertEquals('nyholm', $metadata['x-amz-meta-tobias']);
+        self::assertArrayNotHasKey('x-amz-meta-tobias', $metadata);
+        self::assertArrayHasKey('tobias', $metadata);
+        self::assertEquals('nyholm', $metadata['tobias']);
     }
 }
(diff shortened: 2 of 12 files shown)


src/Service/S3/tests/Unit/Result/GetObjectOutputTest.php
```diff
@@ -57,7 +57,8 @@ public function testMetadata()
 
         $metadata = $result->getMetadata();
         self::assertCount(1, $metadata);
-        self::assertArrayHasKey('x-amz-meta-tobias', $metadata);
-        self::assertEquals('nyholm', $metadata['x-amz-meta-tobias']);
+        self::assertArrayNotHasKey('x-amz-meta-tobias', $metadata);
+        self::assertArrayHasKey('tobias', $metadata);
```
jderusse, line 61: This change mimics AWS SDK

IMHO `x-amz-meta-` is an internal prefix and shoudn't have leaked to end user. When people use the official SDK to send `tobias`, they expect to get `tobias` in return.

Nyholm: Yes, True. I missed this before. 

Thank you

src/Service/S3/src/Result/GetObjectOutput.php
```diff
@@ -454,7 +454,7 @@ protected function populateResult(Response $response): void
         $this->Metadata = [];
         foreach ($headers as $name => $value) {
             if ('x-amz-meta-' === substr($name, 0, 11)) {
-                $this->Metadata[$name] = $value[0];
+                $this->Metadata[substr($name, 11)] = $value[0];
```
Nyholm, line 457: Could the user manually add metadata in the S3 dashboard. Would that be prefixed here?

jderusse: I'm not sure to understand what you mean. 

note: Aws SDK removes the prefix as well

Nyholm: I meant: Will `$name` always include the prefix? No matter how the metadata was added?

jderusse: prefix it checked at line 456

Nyholm: Oh, sorry. I didnt see that.

Nyholm:
Thank you

phamviet:
Thanks for this PR. May i know when this get released? Currently i have to use 1.2-dev to get this fix but would be good to have it stable instead.

Nyholm:
Thank you for the ping. We've must have forgotten this. I've prepared a release here: https://github.com/async-aws/aws/pull/734

phamviet:
Thanks @Nyholm! Did not expect it happens so quickly. Very appreciate your work!

PR #671, opened 2020-06-05
Add support for PHP 8

jderusse:


composer.json
```diff
@@ -4,7 +4,7 @@
     "description": "Root repository cannot be installed with composer. Use subpackages like async-aws/s3",
     "license": "MIT",
     "require": {
-        "php": "^7.2.5",
+        "php": "^7.2.5 || ^8.0",
```
Nyholm, line 7: We should also do this on all other packages.

jderusse: :facepalm: 

good cath. thanks

PR #673, opened 2020-06-06
Added tests for #654

### LINKED ISSUE #655: Add tests for #654
Let's make sure we dont introduce regressions in the future.

Nyholm:
This will fix #655


Nyholm:
Im just testing that the generator is providing the correct value here. 

Im not testing what value is actually being sent. I could try to do an integration test. 

We have manually tested this and know that this is the correct value. However, this will break if the API server changes their requirement. But I think that is fair.

PR #675, opened 2020-06-06
[CloudFront] Update method name

Nyholm:
This generator update will just effect the PHP method name, not the operation name.


Nyholm:
Thank you

PR #677, opened 2020-06-07
[DynamoDb] Adding BatchGetItem

Nyholm:
After @Nemo64's work on recursion, I though I could give this a try. 

There are some issues currently with the generator. I'll look into fix them. 

Related to #566

PR #678, opened 2020-06-07
Remove reference $fn in PopulateResult

jderusse:
This PR replace the `$fn` array (passed by reference) by private methods.

The code is more readable, and recursion is easier to handle.

I wonder if I shouldn't do the same for XmlParser: currently closures are not static and declared in loops, I beleive their is a potential performance optimization by replacing theme by real methods

### DIFF
--- a/src/CodeGenerator/src/Generator/ResponseParser/ParserResult.php
+++ b/src/CodeGenerator/src/Generator/ResponseParser/ParserResult.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace AsyncAws\CodeGenerator\Generator\ResponseParser;
 
+use Nette\PhpGenerator\Method;
+
 /**
  * @internal
  */
@@ -18,10 +20,16 @@ class ParserResult
      */
     private $usedClasses;
 
-    public function __construct(string $body, array $usedClasses)
+    /**
+     * @var Method[]
+     */
+    private $extraMethods;
+
+    public function __construct(string $body, array $usedClasses = [], array $extraMethods = [])
     {
         $this->body = $body;
         $this->usedClasses = $usedClasses;
+        $this->extraMethods = $extraMethods;
     }
 
     public function getBody(): string
@@ -32,5 +40,10 @@ class ParserResult
     public function getUsedClasses(): array
     {
         return $this->usedClasses;
+    }
+
+    public function getExtraMethods(): array
+    {
+        return $this->extraMethods;
     }
 }
--- a/src/CodeGenerator/src/Generator/ResponseParser/ParserResult.php
+++ b/src/CodeGenerator/src/Generator/ResponseParser/ParserResult.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace AsyncAws\CodeGenerator\Generator\ResponseParser;
 
+use AsyncAws\CodeGenerator\Generator\Naming\ClassName;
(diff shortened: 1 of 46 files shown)


src/CodeGenerator/src/Generator/ResponseParser/ParserResult.php
```diff
@@ -0,0 +1,39 @@
...
+     * @var string
+     */
+    private $body;
+
+    /**
+     * @var Method[]
+     */
+    private $extraMethods;
+
+    public function __construct(string $body, array $extraMethods = [])
```
Nyholm, line 24: This will conflict with #677. 

Could you add a 2nd parameter `$usedClasses`?

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
```diff
@@ -308,20 +298,32 @@ private function parseResponseMap(MapShape $shape, string $input, bool $required
...
             'INPUT' => $input,
-            'FUNCTION_KEY' => \var_export($keyName, true),
+            'FUNCTION_NAME' => $functionName,
         ]);
     }
+
+    private function createPopulateMethod(string $functionName, string $body): Method
+    {
+        $method = new Method($functionName);
+        $method->setVisibility(ClassType::VISIBILITY_PRIVATE)
```
Nyholm, line 320: Have a look at #677. Sometimes it is easy to find out the return type. We should add that as a comment.

src/CodeGenerator/src/Generator/ResponseParser/ParserResult.php
jderusse: done

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
jderusse: good idea! it also removes a lot of ignored rule in psalm

src/Service/Sqs/src/Result/ReceiveMessageResult.php
```diff
@@ -29,64 +30,94 @@ protected function populateResult(Response $response): void
...
+    }
+
+    /**
+     * @return array<string, MessageAttributeValue>
+     */
+    private function populateResultMessageBodyAttributeMap(\SimpleXMLElement $xml): array
+    {
+        $items = [];
+        foreach ($xml as $item) {
+            $items[$item->Name->__toString()] = new MessageAttributeValue([
```
jderusse, line 59: previous code were `$items[$item->Name->__toString()] = !$item->Value ? null : new MessageAttributeValue([`

But this is not compatible with docbloc. Do youthink we should handle this (AS returns a map with null value)

Nyholm: Yes. The previous code was more accurate.

jderusse: yes, but re-using the previous code is a BC break with documentation.
Currently maps are typehinted `array<string, Foo>` but the code my generate `array<string, null|Foo>`.

Hmm maybe we should just `continue` when value is null.

jderusse: pushed a better fix that handl null value and compatible with ou docblock

jderusse:
thanks you

PR #680, opened 2020-06-08
Fix typos in an exception message

stof:


stof:
I don't know why phpstan is red, but I'm quite sure it is unrelated to my patch.

stof:
I think these may actually be bugs in PHPStan, not liking `return;` in generator methods.

Nyholm:
Thank you for this. 

Yes, the PHPStan issue is unrelated.

stof:
I reported the PHPStan bug in https://github.com/phpstan/phpstan/issues/3407. It is caused by generators using `yield from` and not `yield`.

stof:
@Nyholm maybe a PHPStan baseline (or an ignore rule) should be generated to ignore these errors until the bug is fixed upstream.

tyx:
As we fixed the phpstan version used through github action but we still get the latest phpstan version : https://github.com/async-aws/aws/runs/749139572?check_suite_focus=true#step:5:1

stof:
Well, you pinned the tag of the Docker image. But does this Docker image pin the phpstan version used to build it ?

stof:
@tyx reported at https://github.com/OskarStark/phpstan-ga/issues/31

PR #681, opened 2020-06-08
Improve docbloc generation

jderusse:


src/CodeGenerator/src/Generator/CodeGenerator/TypeGenerator.php
```diff
@@ -117,58 +117,25 @@ public function getPhpType(Shape $shape): array
...
-
-            if (!empty($listMemberShape->getEnum())) {
-                $memberClassNames[] = $memberClassName = $this->namespaceRegistry->getEnum($listMemberShape);
-                $doc = 'list<' . $memberClassName->getName() . '::*>';
+            [$type, $doc, $memberClassNames] = $this->getPhpType($listMemberShape);
+            if (\substr($doc, -3) === '::*') {
+                $doc = "list<$doc>";
             } else {
-                $doc = $this->getNativePhpType($listMemberShape->getType()) . '[]';
+                $doc .= '[]';
```
jderusse, line 124: I whish using `list<$doc>` everywhere, but incompatible with `array` documented in `@param` input constructor.
This would requires a bigger refactor

src/Service/DynamoDb/src/Result/BatchGetItemOutput.php
```diff
@@ -169,7 +169,7 @@ private function populateResultBatchGetRequestMap(array $json): array
     }
 
     /**
-     * @return array<string, array>
+     * @return array<string, array<string, AttributeValue>[]>
```
Nyholm, line 172: Shouldn't it be?

```suggestion
     * @return array<string, array<string, AttributeValue>>
```

jderusse: given `GetItemOutput` returns a list of map, i think it's OK to says `BatchGetItemOutput` returns a map of list of map...

Nyholm:
Could you rebase so the tests are green?

Nyholm:
Thank you

PR #682, opened 2020-06-09
Use phpstan docker image directly

tyx:
Waiting for GithubAction fix

Currently Github Action uses the latest phpstan version whatever version is specified


jderusse:
thanks

PR #692, opened 2020-06-14
[CognitoIdentityProvider] ResendConfirmationCode

rmikalkenas:


src/Service/CognitoIdentityProvider/tests/Unit/Result/ResendConfirmationCodeResponseTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+class ResendConfirmationCodeResponseTest extends TestCase
+{
+    public function testResendConfirmationCodeResponse(): void
+    {
+        // see https://docs.aws.amazon.com/cognitoidentityprovider/latest/APIReference/API_ResendConfirmationCode.html
+        $response = new SimpleMockedResponse(\json_encode([
+            'CodeDeliveryDetails' => [
+                'AttributeName' => 'a',
+                'DeliveryMedium' => 'b',
```
jderusse, line 21: Coul you please use Enum `DeliveryMediumType`?

rmikalkenas: fixed. By the way, I have noticed that there is a `changelog` file, should I add info about new endpoint to it or is it updated automatically?

jderusse: I'll apreciate it (see how other operation are handled).

Thanks

rmikalkenas: Other PR's are already merged.. Is it ok if I put all the new endpoints to this PR as a separate commit?

rmikalkenas: anyway, pushed as a separate commit. If you think that it has to be as a single commit - say and I will do change

jderusse:
thank you @rmikalkenas for all those contribution. Really appreciated.

PR #712, opened 2020-07-20
Adding S3::DeleteBucket

Nyholm:
I've only generated the code and modified `DeleteBucketRequestTest`

### DIFF
--- a/src/Service/S3/tests/Integration/S3ClientTest.php
+++ b/src/Service/S3/tests/Integration/S3ClientTest.php
@@ -217,13 +217,19 @@ class S3ClientTest extends TestCase
     public function testDeleteBucket(): void
     {
         $client = $this->getClient();
+        $client->CreateBucket(['Bucket' => 'foo-exist']);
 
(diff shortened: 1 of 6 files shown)


src/Service/S3/tests/Integration/S3ClientTest.php
```diff
@@ -213,6 +214,18 @@ public function testCreateMultipartUpload(): void
...
     }
 
+    public function testDeleteBucket(): void
+    {
+        $client = $this->getClient();
+
+        $input = new DeleteBucketRequest([
+            'Bucket' => 'my_bucket',
+        ]);
+        $result = $client->DeleteBucket($input);
```
jderusse, line 224: Could we assert that bucket does not exists aymore? (I don't know if it's implemented by the fakeS3)

Nyholm: Sure. I've fixed that.

Nyholm:
Thank you for the review.

PR #719, opened 2020-07-29
Ability to add additional response content to failing result mock factory.

rmikalkenas:


src/Core/tests/Unit/Test/ResultMockFactoryTest.php
```diff
@@ -80,4 +80,23 @@ public function testCreateFailling()
...
+            Result::class,
+            400,
+            'Boom',
+            'ResourceNotFoundException',
+            'com.amazonaws.dynamodb.v20120810'
+        );
+
+        try {
+            $result->resolve();
+            self::fail('ClientException should be thrown');
```
Nyholm, line 96: We should return in the `catch` clause and move this row after the try-catch. 

Say that `$result->resolve();` throws a `\RuntimeException`, then this test wont fail.

src/Core/src/Test/ResultMockFactory.php
```diff
@@ -30,16 +30,26 @@ class ResultMockFactory
...
+    ) {
         if (Result::class !== $class) {
             $parent = get_parent_class($class);
             if (false === $parent || Result::class !== $parent) {
                 throw new \LogicException(sprintf('The "%s::%s" can only be used for classes that extend "%s"', __CLASS__, __METHOD__, Result::class));
             }
         }
 
-        $httpResponse = new SimpleMockedResponse(\json_encode(['message' => $message]), ['content-type' => 'application/json'], $code);
+        $awsTypeContent = [];
```
Nyholm, line 47: This should be named `$headers`, right?

src/Core/src/Test/ResultMockFactory.php
```diff
@@ -30,16 +30,26 @@ class ResultMockFactory
...
      * @return Result|T
      */
-    public static function createFailing(string $class, int $code, ?string $message = null)
-    {
+    public static function createFailing(
+        string $class,
+        int $code,
+        ?string $message = null,
+        ?string $awsCode = null,
+        ?string $awsType = null
```
Nyholm, line 38: Hm.. What would be the pros and cons by adding a 4th argument `$headers` instead of `$awsCode` and `$awsType`?

src/Core/src/Test/ResultMockFactory.php
rmikalkenas: Its not headers as `__type` argument is passed in the body content: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.Components

src/Core/src/Test/ResultMockFactory.php
rmikalkenas: Thought about it.. I did a bit of investigation and found out that some others aws apis might return additional `error` argument as array (https://docs.aws.amazon.com/storagegateway/latest/userguide/AWSStorageGatewayAPI.html#RESTErrorResponses)
So maybe youre right. Lets replace `$awsCode` and `$awsType` variables with like `array $additionalContent = []`. What do you think about it?

src/Core/src/Test/ResultMockFactory.php
```diff
@@ -30,16 +30,26 @@ class ResultMockFactory
...
         }
 
-        $httpResponse = new SimpleMockedResponse(\json_encode(['message' => $message]), ['content-type' => 'application/json'], $code);
+        $awsTypeContent = [];
+        if (null !== $awsCode || null !== $awsType) {
+            $awsTypeContent['__type'] = \sprintf('%s#%s', (string) $awsType, (string) $awsCode);
+        }
+
+        $httpResponse = new SimpleMockedResponse(\json_encode(\array_merge(['message' => $message], $awsTypeContent)), ['content-type' => 'application/json'], $code);
         $client = new MockHttpClient($httpResponse);
```
jderusse, line 53: Couldn't we just remove typehint `string` fom message. and let user provide an array with the `__type` argument?

rmikalkenas: @jderusse could you clarify what you mean by `removing string typehint from message`? 
Do I understand correctly that yours proposal is to make $message type mixed and check if its string, then add "message" to body and if its array - just pass it as response body?

jderusse: nvm, I prefer the last version of your PR ;-)

jderusse:
Thanks you

PR #725, opened 2020-08-02
[Core] Minor syntax update

Nyholm:


src/Core/src/Response.php
```diff
@@ -373,12 +373,8 @@ private function getResolveStatus(): bool
         }
 
         if ($this->resolveResult instanceof Exception) {
-            if (404 === $code) {
-                $this->logger->info($message ?? $this->resolveResult->getMessage(), $context);
-            } else {
-                $this->logger->error($message ?? $this->resolveResult->getMessage(), $context);
-            }
-
+            $level = 404 === $code ? 'info' : 'error';
```
bendavies, line 376: inline this var?

src/Core/src/Response.php
```diff
@@ -373,12 +373,8 @@ private function getResolveStatus(): bool
         }
 
         if ($this->resolveResult instanceof Exception) {
-            if (404 === $code) {
-                $this->logger->info($message ?? $this->resolveResult->getMessage(), $context);
-            } else {
-                $this->logger->error($message ?? $this->resolveResult->getMessage(), $context);
-            }
-
```
jderusse, line 381: use constants?

Nyholm: Sure

src/Core/src/Response.php
Nyholm: Okey.

PR #726, opened 2020-08-03
Add retry-http client

### LINKED ISSUE #708: Add support for retries if request fails
Hi, I start using this package instead of the official AWS SDK over a week ago, and since then I have been having connection issues, while new sending messages to SQS.

I run a serverless setup: "SQS => Lambda trigger" and execute [Symfony Messenger](https://github.com/symfony/amazon-sqs-messenger) with  [bref/symfony-messenger](https://github.com/brefphp/symfony-messenger/pull/31)
During the handling of the messages, new messages are pushed to SQS. 
Very occasionally this fails, but I dont understand why. The only difference I can think of with the AWS SDK is that it implements an [retry schema](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/faq.html#how-is-retry-scheme-handled-in-the-sdk-php).

Does anyone else experience these occasional failures? Some of the errors messages I extracted:
- fopen(https://52.95.120.176/): failed to open stream: Connection timed out
- fopen(https://52.95.112.152/): failed to open stream: HTTP request failed!
- fopen(): SSL: Handshake timed out
- Server returned nothing (no headers, no data) for "https://sqs.eu-west-1.amazonaws.com/".

So I was considering, would it be possible/wise/useful to create a Retry middleware of some sort, that would handle these failures, and retry if appropriate?
The HttpClient should have some support for middlewares according to https://github.com/symfony/symfony/pull/36779
The retry mechanism could mimimic eg https://github.com/aws/aws-sdk-php/blob/master/src/RetryMiddleware.php

jderusse:
This PR provide an implementation of HttpClientInterface that automatically retries failed requests, which should fixes #708

This decorator retries requests when an `NetworkException` is triggered or  reponse code is `> 500`.

note: The official AWS also retries requests that return specific exceptions like "RequestLimitExceeded" or "RequestThrottledException"... 
I choose to not handle such case, because 
- I think it should be the responsibility of the user to retry such case
- the AWS API is not consistent and Exception code is stored in many places (think JSON vs XML) handling such failure should be handled in an higher level (response should be parsed) and would be much more complicated.

note: I didn't used the `AsyncResponse` to keep compatibility with `symfony/http-client 4.4`

I think we can provide an integration test case, but I'm looking for a server that could return programmatic responses (like return 500, then 200). Do you know how to do it? Maybe with something similar to the symfony `TestHttpServer`?

@nicolas-grekas  could you please review this PR?..


src/Service/S3/tests/Integration/S3ClientTest.php
```diff
@@ -352,6 +352,7 @@ public function testGetObjectConsistent(): void
         self::assertEquals('content', $result->getBody()->getContentAsString());
         self::assertEquals('content', \stream_get_contents($result->getBody()->getContentAsResource()));
         self::assertEquals('content', \implode('', \iterator_to_array($result->getBody()->getChunks())));
+        self::assertEquals('content', \implode('', \iterator_to_array($result->getBody()->getChunks())));
```
Nyholm, line 355: Why do we need this a second time?

jderusse: the `ResponseBodyStream` cache previous iteration in a fallback variable. This second check assert that both original, and fallback return the right response.

I notice (because of another bug) that test failed only on the second call.

jderusse: added comment about that

src/Core/src/HttpClient/RetryResponse.php
```diff
@@ -0,0 +1,208 @@
...
+        $this->options = $options;
+        $this->maxTryCount = $maxTryCount;
+
+        $this->logger = $logger ?? new NullLogger();
+        $this->inner = $this->client->request($this->method, $this->url, $this->options);
+    }
+
+    public function getStatusCode(): int
+    {
+        $this->initialized or $this->initialize();
```
Nyholm, line 53: Isn’t “or” frowned upon? Let’s use double pipe. :)

src/Core/src/HttpClient/RetryResponse.php
```diff
@@ -0,0 +1,208 @@
...
+    public function toArray(bool $throw = true): array
+    {
+        $this->initialized or $this->initialize();
+
+        return $this->inner->toArray($throw);
+    }
+
+    public function cancel(): void
+    {
+        $this->initialized = true;
```
Nyholm, line 81: Good

src/Core/src/HttpClient/RetryResponse.php
```diff
@@ -0,0 +1,208 @@
...
+
+    public static function stream(iterable $responses, float $timeout = null): \Generator
+    {
+        $wrappedResponses = [];
+        $map = new \SplObjectStorage();
+        $client = null;
+
+        foreach ($responses as $r) {
+            if (!$r instanceof self) {
+                throw new \TypeError(sprintf('"%s::stream()" expects parameter 1 to be an iterable of AsyncResponse objects, "%s" given.', static::class, get_debug_type($r)));
```
Nyholm, line 100: “AsyncResponse”, isn’t that wrong? You mean RetryResponse. 

Let’s use self::class

src/Core/src/HttpClient/RetryResponse.php
```diff
@@ -0,0 +1,208 @@
...
+            }
+
+            if (null === $client) {
+                $client = $r->client;
+            } elseif ($r->client !== $client) {
+                throw new TransportException('Cannot stream AsyncResponse objects with many clients.');
```
Nyholm, line 106: Same

src/Core/src/HttpClient/RetryResponse.php
```diff
@@ -0,0 +1,208 @@
...
+                    yield $response => new ErrorChunk(0, sprintf('Idle timeout reached for "%s".', $response->getInfo('url')));
+                }
+
+                break;
+            }
+
+            foreach ($client->stream($wrappedResponses, $subTimeout) as $response => $chunk) {
+                /** @var self $r */
+                $r = $map[$response];
+                if (!$chunk->isTimeout() && $chunk->isFirst()) {
```
Nyholm, line 136: I don’t understand this. 
If it isn’t a retry and the first chunk...

src/Core/src/HttpClient/RetryResponse.php
```diff
@@ -0,0 +1,208 @@
...
+                        'statusCode' => $status,
+                        'attempt' => $this->tryCount,
+                        'maxAttempts' => $this->maxTryCount,
+                    ]);
+                    $this->inner = $this->client->request($this->method, $this->url, $this->options);
+
+                    return true;
+                }
+
+                $this->logger->info('HTTP request failed with status code {statusCode}. Stop after {maxAttempts} attempts.', [
```
Nyholm, line 179: Is anything else logged or is this the only message you see?

We should maybe consider to make this an “error”.

src/Core/src/HttpClient/RetryResponse.php
```diff
@@ -0,0 +1,208 @@
...
+                    'statusCode' => $status,
+                    'maxAttempts' => $this->maxTryCount,
+                ]);
+            }
+            $this->initialized = true;
+
+            return false;
+        } catch (TransportExceptionInterface $e) {
+            if (++$this->tryCount <= $this->maxTryCount) {
```
Nyholm, line 188: These next 15 lines looks duplicated.

Nyholm:
Thank you. 

I have a few questions. But it is most because I don’t know things. I just want to understand.

src/Core/src/HttpClient/RetryResponse.php
jderusse: the method `handleRetry` checks if a retry is needed and possible, and return true id the request is retried.

we only check for a retry on first chunk, because we only need the status code to decide if we need to retry the request.

src/Core/src/HttpClient/RetryResponse.php
jderusse: refactored

src/Core/src/HttpClient/RetryResponse.php
jderusse: good idea

src/Core/src/HttpClient/RetryResponse.php
jderusse: fixed

src/Core/src/HttpClient/RetryResponse.php
jderusse: fixed

src/Core/src/HttpClient/RetryResponse.php
jderusse: replaced by `if`

src/Core/src/HttpClient/RetryResponse.php
Nyholm: But if it is a timeout, dont we want to retry then?

jderusse: I think `timeout` here occurred when you stream the response with a userland timeout. The request is not resolved at this point. 
A timeout in network communication would throw a `NetworkException`.

Am I right @nicolas-grekas ?

src/Core/src/HttpClient/RetryResponse.php
```diff
@@ -0,0 +1,213 @@
...
+     */
+    private function handleRetry(): bool
+    {
+        if ($this->initialized) {
+            return false;
+        }
+
+        $handle = function (string $message, array $context): bool {
+            if (++$this->tryCount <= $this->maxTryCount) {
+                $this->logger->info($message . ' Retry the request {attempt}/{maxAttempts}.', $context + [
```
Nyholm, line 178:
```suggestion
                $this->logger->info($message . ' Retry the request {attempt} of {maxAttempts}.', $context + [
```

It seams weird if the retry request 2 thirds (2/3)

Nyholm (approved):

Nyholm:
I have a project where I do many short bursts of HTTP requests. I often get hit by AWS rate limits. 

This PR opens up for a possibility of doing exponential back-offs and slowing things down if needed.

jderusse:
this PR does not sleep between retry.
do you think we should introduce exponential back of here ?

Nyholm:
This PR (or a future PR) could introduce a "RetryDecider" and a "RetryStrategy". We could let the user provide custom logic for those functions/callbacks/classes. The default should be as you implemented here.

src/Core/src/HttpClient/RetryResponse.php
jderusse: fixed

jderusse:
Thank for the review. Please don't merge it now. 

I've opened a PR to add this client in `symfony`.

jderusse:
Completly refactored that PR to leverage the new RetryableHttpClient provided by symfony.

src/Core/src/AbstractApi.php
```diff
@@ -61,8 +62,14 @@ public function __construct($configuration = [], ?CredentialProvider $credential
             throw new InvalidArgument(sprintf('Second argument to "%s::__construct()" must be an array or an instance of "%s"', __CLASS__, Configuration::class));
         }
 
-        $this->httpClient = $httpClient ?? HttpClient::create();
         $this->logger = $logger ?? new NullLogger();
+        if (!isset($httpClient)) {
+            $httpClient = HttpClient::create(['timeout' => 10]);
```
Nyholm, line 67: Why should we define a timeout?

src/Core/src/AbstractApi.php
```diff
@@ -61,8 +62,14 @@ public function __construct($configuration = [], ?CredentialProvider $credential
...
+            if (\class_exists(RetryableHttpClient::class)) {
+                $httpClient = new RetryableHttpClient($httpClient, null, null, 3, $this->logger);
```
Nyholm, line 69: Maybe make the number of retries configurable?

Ie, Add a config option as "retry: 1". If the number is >1 then we use the retry client.

jderusse: IMHO, People can inject a client if they want configurable retry.

This is available out of the box via FrameworkBundle configuration:

```
framework:
  http_client:
    retry_failed:
      max_retries: 3

async_aws: ~
```

Nyholm: Yeah, But for people that isn't using Symfony.. 

Im hesitating to use RetryableHttpClient by default. That is why I want to use it opt-in. 

And as you say, for Symfony users it will be pretty much automatically without any changes here.

jderusse: hmm. Would you disabled this feature? Do you think most people prefer dealing with errors by them self instead of let the SDK retry them?

nb: retry is enabled by default in official SDK

Nyholm: Hm.. maybe Im wrong. 

I am worried about what the side effects would be... But we are only retrying things that we know are safe to retry. Ie 500, right?

So the side effects would be none? Or... If AWS is down, then my request will be a 500, and I'm retrying it two times before I get back control, right? So it is a slightly longer delay if this (unlikely) scenario happens. 

Do you know if this has any effects on rate limits?

jderusse: The retries status code are currently `423, 425, 429, 500, 502, 503, 504, 507, 510` which should be safe with AWS.
It also retries on NetworkException on the first chunk. 

Meaning, there is an hypothetical case, where, the server receive the request, process it, but does not return the response or the network failed while receiving the first bytes of the requests.

IMHO their is 2 choices:
- the implementation is safe and can be enable by default
- the implementation is dangerous (retries no-idempotent calls) and should not be used. And we have to either fix the RetryableClient or find an alternative for #708

Nyholm: Thank you for the explanation. It is me that dont know enough about this new retry client. 

The implementation is save since these response codes should be retried.

Nyholm (approved):

Nyholm:
This update with Symfony's changes will solve the problems in https://github.com/async-aws/aws/pull/726#issuecomment-691637964

Great work with this feature

jderusse:
After thinking twice, the edge case (request handled by server, and proceeded, but network failed before sending the first byte) is rare enough to be retried.

PR #735, opened 2020-08-20
Migrate PHPUnit configuration to version 9

Nyholm:
This will take care of the warning about migrating configuration. 

Screenshot 2020-08-20 at 14 02 29

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,5 +14,5 @@ trim_trailing_whitespace = true
 [Makefile]
 indent_style = tab
 
-[src/**/Makefile]
-indent_style = tab
+[phpunit.xml.dist]
+indent_size = 2
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,40 +1,37 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.2/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd"
          backupGlobals="false"
          colors="true"
          bootstrap="tests/bootstrap.php"
          failOnRisky="true"
          failOnWarning="true"
 >
-    <php>
-        <ini name="error_reporting" value="-1" />
-    </php>
-
-    <testsuites>
-        <testsuite name="Test Suite">
-            <directory>./tests/</directory>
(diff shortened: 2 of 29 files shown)


jderusse (approved):
looks good to me

Nyholm:
Hm.. On PHP 7.x we are running PHPUnit 8. That gives a warning about config that is not recognized. 

I should make sure we are running PHP 9 as often as we can

jderusse:
> I should make sure we are running PHP 9 as often as we can

that's the job of phpunit-bridge.

a solution would be to run `--migtate-confoguration` in CI before running the test.

Nyholm:
Im more happy to get a warning about invalid config on PHP 7.2 than a warning about deprecated config on PHP8.0. 

I specify what version I want to the bridge.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,9 @@ on:
     branches:
       - master
 
+env:
+  SYMFONY_PHPUNIT_VERSION: '9.3'
```
jderusse, line 10: the phpunit-bridge should take care of this. I'm not comfortable with maintaining a hardcoded version in this repository :-/

Nyholm: Okey. That is fair. 

Im not happy with rewriting config in CI =)

So, the question is: What warning do we like?
Option one: Deprecation warning (shown in the description)
Option two: A unknown config warning: 

Screenshot 2020-08-20 at 14 50 03


Im fine with whatever. When I started the PR I did not know of the "unknown config warning".

jderusse: hmm "unknown config" is just a warning or a behavior is not enabled?
I guess it's about coverage, and given we don't use it, it's not a real issue.

I'm fine with beeing compatible with the last version (fix deprecation). But the lower version should work as expected.

Nyholm: I agree, it is no real issue. 

I've updated the PR

Nyholm:
PR is updated, No CI changes. 

But I did update .editorconfig to use indentation_size=2 on phpunit.xml.dist. (It seams that 2 is preferred by PHPUnit)

.editorconfig
```diff
@@ -14,5 +14,5 @@ trim_trailing_whitespace = true
 [Makefile]
 indent_style = tab
 
-[src/**/Makefile]
-indent_style = tab
+[phpunit.xml.dist]
+indent_size = 2
```
jderusse, line 18: good catch

jderusse (approved):

.editorconfig
Nyholm: Thank you

PR #736, opened 2020-08-20
Make sure we dont have any direct deprecation notices

Nyholm:
This will make the build green... hopefully

### DIFF
--- a/src/Integration/Symfony/Bundle/phpunit.xml.dist
+++ b/src/Integration/Symfony/Bundle/phpunit.xml.dist
@@ -10,7 +10,7 @@
 >
     <php>
         <ini name="error_reporting" value="-1" />
-        <server name="SYMFONY_DEPRECATIONS_HELPER" value="0" />
+        <server name="SYMFONY_DEPRECATIONS_HELPER" value="max[direct]=0" />
     </php>
 
     <testsuites>


Nyholm:
Wohoo!

The build is green on PHP8

jderusse:
I'm not sure about this one.. Shouldn't we warned when having a (maybe unmaintained) dependency that triggers deprecation?

Nyholm:
We are still warned. See

Screenshot 2020-08-20 at 14 45 33

But the build will not fail. 

The deprecation error is only on PHP8 because Symfony is using `libxml_disable_entity_loader()`

Screenshot 2020-08-20 at 14 46 06

To have the build red because if this is too much. There is nothing we can do about it but fixing the dependency's code or to not use that dependency.

jderusse:
> To have the build red because if this is too much. There is nothing we can do about it but fixing the dependency's code or to not use that dependency.

I made the PR, and it has been merged few days ago. Next release should fix it.

I personally rarely looks hat build logs when the CI is green. I afraid to miss issue with this change.

Nyholm:
That is awesome. But.. should we care so much about an indirect dependency’s deprecation notice?

I guess the build should be red at that dependency too.

jderusse:
yeah that's the question... 
given we only rely on highly maintained dependencies, I guess it's ok then

jderusse (approved):

Nyholm:
Thank you

PR #754, opened 2020-08-29
Make sure we select region properly

Nyholm:
Using `@region` had no effect

### DIFF
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## NOT RELEASED
 
+## 1.4.1
+
+### Fixed
+
+- Make sure passing `@region` to an API operation has effect.
+- Check that both AWS access id and secret exists before using them.
+
 ## 1.4.0
 
 ### Added
--- a/src/Core/tests/Unit/AbstractApiTest.php
+++ b/src/Core/tests/Unit/AbstractApiTest.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Core\Tests\Unit;
+
+use AsyncAws\Core\AbstractApi;
+use AsyncAws\Core\Configuration;
+use PHPUnit\Framework\TestCase;
+
+class AbstractApiTest extends TestCase
+{
+    public function testGetEndpointRegion()
+    {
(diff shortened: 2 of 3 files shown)


src/Core/tests/Unit/AbstractApiTest.php
```diff
@@ -0,0 +1,54 @@
...
+    public function testGetEndpointRegion()
+    {
+        $api = new DummyApi();
+
+        // Use default region
+        $endpoint = $api->getEndpoint('/some/path', [], null);
+        self::assertEquals('https://foobar.us-east-1.amazonaws.com/some/path', $endpoint);
+
+        $endpoint = $api->getEndpoint('/some/path', [], 'eu-central-1');
+        self::assertEquals('https://foobar.eu-central-1.amazonaws.com/some/path', $endpoint);
```
Nyholm, line 22: This failed without the patch.

Nyholm:
@jderusse please verify so I can release this after merge

src/Core/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## NOT RELEASED
 
+## 1.4.1
+
+### Fixed
+
+- Make sure passing `@region` to an API operation has effect.
+- Check that both AWS access id and secret exists before using them.
```
jderusse, line 10: We should check that change log is update for each PR. (maybe with a not-required test).

This is hard and a time consuming work when releasing new version

jderusse (approved):

src/Core/CHANGELOG.md
Nyholm: Do you mean that every PR should include a change log?

I've tried that before, but when you have multiple PR open it will just cause merge conflicts. Ie, you cannot merge two PRs to same component without rebasing one of them first. 

If we want to automate the changelog, we should maybe look info what the official SDK is using.

jderusse: yes this is what I mean,

I'm not sure about automated PR (hard to guess what changed... comments, new constant, deprecation...) 

For conflict, yes, it's a drawback..

It was just a thought about this

Nyholm: Let's try to force changelog. Let's see how much problem the conflicts are

PR #758, opened 2020-08-30
Add a "debug" config to print HTTP requests and responses

Nyholm:
It could be helpful to see exactly what HTTP requests and responses we are dealing with. The official SDK prints this to stdout but we use PSR3.

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -136,6 +136,15 @@ abstract class AbstractApi
             ]
         );
 
+        if (\filter_var($configuration->get('debug'), \FILTER_VALIDATE_BOOLEAN)) {
+            $this->logger->debug('AsyncAws HTTP request sent', [
+                'method' => $request->getMethod(),
+                'endpoint' => $request->getEndpoint(),
+                'headers' => json_encode($request->getHeaders()),
+                'body' => 0 === $length ? null : $requestBody,
+            ]);
+        }
+
         return new Response($response, $this->httpClient, $this->logger);
     }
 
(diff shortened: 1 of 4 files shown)


src/Core/src/AbstractApi.php
```diff
@@ -136,6 +136,15 @@ protected function getSignatureScopeName(): string
             ]
         );
 
+        if (\filter_var($configuration->get('debug'), \FILTER_VALIDATE_BOOLEAN)) {
+            $this->logger->debug('AsyncAws HTTP request sent', [
+                'method' => $request->getMethod(),
+                'endpoint' => $request->getEndpoint(),
+                'headers' => json_encode($request->getHeaders()),
+                'body' => 0 === $length ? null : $requestBody,
```
jderusse, line 144: is it an issue when body is a stream ?

jderusse (approved):
couldn't it be provided by the HttpClientLogger ? just asking

src/Core/src/AbstractApi.php
Nyholm: We've converted the body to a string just a few lines above.

Nyholm:
What is the `HttpClientLogger`? I've never heard of it and I cannot find any references to it..

jderusse:
nvm, I thought Symfony provides such things. I was trapped by https://github.com/symfony/symfony/pull/30537/files

Nyholm:
So after reading the source: The logger in the HttpClient does not capture any headers or body. I think that is important to include here.

Nyholm:
This PR is ready for review

jderusse:
thank you

PR #761, opened 2020-08-30
Allow non-standard region with custom endpoint

### LINKED ISSUE #757: S3 client unusable with Scaleway
Hi,

I would like to use the S3 client to connect to a Scaleway S3 bucket, but I get an `UnsupportedRegion` exception.

The region I have to use is `fr-par`, but it's not listed in the switch of the [`getEndpointMetadata`](https://github.com/async-aws/aws/blob/6cf3d03ec4dcb36a41d59adae6236015a7a9901c/src/Service/S3/src/S3Client.php#L735) method.
I tried to omit the region but then I get an error from the endpoint : `The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'fr-par'`

[Documentation](https://async-aws.com/clients/s3.html#non-aws-s3-endpoints) says that it is possible to use alternative endpoints, but either I'm missing something in the configuration or it is not possible.

In my case the `getEndpointMetadata()` is only called in [`getSigner()`](https://github.com/async-aws/aws/blob/6cf3d03ec4dcb36a41d59adae6236015a7a9901c/src/Core/src/AbstractApi.php#L224), maybe the call should be avoided when a custom endpoint is provided in configuration (like it's done in [`getEndpoint()`](https://github.com/async-aws/aws/blob/6cf3d03ec4dcb36a41d59adae6236015a7a9901c/src/Core/src/AbstractApi.php#L192)) ?
But then it may be necessary to provide `signService` and `signVersion` otherwise (via configuration ?)

jderusse:
Fixes #757 by using "default" signature information when calling a custom endpoint

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -228,11 +228,12 @@ abstract class AbstractApi
         if (!isset($this->signers[$region])) {
             $factories = $this->getSignerFactories();
             $factory = null;
-            if (!$this->configuration->isDefault('endpoint') && !$this->configuration->isDefault('region')) {
+            if ($this->configuration->isDefault('endpoint') || $this->configuration->isDefault('region')) {
+                $metadata = $this->getEndpointMetadata($region);
+            } else {
+                // Allow non-aws region with custom endpoint
                 $metadata = $this->getEndpointMetadata(Configuration::DEFAULT_REGION);
                 $metadata['signRegion'] = $region;
-            } else {
-                $metadata = $this->getEndpointMetadata($region);
             }
 
             foreach ($metadata['signVersions'] as $signatureVersion) {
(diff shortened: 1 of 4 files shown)


src/Core/src/AbstractApi.php
```diff
@@ -226,9 +226,15 @@ private function getSigner(?string $region)
...
         if (!isset($this->signers[$region])) {
-            $metadata = $this->getEndpointMetadata($region);
             $factories = $this->getSignerFactories();
             $factory = null;
+            if (!$this->configuration->isDefault('endpoint') && !$this->configuration->isDefault('region')) {
+                $metadata = $this->getEndpointMetadata(Configuration::DEFAULT_REGION);
+                $metadata['signRegion'] = $region;
+            } else {
+                $metadata = $this->getEndpointMetadata($region);
+            }
```
Nyholm, line 236: You could invert this if-statement to improve readability


```php

```


```suggestion
            if ($this->configuration->isDefault('endpoint') || $this->configuration->isDefault('region')) {
                $metadata = $this->getEndpointMetadata($region);
            } else {
                // Allow non-aws region with custom endpoint
                $metadata = $this->getEndpointMetadata(Configuration::DEFAULT_REGION);
                $metadata['signRegion'] = $region;
            }
```

Nyholm:
Thank you

PR #764, opened 2020-09-02
Update to avoid bug in symfony/http-kernel: >5.0, <5.1.5

Nyholm:
Should we do this? 

Dependabot clearly recommends us to do this.. But it is a security issue in a class we are not using. 

Im also considering what responsibility we have compared to the end user.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
         "symfony/deprecation-contracts": "^2.1",
         "symfony/filesystem": "^5.0",
         "symfony/finder": "^4.4 || ^5.0",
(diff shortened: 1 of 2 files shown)


jderusse:
IMHO This is not our call to define such rules.

Moreover, the vulnerability occurs when HttpClient is combined with HttpKernel Cache which is not the case of our library
The vulnerability occurs when you call untrusted server. Which is probably not the case when calling AWS API.

People are free to use `async-aws` with HttpKernel 5.0 without any issue.

Nyholm:
Thank you for a well written argument. 

I agree with you.

PR #766, opened 2020-09-03
Refresh changelog

jderusse:
This PR refactor the way the bot fetch the last version of AWS to regenerate code.

It use changelog files in AWS repo (https://github.com/aws/aws-sdk-php/tree/master/.changes) to generate a changelog in our repo.

sample of changelog generated with the last version  of https://github.com/async-aws/aws/pull/765
```diff
--- a/src/Service/Sqs/CHANGELOG.md
+++ b/src/Service/Sqs/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Fixed
+
+- AWS enhancement: Documentation updates for SQS.
+
 ## 1.3.0
 
 ### Added
```
It can not be 100% accurate, because a same release of AWS may update several operations, and provide several entries in the changelog if one of that update change the source code, the bot will pick all the entries from the changelogs.

But the entries are grouped by service (S3, lambda, ...): the chance to get such case should be LOW.

And I think we can then easily remove bad entries from our CHANGELOG when reviewing the PR (by submiting a suggestion in a comment)

here is an sample of JOB in github action https://github.com/async-aws/aws/runs/1066054304?check_suite_focus=true

### DIFF
--- a/.github/workflows/bin/refresh
+++ b/.github/workflows/bin/refresh
@@ -1,0 +1,149 @@
+#!/usr/bin/env php
+<?php
+
+$rootDir = __DIR__ . '/../../..';
+if (!file_exists($rootDir . '/manifest.json')) {
+    echo "Unable to find the `manifest.json` script in `../../..`.\n";
+    exit(1);
+}
+
+function console_log(string $message): void
+{
+    echo $message.PHP_EOL;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


Nyholm (approved):
Thank you. 

I think this looks really good.

PR #769, opened 2020-09-10
Fix missing env variables

jderusse:
This PR fixes https://github.com/async-aws/aws/pull/633#issuecomment-670718131 about undefined env variable by fallbacking on `$_ENV`, `$_SERVER` then `getenv`.

This is inspired from https://github.com/symfony/symfony/blob/be6146c56625b8b1ad5b8fec5bdb5dc6d4023234/src/Symfony/Component/DependencyInjection/EnvVarProcessor.php#L130-L134

### DIFF
--- a/src/Core/src/EnvVar.php
+++ b/src/Core/src/EnvVar.php
@@ -5,10 +5,10 @@ declare(strict_types=1);
 namespace AsyncAws\Core;
 
 /**
- * Helper to safely read environment variables
+ * Helper to safely read environment variables.
  *
  * @author Jérémy Derussé <jeremy@derusse.com>
-
+ *
  * @internal
  */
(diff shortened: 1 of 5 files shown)


src/Core/src/EnvVar.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Core;
+
+final class EnvVar
```
Nyholm, line 7: Should we mark it as internal?

jderusse: indeed. done

chalasr (approved):
👍

jderusse:
thank you

PR #775, opened 2020-09-13
[Core] Prepare for 1.5 release

Nyholm:


jderusse (approved):

chalasr (changes requested):
The composer branch-alias needs to be updated, right? looks like we need some automation :)

Nyholm:
I will be happy to add that automation =)

I've got some questions for the automation:

What is the rule? When we release 1.5.0, then we should update the branch alias to: 

```json
        "branch-alias": {
            "dev-master": "1.6-dev"
        }
```
Right?

---------

I've heard people saying that the branch alias should be updated only when new changes has been added after 1.5.0 is tagged. 

-------

Is it important that the new branch alias is changed before we tag 1.5.0?

chalasr:
From my understanding and looking at Symfony, the branch alias should indeed be updated right after the tag creation. 
This one should have been changed to `1.5-dev` when 1.4 was released, and should change to `1.6-dev` once 1.5 released.

chalasr (approved):
Ready to go!

Nyholm:
Cool. I'll make sure to automate that. =)

jderusse (approved):

PR #783, opened 2020-09-18
[Main] Minor cleanup

Nyholm:
An earlier version of this test needed this. But that is not true any more. 

Ie, The `*` will no longer match a `dot-file` (`.github`).

### DIFF
--- a/tests/Unit/MetaFileTest.php
+++ b/tests/Unit/MetaFileTest.php
@@ -47,8 +47,6 @@ class MetaFileTest extends TestCase
         $root = \dirname(__DIR__, 2);
         $finder
             ->files()
-            ->ignoreDotFiles(false)
-            ->ignoreVCS(true)
             ->in($root . '/src/*/.github/workflows')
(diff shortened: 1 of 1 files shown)


tests/Unit/MetaFileTest.php
```diff
@@ -47,8 +47,6 @@ public function testBranchAlias()
         $root = \dirname(__DIR__, 2);
         $finder
             ->files()
-            ->ignoreDotFiles(false)
-            ->ignoreVCS(true)
             ->in($root . '/src/*/.github/workflows')
             ->in($root . '/src/*/*/.github/workflows')
             ->in($root . '/src/*/*/*/.github/workflows')
```
jderusse, line 52: does '`src/**/.github/workflows` works? Just asking

jderusse (approved):

Nyholm:
I tried that too. It didn’t work :(

jderusse:
thank you

PR #795, opened 2020-10-01
[Symfony] Make sure we can override config

### LINKED ISSUE #793: Bug on overriding configuration in Symfony Integration
Before I explain the bug I'm not sure if this is a bug on the Symfony Integration or from the Symfony itself.

What is hapenning:

I have the `config/async_aws.yaml` with the following:
image

I'm trying to override it for `dev` enviroment with the following in this file `config/dev/async_aws.yaml`:

image

This is the error I get:

image

I was able to debug it and I noticed that when processing the config files when merging it does not keep the names of the files it is overriding, here is a printscreen of it:

image

What I would expect is that the `sharedCredentialsFile` had the value that is in the `0` key of the `$currentConfig`.

Using:
- Symfony 4.4.13
- async-aws/async-aws-bundle:1.0.0
- async-aws/dynamo-db-session:0.1.1
- PHP 7.2

Nyholm:
This will fix #793 

Related to https://github.com/symfony/symfony/pull/38371

### DIFF
--- a/src/Integration/Symfony/Bundle/CHANGELOG.md
+++ b/src/Integration/Symfony/Bundle/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## NOT RELEASED
 
+## 1.1.1
+
+### Fixed
+
+- Make sure you can override config in different environments.
+
 ## 1.1.0
 
 ### Added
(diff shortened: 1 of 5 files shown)


jderusse (approved):

Nyholm:
Thank you for the review

PR #814, opened 2020-10-13
Add support for league/flysystem-adapter-test-utilities 2.0-beta3

Nyholm:


jderusse (approved):
I'm not sure why the change. I trust you and the CI on this

Nyholm:
Sorry. 

Why? We currently support beta2 only. This PR makes us support beta3. 

There was a tiny BC break in this dependency that we use for tests. The change will speed up tests.

PR #818, opened 2020-10-17
Fix for numeric bucket names like 0344-yyy

emrysal:
When interpolating bucket names that start with a number the $0 is
interpolated to $0\d, causing the https:// part to be stripped from the
endpoint resulting in a "Notice: Undefined index: scheme" error.

### DIFF
--- a/src/Service/S3/CHANGELOG.md
+++ b/src/Service/S3/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Fixed endpoint issue when a bucket name started with a number.
 - Improve StorageClass documentation.
 
 ## 1.5.0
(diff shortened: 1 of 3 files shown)


Nyholm:
Thank you for this. 
Could you also add a unit test for this?

jderusse (commented):
Thank you.
Could you please add test and an entry in the changeling?

emrysal:
Added test which passes, but before this fix:

2) AsyncAws\S3\Tests\Unit\S3ClientTest::testBucketToHost
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'https://61515-bar.s3.amazonaws.com/'
+'1515-bar.s3.amazonaws.com/'

+ Added a changelog entry

Nyholm (approved):
Thank you very much for this"

Nyholm:
This is now released: https://github.com/async-aws/s3/releases/tag/1.5.1

PR #819, opened 2020-10-18
fix php-cs-fixer running on generated tests

bendavies:
running `php-cs-fixer` on generated test was not working as the wrong directory was being formatted - this fixes it.


Nyholm:
Thank you for this. 

Just be aware that not all clients use the same namespace. See the STSClient. 

Also note that you may run `composer update` on a specific client. Then we got a vendor folder. We dont want to run `php-cs-fixer` on a vendor folder.

bendavies:
right ok, i've popped off that 2nd commit. just the fix is in place now!

bendavies:
BTW, `vendor` is excluded by default by `PhpCsFixer\Finder::create()`

jderusse (approved):
Thank you for catching this  :heart: .. 

I already noticed that adding new methods didn't run "cs-fixer" on tests, but didn't get time to look into it

Nyholm:
Awesome. Thank you

PR #824, opened 2020-10-20
Retry non-idempotent methods

### LINKED ISSUE #708: Add support for retries if request fails

jderusse:
Fixes #708 (again)

This PR override the default Configuration of Symfony's retried status code. To always retries failed requests (even on non-idempotent methods)

With AWS almost all call use POST (not idempotent) meaning errors likes 500 or network issue will not be retried.

But almost all operations are either safe (like Sqs::ListQueues) or idempotent (like Iam::CreateUser, that check user uniqueness)
I think, only Sqs::SendMessage, Ses::SendEmail, Sns::Publish would be "vulnerable". 

My personal opinion, I much prefer sending Sqs Message twice than failing sending one.

This default behavior is overriddable by the user by providing an `AwsRetryStrategy` with `GenericRetryStrategy::DEFAULT_RETRY_STATUS_CODES`.

/cc @starred-gijs

### DIFF
--- a/src/Core/src/HttpClient/AwsRetryStrategy.php
+++ b/src/Core/src/HttpClient/AwsRetryStrategy.php
@@ -13,6 +13,14 @@ use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
  */
 class AwsRetryStrategy extends GenericRetryStrategy
 {
+    public const DEFAULT_RETRY_STATUS_CODES = [0, 423, 425, 429, 500, 502, 503, 504, 507, 510];
+
+    // Override Symfony default options for a better integration of AWS servers.
+    public function __construct(array $statusCodes = self::DEFAULT_RETRY_STATUS_CODES, int $delayMs = 1000, float $multiplier = 2.0, int $maxDelayMs = 0, float $jitter = 0.1)
+    {
+        parent::__construct($statusCodes, $delayMs, $multiplier, $maxDelayMs, $jitter);
+    }
+
     public function shouldRetry(AsyncContext $context, ?string $responseContent, ?TransportExceptionInterface $exception): ?bool
     {
         if (parent::shouldRetry($context, $responseContent, $exception)) {


starred-gijs:
Thanks for your hard work! I deployed the default strategy 6 days ago (including that connection timeout fix), and until now 0 errors reported 👍  🙏 

This improvement will make it even more stable! 

> My personal opinion, I much prefer sending Sqs Message twice than failing sending one.

This could be addressed with deduplication if it becomes an issue?

nicolas-grekas:
Can't the retry strategy be aware of Sqs::SendMessage, Ses::SendEmail and Sns::Publish?

nicolas-grekas:
> Can't the retry strategy be aware of Sqs::SendMessage, Ses::SendEmail and Sns::Publish?

an idea: add an `extra.retryable` option managed by RetryableHttpClient to tell bypass the strategy and tell the client this request is retryable. Then use it for requests inside Sqs::SendMessage, Ses::SendEmail and Sns::Publish.

jderusse:
> > Can't the retry strategy be aware of Sqs::SendMessage, Ses::SendEmail and Sns::Publish?
> 
> an idea: add an `extra.retryable` option managed by RetryableHttpClient to bypass the strategy and tell the client that this request is retryable. Then use it for requests inside Sqs::SendMessage, Ses::SendEmail and Sns::Publish.

Yes, I started implementing exactly this feature: Some operations adds extra options to tell the RetryStrategy when the call is idempotent.

But when starting flagging the non-idempotent operation, I realized that just few operations are really idempotent. Moreover, I really want retrying those non-idempotent operation. 
ie. I don't care if a message is send twice (the Sqs adapter for symfony/messenger don't guarantee that messages are delivered exactly once...). 
Same for email: I prefer sending an email twice than missing a sent. Moreover user as no-way to know if email is really sent or not.

At the end, it sounded like to much work for nothing.

jderusse:
Friendly ping to reviewers :)

Nyholm (approved):
Thank you. 

I agree with you. Good work!

jderusse:
Thank you.

We may improve this if someone complains about retries on such methods

PR #828, opened 2020-10-23
Prevent Github Action running jobs on forks

jderusse:


chalasr:
Can you explain why we should skip running the CI on forks?

jderusse:
> Can you explain why we should skip running the CI on forks?

Why should we run CI on forks?.

Currently, when I push master to my own fork, it triggers build. Including sensitive operation like publishing the web-site. 
And given I've permission to push to master on async-aws, it doesn't fail (even when my version of master is not up to date) => https://github.com/jderusse/aws/runs/1297843921?check_suite_focus=true

Sheduled operations also run on forks. Including the "watch operation" that check if aws-sdk publish a new version. ie. https://github.com/async-aws-bot/aws/actions

Maybe I didn't understand how Github Action works with forks, I think that actions should only run on main Repo (which include jobs for PR, if it's what you are thinking)

jderusse:
hmm..shouldn't we just enable  "Allow local action only"?

Screenshot from 2020-10-23 14-35-09

chalasr:
I used to do some debugging on CI platforms (travis or appveyor) on my fork before opening a PR on the upstream repository for e.g. installing a package, reproducing a bug that I couldn't manage to trigger locally, or even submit a PR on the fork of someone else in order to complete a PR.
I think it'd be nice to keep the test part running (for everything except master :))

jderusse:
> I think it'd be nice to keep the test part running (for everything except master :))

Reverted change on "testing jobs".
But keep it on "deploy jobs"

chalasr (approved):

jderusse:
thanks you for the review

.github/workflows/watch.yml
```diff
@@ -6,6 +6,7 @@ on:
 
 jobs:
   generated-code:
+    if: github.repository == 'async-aws/aws'
```
bendavies, line 9: doesn't this break the `async-aws-bot` submitting PRs, since it runs on the fork?

jderusse: The Action run on the main repo (https://github.com/async-aws/aws/actions/runs/325544202)
But use the Bot's crendentials to push commits and open pull request on the Fork

bendavies: ah right, thanks

PR #832, opened 2020-10-29
[Core] Use trait and add test

Nyholm:
I refactored a few complicated lines into a trait and added a test for it. 

The functionality is the same.. Im not 100% we want this refactoring. 

Hm.. or maybe I should add a static method instead..


jderusse (approved):

Nyholm:
Is use a trait a good idea?

jderusse:
With 3 duplicate code, I thing it's a good idea to put this logic in a single place.
Having a "DateUtil" class with static method is not better.

I'm :+1: with a trait

Nyholm:
Thank you

PR #839, opened 2020-10-30
Update to support flysystem 2@rc

Nyholm:


src/Integration/Flysystem/S3/src/S3FilesystemV2.php
```diff
@@ -91,31 +93,36 @@ class S3FilesystemV2 implements FilesystemAdapter
...
-            $this->client->getObject(
+            return $this->client->objectExists(
                 [
                     'Bucket' => $this->bucket,
                     'Key' => $this->prefixer->prefixPath($path),
                 ]
-            )->resolve();
-
-            return true;
+            )->isSuccess();
```
Nyholm, line 123: This is non-blocking, right?

jderusse: nop, this just returns if the object exist. 
To be blocking (waiting until a final state (either success or failure)) use `$waiter->wait() && $waiter->isSuccess()`

jderusse (approved):
LGTM

src/Integration/Flysystem/S3/src/S3FilesystemV2.php
Nyholm: Cool cool. 

Then it is all good. It should not be blocking.

Nyholm:
Thank you

PR #840, opened 2020-10-31
Make sure we update branch alias after major version

Nyholm:
The current branch alias on SNS is 0.6. We failed to update that because our branch alias workflow did not look at the major version. 

It just saw that the minor version was bigger so it exited: https://github.com/async-aws/sns/runs/1331919079?check_suite_focus=true


Nyholm:
Note: I have not tested this =)

jderusse (approved):

PR #841, opened 2020-11-03
Make sure we are always throwing an exception from Core

Nyholm:


tests/Unit/ExceptionTest.php
```diff
@@ -0,0 +1,40 @@
...
+class ExceptionTest extends TestCase
+{
+    public function testBranchAlias()
+    {
+        $finder = new Finder();
+        $root = \dirname(__DIR__, 2);
+        $finder
+            ->files()
+            ->in($root . '/src/Core/src')
+            ->in($root . '/src/Service/*/src')
```
Nyholm, line 23: I dont check integrations

src/Core/src/Exception/UnexpectedValue.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Core\Exception;
+
+class UnexpectedValue extends \UnexpectedValueException implements Exception
```
jderusse, line 7: `UnexpectedValueException`

Nyholm: Are you sure? We usually dont use "Exception" suffix. 

Screenshot 2020-11-03 at 11 23 46

jderusse: ohh.. I was trapped by `RuntimeException` and `LogicException`... We are not consistent :cry:

Nyholm: True. But what we currently have sounds best. 

This is not super nice: 
```php
throw new Logic('Wrong');
```

Nyholm:
Let's not merge this until Symfony 5.2 is released.

jderusse (approved):

Nyholm:
Oh, no need to wait. 

The only new thing is `UnexpectedValue`, and no other packages are using that.

PR #844, opened 2020-11-04
Add some more introduction on the different clients

Nyholm:
I notice myself linking to the S3 page directly time to time. But that is not a "95% what you need to use it". 

I added some generic introduction for each client. They link to more detailed docs. 

Screenshot 2020-11-04 at 19 29 46


jderusse (approved):
Big :+1: for me

I really like it!

Nyholm:
Thank you

Issue #849, opened 2020-11-06
aws config and credential file.


alpha1125:
# The problem
When utilizing the aws/config and aws/credential files.

It seems asyncaws, is not picking up the config file settings. The config file specifies the region.  Both config and credential files were created by aws-cli, version 2.0.54

environment configurations posted here: https://gist.github.com/alpha1125/8ffe0a4c32f2d4368029492e8a4a911c

Other environment details:
Symfony 5.1.8

# Observed behaviour:
* When .env AWS_REGION is not set, asyncaws is not able to correctly make an sqs connection in another region. Ignoring the region setting in the config.
* When .env AWS_REGION is set, function as expected.

# Expected behaviour
Async-aws should respect the ./aws/config settings, which it does not seem to do so.

Nyholm:
Thank you for this issue. 

Have a look at this PR: https://github.com/async-aws/aws/pull/852
I think that will fix your issues.

PR #850, opened 2020-11-06
Make sure we merge profile data from multiple files.

Nyholm:
The order of the file matters as we never overwrite

This is related to #849 

It is actually needed for `WebIdentityProvider` if you have your region in the `~/.aws/config` and credentials in `~/.aws/credentials`


jderusse (approved):

Nyholm:
Thank you

PR #851, opened 2020-11-06
Extract region from credentials

Nyholm:
This will solve #849 

I've added a few TODOs, but I would be happy to get a review now.


jderusse (commented):
Even if it fix the solution, I don't really like using the credentials to store the region.

This use the IniFileProvider to store "configuration" data in the credentials, which is not it responsibility... :(

I'd rather use a `ConfigurationFactory` (that internally use `IniFileLoader`) to to generate (or enrich) the `Configuration` object.
Each property of the configuration would be `$injectedConfig ?? $envVariable ?? $iniFile`

This will also solve the almost same issue with `web_identity_token_file`.

Nyholm:
Hm.. sure. I can agree with that. 

#850 is still valid though.

PR #852, opened 2020-11-07
Read region into Configuration

### LINKED ISSUE #849: aws config and credential file.

Nyholm:
This will fix #849

It is better than #851

### DIFF
--- a/src/Core/src/Configuration.php
+++ b/src/Core/src/Configuration.php
@@ -110,6 +110,54 @@ final class Configuration
         return isset(self::AVAILABLE_OPTIONS[$optionName]);
     }
 
+    /**
+     * @psalm-return (
+     *     $name is
+     *       self::OPTION_REGION
+     *       |self::OPTION_DEBUG
+     *       |self::OPTION_PROFILE
+     *       |self::OPTION_ACCESS_KEY_ID
+     *       |self::OPTION_SECRET_ACCESS_KEY
+     *       |self::OPTION_SESSION_TOKEN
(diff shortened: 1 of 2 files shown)


jderusse (commented):
I like it much more !!!

What's about fetching `KEY_ROLE_ARN` and `KEY_WEB_IDENTITY_TOKEN_FILE` also, and removing the iniFileLoader from `WebIdentityProvider`.

Same for KeyId + Secret and deprecated IniFileProvider...

Nyholm:
Hm.. If I do that, then I've got no way to disable looking at the ini files (ie reconfigure the Credentials chain) .

jderusse (approved):
Thank you very much. Let's move forward with this first achievement.

About WebIdentityProvider and IniFileProvider, we'll see if we can leverage this in an other PR.

PR #869, opened 2020-11-21
Rename ParseResponse

Nyholm:
I think `ParseResponse` sounds like a command. I've considered the following alternative names: 

- UnparsableResponse
- ParseResponseException
- ResponseParsingException
- UnableToParseResponse

But I think `NonParsableResponse` is better if we dont want the `Exception` suffix. 

Let me know what you think is best.

### DIFF
--- a/src/Core/src/Exception/NonParsableResponse.php
+++ b/src/Core/src/Exception/NonParsableResponse.php
@@ -4,6 +4,6 @@ declare(strict_types=1);
 
 namespace AsyncAws\Core\Exception;
 
-class ParseResponse extends \RuntimeException implements Exception
+class NonParsableResponse extends \RuntimeException implements Exception
 {
 }
--- a/src/Core/src/Exception/NonParsableResponse.php
+++ b/src/Core/src/Exception/NonParsableResponse.php
@@ -1,9 +1,0 @@
-<?php
-
-declare(strict_types=1);
-
-namespace AsyncAws\Core\Exception;
-
-class NonParsableResponse extends \RuntimeException implements Exception
-{
-}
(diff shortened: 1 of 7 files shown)


src/Core/src/Exception/NonParsableResponse.php
```diff
@@ -4,6 +4,6 @@
 
 namespace AsyncAws\Core\Exception;
 
-class ParseResponse extends \RuntimeException implements Exception
```
jderusse, line 7: Tha't a BC break isn't it?
What if people use `catch (ParseResponse $e)`
unless the class is internal

jderusse (commented):
That's why I like suffixing names (FooException, FooInterface). 
There are less ambiguity...

src/Core/src/Exception/NonParsableResponse.php
Nyholm: Yes, it is a BC break. 
But we have not released this class yet. So this is our last chance =)

Nyholm:
> That's why I like suffixing names (FooException, FooInterface).

I thought you didn't like suffixing... hm.. I think it makes sense to drop suffix some times. Like `InvalidArgument`. But let's use `ParseResponseException` here.

jderusse:
> > That's why I like suffixing names (FooException, FooInterface).
> 
> I thought you didn't like suffixing... hm.. I think it makes sense to drop suffix some times. Like `InvalidArgument`. But let's use `ParseResponseException` here.

For consistency with other exception, I prefer the `NonParsableResponse`  name

Nyholm:
Hm.. I think the "consistency" ship has sailed...

Screenshot 2020-11-21 at 11 30 56

jderusse:
I would vote for `UnparsableResponse` which is closer to `UnsuportedRegion`, 
And to non-generic exception without suffix like `MissingDependency`

Nyholm:
Cool cool cool. 
The PR is updated

jderusse (approved):

Nyholm:
Thank you for the review and input

PR #870, opened 2020-11-21
Prepare patch releases

Nyholm:
It is mostly non-code changes in this repos. But it will be nice to reduce the diff. 

Also, these libraries (like all libraries in async-aws) are pretty stable.


jderusse (approved):

Nyholm:
Thank you

PR #872, opened 2020-11-22
Default empty DynamoDB M value to an empty object

stayallive:
Currently an empty `M` value fails with: `Unrecognized collection type class` which is correct since when you set `M` to `[]` it will leave the `$payload` empty.

It cannot however be an empty array since that encodes to an incorrect value it needs to be an object so it encodes to `{}` (json) correctly.

### DIFF
--- a/src/Service/DynamoDb/src/ValueObject/AttributeValue.php
+++ b/src/Service/DynamoDb/src/ValueObject/AttributeValue.php
@@ -192,8 +192,12 @@ final class AttributeValue
             }
         }
         if (null !== $v = $this->M) {
-            foreach ($v as $name => $v) {
-                $payload['M'][$name] = $v->requestBody();
+            if (empty($v)) {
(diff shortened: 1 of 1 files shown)


stayallive:
Hehe I very much underestimated the complexity of this project... my bad... I'm not sure if I want to invest the time in finding out how to get the generator to generate this change, any help would be appreciated!

Nyholm:
Yes. We need to modify the generator. Could you update this PR with a unit test?

Just make sure the tests are correct and I can try helping with the generator.

Nyholm:
Is it the same with L btw?

stayallive:
Hi @Nyholm, this `L` was already fixed in #668 so that one works as expected.

I see @jderusse has already done a great effort and from what I can tell that looks good!

Nyholm:
Could you still help out with a unit test for this bug?

PR #873, opened 2020-11-23
Fix empty map sent as array in Json requests

jderusse:
Fixes #872 

/cc @stayallive

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -21,6 +21,8 @@
   </coverage>
   <php>
     <ini name="error_reporting" value="-1"/>
+    <env name="AWS_SHARED_CREDENTIALS_FILE" value="/dev/null" />
+    <env name="AWS_CONFIG_FILE" value="/dev/null" />
   </php>
   <testsuites>
     <testsuite name="Test Suite">
--- a/src/Service/DynamoDb/tests/Unit/DynamoDbClientTest.php
+++ b/src/Service/DynamoDb/tests/Unit/DynamoDbClientTest.php
@@ -104,7 +104,7 @@ class DynamoDbClientTest extends TestCase
 
         $input = new DeleteItemInput([
             'TableName' => 'Foobar',
-
+            'Key' => ['ID' => ['S' => 'foobar']],
         ]);
         $result = $client->DeleteItem($input);
 
@@ -144,7 +144,7 @@ class DynamoDbClientTest extends TestCase
 
         $input = new GetItemInput([
             'TableName' => 'Foobar',
-
+            'Key' => ['ID' => ['S' => 'foobar']],
         ]);
         $result = $client->GetItem($input);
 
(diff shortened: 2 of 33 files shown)


phpunit.xml.dist
```diff
@@ -21,6 +21,8 @@
   </coverage>
   <php>
     <ini name="error_reporting" value="-1"/>
+    <env name="AWS_SHARED_CREDENTIALS_FILE" value="/dev/null" />
```
jderusse, line 24: Not related to that PR but tests are broken if your machine has a region configured in your Default AWS profile

stayallive:
Wow, this would have taken me quite a bit more time to figure out!

Thanks for taking the time to make this!

src/Service/DynamoDb/tests/Unit/DynamoDbClientTest.php
```diff
@@ -104,7 +104,7 @@ public function testDeleteItem(): void
 
         $input = new DeleteItemInput([
             'TableName' => 'Foobar',
-
+            'Key' => ['ID' => ['S' => 'foobar']],
```
Nyholm, line 107: Why is this needed?

Nyholm (commented):
Thank you for this PR. 

@stayalive, it is just a small change in the generator =)

We still need a test for #872.

src/Service/DynamoDb/tests/Unit/DynamoDbClientTest.php
jderusse: hehe... previous code had a "bug"

`Key` is required.. But because previously we initialized maps with an empty array (see https://github.com/async-aws/aws/pull/873/files#diff-324c911f8afbcd579bb76e189d1af7df134c2f519be2414f97d830934ec23588L119) it never triggered the test `if ($this->key === null) {throw Exception(...);}`

phpunit.xml.dist
Nyholm: That is correct. 

It is also the correct behavior. 

Another fix for this would be to set `AWS_REGION=us-east-1`.

src/Service/DynamoDb/tests/Unit/DynamoDbClientTest.php
Nyholm: Okey. Can you also update the changelogs with this?

Nyholm:
The PHP8 build failure is not related

Nyholm (approved):
I think this PR looks good. 

But I really really do want a test for the bug @stayalive mentioned.

Nyholm:
Thank you

stayallive:
Sorry for not being able to provide some tests to prove my issue (was a bit crazy at work) but I appreciate both of you for fixing my half-assed attempt to do so ❤️

Nyholm:
Thank you. 

Could you verify that this fix solve your problem? If so, I'll tag a new release.

stayallive:
Yes, just confirmed that `"async-aws/dynamo-db": "dev-master#a4808ec37d9da540f096465717324c8b2d393bdb"` works as intended 💪

PR #880, opened 2020-11-25
Fix endpoint without region

### LINKED ISSUE #879: Missing regions eu-west-2 / eu-west-3 in src/Service/Ses/src/SesClient.php
Hi, 

async-aws/ses does not work for region eu-west-3, this case is missing in file `src/Service/Ses/src/SesClient.php`

I try to figure out how this code is generated to make a PR but I didn't find it …

jderusse:
After checking #879, I discovered that the official AWS-SDK always fallback to the default configuration (when it exists) even when the region is not listed in the "endpoints.json" file.

This will fix #879


Nyholm:
Awesome. Thank you @jderusse <3

I pushed some minor changes to the tests.

PR #884, opened 2020-11-30
Fix signature of method

jderusse:
These options does not have "default" value, and can be null


Nyholm:
Oh, my bad.. I misread this annotations when I added these.

PR #885, opened 2020-11-30
Fixed DynamoDB tests

Nyholm:
I dont know how this have worked before...

### DIFF
--- a/src/Service/DynamoDb/tests/Integration/DynamoDbClientTest.php
+++ b/src/Service/DynamoDb/tests/Integration/DynamoDbClientTest.php
@@ -381,7 +381,7 @@ class DynamoDbClientTest extends TestCase
         $client = $this->getClient();
 
         $input = new DescribeTableInput([
-            'TableName' => $this->tableName.'_non_exist',
+            'TableName' => $this->tableName,
         ]);
 
         self::assertFalse($client->tableNotExists($input)->isSuccess());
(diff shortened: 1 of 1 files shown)


src/Service/DynamoDb/tests/Integration/DynamoDbClientTest.php
```diff
@@ -380,7 +381,7 @@ public function testTableNotExists(): void
         $client = $this->getClient();
 
         $input = new DescribeTableInput([
-            'TableName' => $this->tableName,
+            'TableName' => $this->tableName.'_non_exist',
```
jderusse, line 384: This is wrong, the test asserts that `tableNotExists($input)` is false. which mean the table should exists

src/Service/DynamoDb/tests/Integration/DynamoDbClientTest.php
```diff
@@ -414,6 +415,6 @@ private function getClient(): DynamoDbClient
 
         return $this->client = new DynamoDbClient([
             'endpoint' => 'http://localhost:8000',
-        ], new Credentials('aws_id', 'aws_secret'));
+        ], new Credentials('aws_id', 'aws_secret'), $httpClient = HttpClient::create());
```
jderusse, line 418: I suggest to keep the default client to see if there is something wrong with the retryiableHttpClient

src/Service/DynamoDb/tests/Integration/DynamoDbClientTest.php
Nyholm: You are correct. I didnt see that until now. 

Thank you

src/Service/DynamoDb/tests/Integration/DynamoDbClientTest.php
Nyholm: I agree. This is just me debugging. I failed to start the docker locally....

PR #890, opened 2020-12-02
fix: edge case where aws config file could be a directory

shouze:
I have encountered this issue on github actions

This is weird but I an error with file_get_contents failing.
image

### DIFF
--- a/src/Core/src/Credentials/IniFileLoader.php
+++ b/src/Core/src/Credentials/IniFileLoader.php
@@ -48,7 +48,7 @@ final class IniFileLoader
                 $homeDir = $homeDir ?? $this->getHomeDir();
                 $filepath = $homeDir . \substr($filepath, 1);
             }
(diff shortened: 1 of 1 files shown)


jderusse (approved):
Good catch

shouze:
> Good catch

Anyway you could argue: why ~/.aws/credentials is a directory for you? I have to admit this is a very strange edge case, some kind of monster that we've created ourselves :japanese_ogre:.

jderusse:
> > Good catch
> 
> Anyway you could argue: why ~/.aws/credentials is a directory for you? I have to admit this is a very strange edge case, some kind of monster that we've created ourselves .

This kind of things happens when you mount a docker volume to a file and the file does not exists, docker creates a directory

shouze:
> This kind of things happens when you mount a docker volume to a file and the file does not exists, docker creates a directory

oh my, you got me! :wink:

PR #892, opened 2020-12-04
Ignore cache errors

### LINKED ISSUE #891: Issue with Symfony Cache Adapter
Hi, thanks for the awesome package(s)!

I found a small thing that's worth mentioning.
Normally, when using symfony cache, and there is an error, it doesn't break the app, it just gives a warning log that reading or writing a key has failed. Now when I use a broken cache adapter (e.g. unreachable redis) with async-aws credential cache, then the exception propagates all the way and causes a hard break instead of a warning.

I believe it would be worth following a similar pattern and make the cache optional – allow it to error without breaking the flow.

Nyholm:
This will fix #891 

It feels weird just muting exceptions like this. However, both the Symfony cache and PSR-6 cache says that the only exception that should be thrown is a InvalidArgumentException when we have an invalid key. We are in very much control of the keys we are using and we make sure that we only use valid characters. 

The reason we should mute this is because the application should work even if the cache is broken. 

Ideally we would log this... but.. That seams overkill, right?


jderusse (approved):
Thank you for working into this

src/Core/src/Credentials/PsrCacheProvider.php
```diff
@@ -27,6 +28,18 @@ public function __construct(CredentialProvider $decorated, CacheItemPoolInterfac
     }
 
     public function getCredentials(Configuration $configuration): ?Credentials
+    {
+        try {
+            return $this->getFromCache($configuration);
+        } catch (CacheException $e) {
```
TamasSzigeti, line 34: This behaviour is already implemented inside Symfony Cache, if it was fine to rely on `\Symfony\Contracts\Cache\CacheInterface` instead, then the logic doesn't need to be repeated. But I assume it is intentional to rely on the PSR interface instead?

TamasSzigeti (approved):
I've tested and it does solve #892, thanking muchly.
I don't think logging a warning would be an overkill though…

src/Core/src/Credentials/PsrCacheProvider.php
TamasSzigeti: For reference: https://github.com/symfony/cache/blob/5.x/Traits/AbstractAdapterTrait.php#L110

Nyholm: Yeah. You are correct. For a Symfony PSR6 cache, this will never happen. 

But there is many more PSR6 cache libraries. I need to look at the specifications and not the implementations.

phpstan.neon.dist
```diff
@@ -19,3 +20,4 @@ parameters:
         - '#^PHPDoc tag @param has invalid value#'
         - '#^Parameter \#1 \$input of class AsyncAws\\[^\\]+\\ValueObject\\[^ ]+ constructor expects (array\(\)\|)?array\([^\)]+\), array\([^\)]+\) given\.#'
         - '#Method AsyncAws\\[^:]+::[^(]+\(\) should return array<[^>]+> but returns array<int(\|string)?, string>\.$#'
+        - '#PHPDoc tag @throws with type Psr\\Cache\\CacheException is not subtype of Throwable$#'
```
Nyholm, line 23: It is safe to catch a non-throwable. But you cannot throw it.

https://3v4l.org/D7nk9

jderusse (approved):

Nyholm:
The psalm error is a false positive and unrelated.

PR #894, opened 2020-12-08
Fix website generation

jderusse:
new version of highlight does not export the library

Here is the content of the file
```js
// node_modules/highlight.js/lib/highlight.js

// This file has been deprecated in favor of core.js
var hljs = require('./core');
```


Nyholm (approved):
Oh. Thank you

PR #896, opened 2020-12-10
Exclude 5.2.0 that bug with our RetryStrategy

jderusse:
Should be fixed in 5.2.1 by https://github.com/symfony/symfony/pull/39228


Nyholm (approved):
Thank you

PR #898, opened 2020-12-10
Fix type error

mnapoli:
The profile name seems to be cast to an integer if it looks like an integer.

That's the case for example with a `7777` profile, I got the following error:

```
PHP Fatal error:  Uncaught TypeError: preg_replace(): Argument #3 ($subject) must be of type array|string, int given in .../vendor/async-aws/core/src/Credentials/IniFileLoader.php:56
Stack trace:
#0 .../vendor/async-aws/core/src/Credentials/IniFileLoader.php(56): preg_replace('/^profile /', '', 7777)
```


jderusse (commented):
Good Catch

Nyholm (approved):
Thanky ou

PR #899, opened 2020-12-11
[Core] Release 1.7.1

Nyholm:


Nyholm:
The BC break checks warns of 3 BC breaks: https://github.com/async-aws/core/runs/1536145268?check_suite_focus=true

They are all related to changes in Symfony. We should not care, right?

jderusse (approved):

jderusse:
Agreed, this is not related to us

Nyholm:
Thank you.

PR #901, opened 2020-12-11
Clear realpath cache when reading the token from file

Jean85:
This fixes #900 by applying the same fix of the official SDK in aws/aws-sdk-php#2043

PR #903, opened 2020-12-15
Fix wrong check from #901

### LINKED ISSUE #900: Intermittent failure while loading token in EKS with service accounts
I'm running a Symfony app in EKS, and I'm using SQS with Messenger + this lib.

I'm having small clusters of failures at long intervals with 403 responses from the SQS APIs:
```
HTTP 403 returned for "https://sqs.eu-central-1.amazonaws.com/".

Code:    AccessDenied
Message: Access to the resource https://sqs.eu-central-1.amazonaws.com/ is denied.
Type:    Sender
Detail:
```
I've dig into the logs, and I've found this warning:
```
php.WARNING: Warning: file_get_contents(/var/run/secrets/eks.amazonaws.com/serviceaccount/token): failed to open stream: No such file or directory {"exception":"[object] (ErrorException(code: 0): Warning: file_get_contents(/var/run/secrets/eks.amazonaws.com/serviceaccount/token): failed to open stream: No such file or directory at /var/www/my-project/vendor/async-aws/core/src/Credentials/WebIdentityProvider.php:92)"} []
```
So I've traced down the issue to something that happened on the official SDK too: https://github.com/aws/aws-sdk-php/issues/2014

Basically, it's a combination of:
 * tokens are short lived
 * EKS replaces tokens on-the-fly
 * PHP's realpath cache gets in the way...

... so you get brief intervals in which you try to read the old, now missing token.

In the offical SDK, the issue was mitigated by clearing the cache in a precise and ordered way of the following paths:
```
/var/run/secrets/eks.amazonaws.com/serviceaccount/..data/token
/var/run/secrets/eks.amazonaws.com/serviceaccount/..data
/var/run/secrets/eks.amazonaws.com/serviceaccount/token
```
in this PR: https://github.com/aws/aws-sdk-php/pull/2043

Can we apply the same logic here too?

Jean85:
This fixes #900, again.

Sorry for getting it wrong the first time around.

### DIFF
--- a/src/Core/src/Credentials/WebIdentityProvider.php
+++ b/src/Core/src/Credentials/WebIdentityProvider.php
@@ -128,13 +128,17 @@ final class WebIdentityProvider implements CredentialProvider
      */
     private function getTokenFileContent(string $tokenFile): string
     {
-        if (!file_exists($tokenFile)) {
-            $tokenDir = \dirname($tokenFile);
-            $tokenLink = readlink($tokenFile);
-            clearstatcache(true, $tokenDir . \DIRECTORY_SEPARATOR . $tokenLink);
-            clearstatcache(true, $tokenDir . \DIRECTORY_SEPARATOR . \dirname($tokenLink));
-            clearstatcache(true, $tokenFile);
+        $token = @file_get_contents($tokenFile);
+
+        if (false !== $token) {
+            return $token;
         }
+
+        $tokenDir = \dirname($tokenFile);
+        $tokenLink = readlink($tokenFile);
+        clearstatcache(true, $tokenDir . \DIRECTORY_SEPARATOR . $tokenLink);
+        clearstatcache(true, $tokenDir . \DIRECTORY_SEPARATOR . \dirname($tokenLink));
+        clearstatcache(true, $tokenFile);
 
         if (false === $token = file_get_contents($tokenFile)) {
             throw new RuntimeException('Failed to read data');


src/Core/src/Credentials/WebIdentityProvider.php
```diff
@@ -128,14 +128,18 @@ private function getCredentialsFromRole(string $roleArn, string $tokenFile, ?str
      */
     private function getTokenFileContent(string $tokenFile): string
     {
-        if (!file_exists($tokenFile)) {
-            $tokenDir = \dirname($tokenFile);
-            $tokenLink = readlink($tokenFile);
-            clearstatcache(true, $tokenDir . \DIRECTORY_SEPARATOR . $tokenLink);
-            clearstatcache(true, $tokenDir . \DIRECTORY_SEPARATOR . \dirname($tokenLink));
-            clearstatcache(true, $tokenFile);
+        $token = @file_get_contents($tokenFile);
```
Jean85, line 131: `@` is to suppress warning in the exact case of hitting the realpath cache with the wrong valued.

Nyholm (approved):
Im happy you verified and checked.

src/Core/src/Credentials/WebIdentityProvider.php
```diff
@@ -128,14 +128,18 @@ private function getCredentialsFromRole(string $roleArn, string $tokenFile, ?str
...
+            return $token;
         }
 
+        $tokenDir = \dirname($tokenFile);
+        $tokenLink = readlink($tokenFile);
+        clearstatcache(true, $tokenDir . \DIRECTORY_SEPARATOR . $tokenLink);
+        clearstatcache(true, $tokenDir . \DIRECTORY_SEPARATOR . \dirname($tokenLink));
+        clearstatcache(true, $tokenFile);
+
         if (false === $token = file_get_contents($tokenFile)) {
```
jderusse, line 143: shouldn't it be `@file_get_contents` here?

jderusse (approved):
Huge thank you for the investigation and the quick fix

src/Core/src/Credentials/WebIdentityProvider.php
Jean85: I preferred to have the warning here anyway, since it helped me a lot debugging this regression.

IIRC, we get here only if we get the token file path from the env variable, so we shouldn't arrive to this point if there's no token to start with.

Nyholm:
Cool cool cool. Thank you for the fix. 

Can you confirm that this fix is working for you?

Jean85:
I've deployed it in prod. I'll report back in a few days to be sure, since it didn't happen continuously.

Nyholm:
How does it look? Did this PR fix the issue?

Jean85:
Confirming here too as in #906: this works! :+1:

PR #910, opened 2020-12-29
Add documentation about declaring multiple clients

### LINKED ISSUE #908: [Symfony] Multiple AWS clients
The Symfony bundle allows me to configure my AWS clients like this:

```YAML
async_aws:
    clients:
        s3:
            config:
                # insert S3 configuration here
```

This format assumes that I use the same AWS credentials for all S3 buckets I want to access. However, in my application, this is not the case. I need to access two different S3 buckets in different regions with different credentials. Right now, I need to configure one of the clients manually.

It would be awesome, if the bundle would allow me to configure multiple AWS clients.

jderusse:
Fixes #908 ping @derrabus

### DIFF
--- a/docs/integration/symfony-bundle.md
+++ b/docs/integration/symfony-bundle.md
@@ -52,6 +52,10 @@ For a complete reference of the configuration please run:
 php bin/console config:dump-reference async_aws
 ```
 
+> **Note**: Multiple clients with different configuration for the same service
+> can be created by providing a unique name and defining the service in the
+> `type` sub-property. See client `delayed` in above example.
+
 ## Using SSM to store secrets
 
 Since version 4.4, Symfony provides [secrets management](https://symfony.com/doc/current/configuration/secrets.html).


derrabus (approved):

docs/integration/symfony-bundle.md
```diff
@@ -43,15 +43,17 @@ The config above will create the following services:
 | `async_aws.client.ses`     | AsyncAws\Ses\SesClient
 | `async_aws.client.sqs`     | AsyncAws\Sqs\SqsClient
 | `async_aws.client.delayed` | AsyncAws\Sqs\SqsClient $delayed
-| `async_aws.client.s3`      | AsyncAws\S3\S3Client
-| `async_aws.client.sts`     | AsyncAws\Core\Sts\StsClient
```
Nyholm, line 47: These should not be removed. 

> This example assume you have installed core, ses, sqs and s3

Nyholm (changes requested):
Thank you

docs/integration/symfony-bundle.md
jderusse: Oh right. I forgot about this behavior.

Nyholm (approved):

Nyholm:
Thank you

PR #911, opened 2020-12-29
Deactivate chunked body in S3

### LINKED ISSUE #874: [S3] Disable transferring payload in multiple chunks
Upload with multiple chunks realised in https://github.com/async-aws/aws/pull/226 not supported by S3-compatible services. For example, S3 middleware for Openstack Swift does [not support this](https://github.com/openstack/swift/blob/master/swift/common/middleware/s3api/s3request.py#L819).

We need a flag to switch to using chunks.

jderusse:
Fixes #874

### DIFF
--- a/docs/clients/s3.md
+++ b/docs/clients/s3.md
@@ -117,6 +117,18 @@ use AsyncAws\S3\S3Client;
 $s3 = new S3Client(['pathStyleEndpoint' => true]);
 ```
 
+## Chunked body
+
+When sending data to AWS endpoints, AsyncAws split the content in multiple
+chunks. To change this behavior, set `sendChunkedBody` parameter to `false` when
+initializing the client.
+
+```php
+use AsyncAws\S3\S3Client;
+
+$s3 = new S3Client(['sendChunkedBody' => false]);
+```
+
 ## Non-AWS S3 endpoints
 
 To use the `S3Client` with example Digital Oceans' Spaces, you need to initialize
--- a/docs/clients/s3.md
+++ b/docs/clients/s3.md
@@ -120,8 +120,13 @@ $s3 = new S3Client(['pathStyleEndpoint' => true]);
 ## Chunked body
 
 When sending data to AWS endpoints, AsyncAws split the content in multiple
-chunks. To change this behavior, set `sendChunkedBody` parameter to `false` when
-initializing the client.
+chunks. This improves UX by avoiding reading the file twice (required
+[to compute the signature](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html)
+) which could be a performance issue when file is really big, or the uploaded
+content is not a file (ie. streamed from an HTTP request). But some 3rd party
+services like Openstack Swift, pretending being "S3-compatible" does not
+support chunked body. To change this behavior, set `sendChunkedBody` parameter
+to `false` when initializing the client.
 
 ```php
 use AsyncAws\S3\S3Client;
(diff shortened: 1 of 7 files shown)


docs/clients/s3.md
```diff
@@ -117,6 +117,18 @@ use AsyncAws\S3\S3Client;
 $s3 = new S3Client(['pathStyleEndpoint' => true]);
 ```
 
+## Chunked body
+
+When sending data to AWS endpoints, AsyncAws split the content in multiple
+chunks. To change this behavior, set `sendChunkedBody` parameter to `false` when
```
Nyholm, line 123: Maybe explain why we are splitting it into chunks. 

Maybe also mention why you would like to disable it.

jderusse: doc updated

src/Service/S3/src/Signer/SignerV4ForS3.php
```diff
@@ -86,8 +95,10 @@ protected function convertBodyToStream(SigningContext $context): void
         }
 
         // no need to stream small body. It's simple to convert it to string directly
-        if ($contentLength < self::CHUNK_SIZE) {
-            $request->setBody(StringStream::create($body));
+        if ($contentLength < self::CHUNK_SIZE || !$this->sendChunkedBody) {
+            if ($body instanceof ReadOnceResultStream) {
+                $request->setBody(RewindableStream::create($body));
+            }
```
Nyholm, line 101: What if the body is not an instance of ReadOnceResultStream?

jderusse: `$body` is initialized from `$body = $request->getBody()`.
If it's not an instance of `ReadOnceResultStream`, then we are good, we can rewind the stream.

src/Service/S3/src/Signer/SignerV4ForS3.php
```diff
@@ -36,6 +36,15 @@ class SignerV4ForS3 extends SignerV4
         'PutObjectLockConfiguration' => true,
     ];
 
+    private $sendChunkedBody;
+
+    public function __construct(string $scopeName, string $region, bool $sendChunkedBody = true)
```
Nyholm, line 41: Does it make sense to pass an array with options here instead? Or is it just me that is defensive?

src/Service/S3/src/Signer/SignerV4ForS3.php
Nyholm: I fail to understand why though.. Before, we always converted to a `StringStream` if the body was small, but now we dont anymore. Was it a misstake before?

Nyholm (commented):
Sorry if Im extra slow on this PR.

src/Service/S3/src/Signer/SignerV4ForS3.php
jderusse: Yes, I believe the conversation were not needed.

If I remember well, when I implemented the conversion to string, the `ReadOnceResultStream` did not exists and stream have to be converted into string at some point.

Now, thanks to the RewindableStream, the signature is able to works with all cases, but I forgot to revert upgrade this statement.

src/Service/S3/src/Signer/SignerV4ForS3.php
jderusse: updated

Nyholm (approved):
Can you see why the tests are failing?

jderusse:
last test fails because of https://github.com/doctrine/annotations/issues/398

jderusse:
Thank you for the review

Nyholm:
Awesome. Thank you!

PR #916, opened 2021-01-16
[S3] Release 1.7.0

Nyholm:
Release core 1.8 and S3 1.7. 

Note that S3 1.7 has an updated composer.json.

### DIFF
--- a/src/Service/S3/composer.json
+++ b/src/Service/S3/composer.json
@@ -16,7 +16,7 @@
         "ext-dom": "*",
         "ext-filter": "*",
         "ext-hash": "*",
-        "async-aws/core": "^1.2"
+        "async-aws/core": "^1.8"
     },
     "extra": {
         "branch-alias": {
--- a/src/Service/S3/composer.json
+++ b/src/Service/S3/composer.json
@@ -16,7 +16,7 @@
         "ext-dom": "*",
         "ext-filter": "*",
         "ext-hash": "*",
-        "async-aws/core": "^1.8"
+        "async-aws/core": "^1.2"
     },
     "extra": {
         "branch-alias": {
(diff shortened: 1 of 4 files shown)


src/Service/S3/composer.json
```diff
@@ -16,7 +16,7 @@
         "ext-dom": "*",
         "ext-filter": "*",
         "ext-hash": "*",
-        "async-aws/core": "^1.2"
+        "async-aws/core": "^1.8"
```
jderusse, line 19: Really? S3 is not compatible with version above 1.8? Why prefer-lowest test never failed?

Nyholm:
We added a new config option. That option cannot be set in versions below 1.8.

jderusse:
> We added a new config option. That option cannot be set in versions below 1.8.
> […](#)
> On 16 Jan 2021, at 12:52, Jérémy Derussé ***@***.***> wrote: ﻿ @jderusse commented on this pull request. In src/Service/S3/composer.json: > @@ -16,7 +16,7 @@ "ext-dom": "*", "ext-filter": "*", "ext-hash": "*", - "async-aws/core": "^1.2" + "async-aws/core": "^1.8" Really? S3 is not compatible with version above 1.8? Why prefer-lowest test never failed? — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Indeed, but the code is compatible with version that does not support this option (see check here https://github.com/async-aws/aws/blob/master/src/Service/S3/src/S3Client.php#L946)

I don't have strong opinion about compatibility layer versus version bump in our own dependencies.
If we bump the version, I suggest to remove the check that will not be neded anymore

Nyholm:
You are correct. I missed that. 

> If we bump the version, I suggest to remove the check that will not be needed anymore

Correct. I will revert the changes to composer.json

jderusse (approved):

Nyholm:
Thank you

PR #921, opened 2021-01-19
Generate Exceptions

jderusse:
This PR adds generation of Exceptions objects.

I only generate code for 3 methods to see the result. Once PR reviewed, I'll generate everything.

* ResultGenerator and ExceptionGenerator share the same code to populate responses
* I had to `lcfirst` property name to avoid conflict in `Exception::$Message`. This rule applied also for all objects because, the same CodeGenerator is used (cf previous point);
* I added a new key `RequestContext::exceptionMapping`, therefor all clients must bump the version of core to accept this new key. For the future, I wonder if unknown extra key in `RequestContext` shouldn't be silently ignored to ease BC
* Removed final from ClientException and ServerExeptoin
* `@throws` doc bloc is setup on Client's method. Which is technically wrong (because exception is triggered on lazy loading or destruct) but is the best place for this documentation IMHO.

TODO:
* [x] Regenerate ALL classes
* [x] Fix composer version to require latest `async-aws/core` version (required for RequestContext::exceptionMapping)

### DIFF
--- a/src/CodeGenerator/src/Generator/CodeGenerator/PopulatorGenerator.php
+++ b/src/CodeGenerator/src/Generator/CodeGenerator/PopulatorGenerator.php
@@ -1,0 +1,262 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\CodeGenerator\Generator\CodeGenerator;
+
+use AsyncAws\CodeGenerator\Definition\ListShape;
+use AsyncAws\CodeGenerator\Definition\MapShape;
+use AsyncAws\CodeGenerator\Definition\Operation;
+use AsyncAws\CodeGenerator\Definition\StructureShape;
+use AsyncAws\CodeGenerator\File\FileWriter;
+use AsyncAws\CodeGenerator\Generator\EnumGenerator;
+use AsyncAws\CodeGenerator\Generator\GeneratorHelper;
+use AsyncAws\CodeGenerator\Generator\Naming\NamespaceRegistry;
+use AsyncAws\CodeGenerator\Generator\ObjectGenerator;
+use AsyncAws\CodeGenerator\Generator\PhpGenerator\ClassBuilder;
+use AsyncAws\CodeGenerator\Generator\PhpGenerator\ClassRegistry;
+use AsyncAws\CodeGenerator\Generator\ResponseParser\ParserProvider;
+use AsyncAws\Core\Response;
+use AsyncAws\Core\Stream\ResponseBodyStream;
+use AsyncAws\Core\Stream\ResultStream;
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+/**
+ * Generate method and properties to populate object from AWS response.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ * @author Jérémy Derussé <jeremy@derusse.com>
+ *
+ * @internal
+ */
+class PopulatorGenerator
+{
+    /**
+     * @var ObjectGenerator
+     */
+    private $objectGenerator;
+
+    /**
+     * @var EnumGenerator
+     */
+    private $enumGenerator;
+
+    /**
+     * @var TypeGenerator
+     */
+    private $typeGenerator;
+
+    /**
+     * @var ParserProvider
+     */
+    private $parserProvider;
+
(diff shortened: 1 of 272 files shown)


Nyholm (commented):
This looks great. Thank you. Just some small comments.

src/Core/src/Response.php
```diff
@@ -380,6 +386,14 @@ private function getResolveStatus(): bool
             [$class, $args] = $this->resolveResult;
             /** @psalm-suppress PropertyTypeCoercion */
             $this->resolveResult = new $class(...$args);
+            if ($this->resolveResult instanceof HttpException) {
+                $awsCode = $this->resolveResult->getAwsCode();
+                if (isset($this->exceptionMapping[$awsCode])) {
+                    $class = $this->exceptionMapping[$awsCode];
+                    /** @psalm-suppress PropertyTypeCoercion */
+                    $this->resolveResult = new $class($this->httpResponse);
```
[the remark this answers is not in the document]
jderusse: yeah this is how it works. I needed the original exception to extract `awsCode`.
I think this part could be improved to removed logic from the `ExceptionTrait`, but that should be part of another PR IMHO

src/Service/CloudFront/src/CloudFrontClient.php
```diff
@@ -21,11 +28,27 @@ class CloudFrontClient extends AbstractApi
...
+     * @throws TooManyInvalidationsInProgressException
+     * @throws InconsistentQuantitiesException
      */
     public function createInvalidation($input): CreateInvalidationResult
     {
         $input = CreateInvalidationRequest::create($input);
-        $response = $this->getResponse($input->request(), new RequestContext(['operation' => 'CreateInvalidation2019_03_26', 'region' => $input->getRegion()]));
+        $response = $this->getResponse($input->request(), new RequestContext(['operation' => 'CreateInvalidation2019_03_26', 'region' => $input->getRegion(), 'exceptionMapping' => [
+            'AccessDenied' => 'AsyncAws\\CloudFront\\Exception\\AccessDeniedException',
+            'MissingBody' => 'AsyncAws\\CloudFront\\Exception\\MissingBodyException',
```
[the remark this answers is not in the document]
jderusse: done

Nyholm:
Could I ask for a rebase please?

src/Service/CloudFront/src/Exception/BatchTooLargeException.php
```diff
@@ -0,0 +1,30 @@
...
+        $this->populateResult($response);
+    }
+
+    protected function populateResult(ResponseInterface $response): void
+    {
+        $data = new \SimpleXMLElement($response->getContent(false));
+        if (0 < $data->Error->count()) {
+            $data = $data->Error;
+        }
+        if (null !== $v = (($v = $data->message) ? (string) $v : null)) {
```
jderusse, line 26: This is why we need lowercase property @Nyholm

src/Service/CloudFront/src/Exception/AccessDeniedException.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+final class AccessDeniedException extends ClientException
+{
+    public function __construct(ResponseInterface $response, ?AwsError $awsError)
+    {
+        parent::__construct($response, $awsError);
+        $this->populateResult($response);
+    }
+
+    protected function populateResult(ResponseInterface $response): void
```
Nyholm, line 20: This can be private

Nyholm (approved):
Excellent. 

It has been a long way here =)
Thank you

src/CodeGenerator/src/Generator/CodeGenerator/PopulatorGenerator.php
```diff
@@ -0,0 +1,261 @@
...
+            }
+            $classBuilder->setMethods($parserResult->getExtraMethods());
+        }
+        if (empty(trim($body))) {
+            return;
+        }
+
+        $method = $classBuilder->addMethod('populateResult')
+            ->setReturnType('void')
+            ->setProtected()
```
Nyholm, line 251: This can be private

Nyholm (approved):
Im happy with this when the tests are green

jderusse:
~all green. The last fail test is pref-lowest. Because each component need something provided in that PR. 
I've bumper their composer.json to `core: 1.9` (the next version) but it won't work until we merge this PR.
If we want an all-green PR, we have to split it in 2:
 - Add RequestContext::exceptionMapping + remove final from Client and Server Exception
 - Rebase this PR.

What do you prefer?

Nyholm:
Na, let’s merge now and fix any problem in master :)

PR #924, opened 2021-01-20
Add documentation on top of classes

jderusse:
This is part of #921

Its adds documentation on top of classes (enum, object value, input, result).
It also handle complexe cases like src/Service/DynamoDb/src/Enum/AttributeAction.php ;-)

### DIFF
--- a/src/CodeGenerator/src/Generator/GeneratorHelper.php
+++ b/src/CodeGenerator/src/Generator/GeneratorHelper.php
@@ -36,8 +36,6 @@ class GeneratorHelper
         ]);
         $s = preg_replace('/\n*<(\/?(note|important|ul|li))>\n*/', "\n<\$1>\n", $s);
         $s = preg_replace('/\n+/', "\n", $s);
-//        $s = preg_replace('/\n?<(note|important|ul|li)>/', "\n<\$1>", $s);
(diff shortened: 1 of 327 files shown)


src/CodeGenerator/src/Generator/GeneratorHelper.php
```diff
@@ -30,6 +34,10 @@ public static function parseDocumentation(string $documentation): string
             '<b>' => '**',
             '</b>' => '**',
         ]);
+        $s = preg_replace('/\n*<(\/?(note|important|ul|li))>\n*/', "\n<\$1>\n", $s);
+        $s = preg_replace('/\n+/', "\n", $s);
+//        $s = preg_replace('/\n?<(note|important|ul|li)>/', "\n<\$1>", $s);
+//        $s = preg_replace('/\n*<\/li>\n/', "</li>\n", $s);
```
Nyholm, line 40: Is this debug code?

Nyholm (approved):
This is great. Thank you

src/CodeGenerator/src/Generator/GeneratorHelper.php
jderusse: yep, sorry about that

PR #925, opened 2021-01-20
Uncapitalize objects's property

jderusse:
This PR converts all properties to camelCase

The API (with end-user) does not changes, the input still respect the no-convension AWS randomCase and match the documentation.

BUT the internal private properties used inside our object now respect the camelCase convension.

This PR is all about replacing `$name` by `lcfirst($name)` in generator. Except in MockFactory (that inject property inside the object), that contains a check with old-naming) to be backward compatible

note: Deprecation message and Exeption message still use the original AWS Name

note: using camelCase is also a requirement for #921


src/Service/CognitoIdentityProvider/src/Input/SetUserMFAPreferenceRequest.php
```diff
@@ -16,14 +16,14 @@ final class SetUserMFAPreferenceRequest extends Input
      *
      * @var SMSMfaSettingsType|null
      */
-    private $SMSMfaSettings;
+    private $sMSMfaSettings;
```
Nyholm, line 19: Properties like these are not super pretty.

src/Core/src/Test/ResultMockFactory.php
```diff
@@ -89,7 +89,11 @@ public static function create(string $class, array $data = [])
 
         $initializedProperty->setValue($object, true);
         foreach ($data as $propertyName => $propertyValue) {
-            $property = $reflectionClass->getProperty($propertyName);
+            if ($reflectionClass->hasProperty($propertyName)) {
+                $property = $reflectionClass->getProperty($propertyName);
+            } else {
+                $property = $reflectionClass->getProperty(lcfirst($propertyName));
```
Nyholm, line 95: Could you add a comment before this line that this is legacy code?

Nyholm (commented):
Thank you. 
I don't yet understand why this is a requirement for #921. I'll review that PR and it may be clear.

src/Service/CognitoIdentityProvider/src/Input/SetUserMFAPreferenceRequest.php
jderusse: You're right. I've submitted a patch to fix that

src/Core/src/Test/ResultMockFactory.php
jderusse: done

src/Service/S3/src/Result/CreateMultipartUploadOutput.php
```diff
@@ -125,28 +125,28 @@ public function getSSECustomerAlgorithm(): ?string
...
-        return $this->sSECustomerKeyMD5;
+        return $this->sseCustomerKeyMD5;
     }
 
     public function getSSEKMSEncryptionContext(): ?string
     {
         $this->initialize();
 
-        return $this->sSEKMSEncryptionContext;
+        return $this->sseKmsEncryptionContext;
```
Nyholm: Do we care about this?
We use one type of case on the methods, and another on the property.

Nyholm: FYI: https://3v4l.org/t5BlU

jderusse: What do you mean?
- just `sseKMS` we just focus on first word only
- rename `getSSEKMS` into `getSseKms` (`getARN` into `getArn`) we apply the same naming rule also on methods

note: renaming method is Backward compatible as you pointed it in your link.

Nyholm: I meant that the function is called `getSSEKMSEncryptionContext()` and the property is `sseKmsEncryptionContext`. I like the property name. Should we rename the function too?

jderusse: no strong opinion about this, but it would be ~easy to fix (the method already exists).

By the way, I reverted it, to only camelCase the first word.
 But if we choose to also camelCase `KMS` here I should, for consistency, refactor the function to also fix words that are in the middle of the name (ie `sseCustomerKeyMD5`)

Tell me what you prefer.

jderusse: So I pushed a new commit that also normalize Methods name and class name

This fixes things like:
```php
public function getFOOBar() {
  return $this->fooBar;
}
```

jderusse: hmm.. changed my mind, revert the normalization on class name (this is a BC break because of autoloading)
And keep Method and Property only

Nyholm:
Sorry, this PR is massive to review =)

So, as far as I've understood it: 
We rewrite the case of properties and methods. We also do some smart "lower-casing" with works like ARN. 

Is that correct?

src/Core/src/Test/ResultMockFactory.php
```diff
@@ -89,7 +89,27 @@ public static function create(string $class, array $data = [])
...
         foreach ($data as $propertyName => $propertyValue) {
-            $property = $reflectionClass->getProperty($propertyName);
+            if ($reflectionClass->hasProperty($propertyName)) {
+                // backward compatibility with `UpperCamelCase` naming
+                $property = $reflectionClass->getProperty($propertyName);
+            } elseif ($reflectionClass->hasProperty(\lcfirst($propertyName))) {
+                // backward compatibility with `UpperCamelCase` naming
+                $property = $reflectionClass->getProperty(\lcfirst($propertyName));
+            } else {
+                $lowerPropertyName = \strtolower($propertyName);
```
Nyholm, line 99: I just moved down the comment. 

```suggestion
            if ($reflectionClass->hasProperty($propertyName)) {
                $property = $reflectionClass->getProperty($propertyName);
            } elseif ($reflectionClass->hasProperty(\lcfirst($propertyName))) {
                // backward compatibility with `UpperCamelCase` naming (fast)
                $property = $reflectionClass->getProperty(\lcfirst($propertyName));
            } else {
                // backward compatibility with `UpperCamelCase` naming (slow)
                $lowerPropertyName = \strtolower($propertyName);
```

jderusse:
> Sorry, this PR is massive to review =)

Yeah, sorry about that

> So, as far as I've understood it:
> We rewrite the case of properties and methods. We also do some smart "lower-casing" with works like ARN.
> 
> Is that correct?

The goal is to rename properties and methods in order to be consistent in our naming.
I applied camelCase everywhere `Md5`, `Arn`, `Sse`, `Kms`.... Which is, by the way, consistent with our packages Namming (`Sqs`, `Ssm`, ...) 
I also provided a check that throw an exception when a word contains 2 consecutive uppercase letters. To prevent futur addition that won't respect this rule.

I just left 1 exception `GB` (GigaByte)

src/Core/src/Test/ResultMockFactory.php
jderusse: This is not totally accurate. I reword the 3 coments

src/Core/src/Test/ResultMockFactory.php
```diff
@@ -89,7 +89,28 @@ public static function create(string $class, array $data = [])
 
         $initializedProperty->setValue($object, true);
         foreach ($data as $propertyName => $propertyValue) {
-            $property = $reflectionClass->getProperty($propertyName);
+            if ($reflectionClass->hasProperty($propertyName)) {
+                // backward compatibility with `UpperCamelCase` naming
```
Nyholm, line 93: Is this comment correct? Or should it be removed?

Nyholm (approved):
Thank you for the clarification and for the great work

:+1:

src/Core/src/Test/ResultMockFactory.php
jderusse: hmm.. Maybe it can be improved

jderusse: oohhh.... OK, I understand, I was talking about the property's name, while you talked about the user's input...
You're right, your point of view is simpler, applied your first suggestion

Nyholm: Sorry for being unclear.

PR #926, opened 2021-01-21
Bump CS tools

jderusse:
Bump versions of phpstan, php-cs-fixer and psalm


Nyholm (approved):
Nice to finally use psalm 4

PR #929, opened 2021-01-22
Add github-action cache

jderusse:


.github/workflows/checks.yml
```diff
@@ -49,12 +49,26 @@ jobs:
...
 
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+      - name: Cache Composer
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ hashFiles('composer.json') }}
```
bendavies, line 60: why do you alternate between
`composer-${{ hashFiles('composer.json') }}`
and
`composer-${{ hashFiles('**/composer.json') }}`
?
Anyway, I think you should use this to make the cache additive, since you not commit a `composer.lock`
`composer-${{ github.sha }}`

jderusse: I used `**/composer.json` when running `composer update` against each component individualy
And `composer.json` when running `composer update` on the root directory only

> Anyway, I think you should use this to make the cache additive, since you not commit a composer.lock
composer-${{ github.sha }}
:facepalm: of course... the composerjson does not vary, but the reste of the world move and new version could be available..

thanks!

.github/workflows/checks.yml
```diff
@@ -49,12 +49,26 @@ jobs:
...
+
       - name: Checkout code
         uses: actions/checkout@v2
 
       - name: Download
         run: |
+          ls -al ~/.composer/
           composer update --no-interaction --prefer-dist --optimize-autoloader --prefer-stable
+          ls -al ~/.composer/
+          # /home/runner/.composer
```
Nyholm, line 71: Is this just for debugging?

.github/workflows/ci.yml
```diff
@@ -29,6 +29,17 @@ jobs:
...
 
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+      - name: Cache Composer
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ github.sha }}R
```
Nyholm, line 40: Why R?

.github/workflows/ci.yml
```diff
@@ -130,6 +141,17 @@ jobs:
...
```
Nyholm, line 152: Same

Nyholm (commented):
Thank you. 

I like the idea of the .cache directory

.github/workflows/checks.yml
jderusse: good catch

.github/workflows/ci.yml
jderusse: `shit - r` to reload changes in my `tig` (my git editor) ^^ 
The focus were in the wrong window...

PR #930, opened 2021-01-22
Simplify CI file

jderusse:
Use matrix to configure prefer-lowest and minimum-stability

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           coverage: none
-          tools: ${{ matrix.strategy == 'lowest' && 'flex' || '' }}
+          tools: flex
 
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,19 +9,25 @@ on:
...
+        - php: '7.2'
+          strategy: 'lowest'
 
     steps:
       - name: Set up PHP
         uses: shivammathur/setup-php@2.7.0
         with:
           php-version: ${{ matrix.php }}
           coverage: none
+          tools: ${{ matrix.strategy == 'lowest' && 'flex' || '' }}
```
Nyholm, line 30: We can always add flex as a tool.

Nyholm (approved):
Thank you

.github/workflows/ci.yml
jderusse: no strong opinion about this. But I feel it's no more necessary with `composer 2` (unless combined with SYMFONY_REQUIRE).

I changed it to always add flex.
BUT I also change the `SYMFONY_REQUIRE` to use `>=` constraint and avoid that someday we forgot to include latest releases

Nyholm: I wanted to always include flex because the config file would be simpler. I agree, there is no real benefit on the actual execution.

PR #933, opened 2021-01-25
Move logic outside exception

jderusse:
This fixes several design issues:
- Remove logic from Exception's constructor
- Replace static calls to AwsErrorFactory by dependency injection
- Use specialized parser (Xml, JsonRpc, Rest, ...) instead of blindly try json_decode, and presume it's XML

This factory could be reused in #921 

All classes, but RetryStrategy are internal and changing the signature should be BC compatible.

### DIFF
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace AsyncAws\Core;
 
+use AsyncAws\Core\AwsError\AwsErrorFactoryInterface;
+use AsyncAws\Core\AwsError\ChainAwsErrorFactory;
 use AsyncAws\Core\Credentials\CacheProvider;
 use AsyncAws\Core\Credentials\ChainProvider;
 use AsyncAws\Core\Credentials\CredentialProvider;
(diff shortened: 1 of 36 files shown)


src/Core/src/Response.php
```diff
@@ -339,30 +351,47 @@ public function toStream(): ResultStream
...
 
+    /**
+     * In PHP < 7.4, a reference to the arguments is present in the stackTrace of the exception.
+     * This creates a Circular reference: Response -> resolveResult -> Exception -> stackTrace -> Response.
+     * This mean, that calling `unset($response)` does not call the `__destruct` method and does not throw the
+     * remaining exception present in `resolveResult`. The `__destruct` method will be called once the garbage collector
+     * will detect the loop.
+     * That's why this method does not creates exception here, but creates closure instead that will be resolved right
+     * before throwing the exception.
+     */
```
Nyholm, line 362: Thank you <3

src/Core/src/AbstractApi.php
```diff
@@ -169,6 +182,11 @@ protected function getSignerFactories(): array
         ];
     }
 
+    protected function getAwsErrorFactory(): AwsErrorFactoryInterface
+    {
+        return new ChainAwsErrorFactory();
```
Nyholm, line 187: The ChainAwsErrorFactory is only used as a fallback, correct?

Nyholm (approved):
Thank you

src/Core/src/AbstractApi.php
jderusse: yes, this is for backward compatibility

Nyholm: Cool cool 

Well done. Im really impressed by your work in your PRs

jderusse: thanks you :hugs:

PR #934, opened 2021-01-26
Fix cache invalid because of concurrency

jderusse:
It sometime happens when stoping and resuming the generator


jderusse:
note: superseded by #938

PR #938, opened 2021-01-27
Cache generated files

jderusse:
This PR was on my TODO list for ages!
Sorry for its size :(

## Objective
It improve A LOT (x100) generation of classes:
```diff
 $ time ./generate --all                                                                                                                                            

  [OK] Operations generated                                                                                                                                                                                          

-real    0m46.040s
-user    3m9.335s
-sys     0m5.024s
+real    0m0.582s
+user    0m3.052s
+sys     0m0.419s
```

## How does it work?.

96% of the time is spent in php-cs-fixer. The easy fix, is to enable cache for php-cs-fixer.
But that's not that easy, because the generator override classes all the time, and php-cs-fixer's cache is never fresh.

This PR adds a Cache layer on top of FileWriter. It stores the md5 of the file before fixing and after fixing.
When the generator want dumping a new Class, it check, in the cache if it's necessary to dump that file.

To achieve that, I have to refactor several things:
- added a `Cache` class that allows read/write in a multi-concurent process without locking for too long
- removing the FileWriter injected in all generator (I noticed that the same file were dumped several times)
- add a ClassBuiler (that is a merge of nette PhpNamespace + ClassType)
- adding a new ClassRegistry to maintain the list of updated classes


Nyholm (approved):
Awesome! 
This is great. Thank you =)

PR #941, opened 2021-01-28
Add integration test via localstack

jderusse:
This adds localstack and enable integration tests for:
* sts
* cloudformation
* cloudwatch logs
* events bridge
* iam
* sns
* ssm

When possible, I migrate existing services to localstack:
* dynamodb 

Other services can't be migrated because use a specific set of feature that are not provided by localstack
* s3
* sqs
* lambda

### DIFF
--- a/src/Core/composer.json
+++ b/src/Core/composer.json
@@ -18,7 +18,7 @@
         "psr/cache": "^1.0",
         "psr/log": "^1.0",
         "symfony/deprecation-contracts": "^2.1",
-        "symfony/http-client": "^4.4 || ^5.0,!=5.2.0",
+        "symfony/http-client": "^4.4.16 || ^5.1.7,!=5.2.0",
         "symfony/http-client-contracts": "^1.1.8 || ^2.0",
         "symfony/service-contracts": "^1.0 || ^2.0"
     },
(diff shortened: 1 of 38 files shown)


src/Core/composer.json
```diff
@@ -18,7 +18,7 @@
         "psr/cache": "^1.0",
         "psr/log": "^1.0",
         "symfony/deprecation-contracts": "^2.1",
-        "symfony/http-client": "^4.4 || ^5.0,!=5.2.0",
+        "symfony/http-client": "^4.4.16 || ^5.1.7,!=5.2.0",
```
jderusse, line 21: Not really related to this PR, but the PR reveal the bug.
I old versions (prior this PR https://github.com/symfony/symfony/pull/38551) the http-client throws error when the content-type were not occurate.

Which is the case for localstack **AND** for few endpoint in AWS. We once maintained a custom implementation of `toArray`, but when we removed this patch we forgot to bump the composer.json

Nyholm (approved):
Thank you!

Nyholm:
Hm.. Tests seams to be failing: https://github.com/async-aws/core/runs/1791266599?check_suite_focus=true

jderusse:
> Hm.. Tests seams to be failing: https://github.com/async-aws/core/runs/1791266599?check_suite_focus=true

Hmm.. I added a makefile, but forgot to call it in github action

jderusse:
Fixed by #944

PR #943, opened 2021-01-29
[S3] Add methods to handle bucket cors

sunkan:
This is not a 100% complete because what amazon says it returns when getting cors setting and what the generated code says are not the same.

According to [the documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketCors.html#API_GetBucketCors_ResponseSyntax) the response should look like this
```
<CORSConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <CORSRule>
      <AllowedHeader>string</AllowedHeader>
      <AllowedMethod>string</AllowedMethod>
      <AllowedOrigin>string</AllowedOrigin>
      <ExposeHeader>string</ExposeHeader>
      <MaxAgeSeconds>integer</MaxAgeSeconds>
   </CORSRule>
</CORSConfiguration>
```

But according to the generated code it should look like this:
```
<CORSConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <member>
      <AllowedHeader><member>string</member></AllowedHeader>
      <AllowedMethod><member>string</member></AllowedMethod>
      <AllowedOrigin><member>string</member></AllowedOrigin>
      <ExposeHeader><member>string</member></ExposeHeader>
      <MaxAgeSeconds>integer</MaxAgeSeconds>
   </member>
</CORSConfiguration>
```

I don't know how to resolve this. 

If we can't find a simple solution to this, I'm fine with removing the `GetBucketCors` method for now.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -202,6 +202,7 @@
                 "CreateBucket",
                 "CreateMultipartUpload",
                 "DeleteBucket",
+                "DeleteBucketCors",
                 "DeleteObject",
                 "DeleteObjects",
                 "GetObject",
--- a/manifest.json
+++ b/manifest.json
@@ -212,6 +212,7 @@
                 "ListParts",
                 "ObjectExists",
                 "ObjectNotExists",
+                "PutBucketCors",
                 "PutBucketNotificationConfiguration",
                 "PutObject",
                 "PutObjectAcl",
--- a/manifest.json
+++ b/manifest.json
@@ -205,6 +205,7 @@
                 "DeleteBucketCors",
                 "DeleteObject",
                 "DeleteObjects",
+                "GetBucketCors",
                 "GetObject",
                 "GetObjectAcl",
                 "HeadObject",
--- a/src/CodeGenerator/src/Generator/GeneratorHelper.php
+++ b/src/CodeGenerator/src/Generator/GeneratorHelper.php
@@ -24,6 +24,7 @@ class GeneratorHelper
         // Ordered by search length to avoid collision and wrong substitution
         static $replacements = [
             'BOOL' => 'Bool',
+            'CORS' => 'Cors',
             'ETag' => 'Etag',
             'NULL' => 'Null',
             'AWS' => 'Aws',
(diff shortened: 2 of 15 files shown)


Nyholm:
Thank you for this PR. 

If you run the code, is it working?

sunkan:
Yes it dose. I don't know why I couldn't get the tests to work with the other response. 

Have update the test with a working payload

Nyholm:
Oh, you are talking about the generated test code, right?

jderusse:
Looks like the generated code expect a `<CORSRule>` tag (https://github.com/async-aws/aws/pull/943/files#diff-baa90726f9ec827d2b4ed2c3687eb480354577f09df25e3aa5e73ac725b3c239R30)

This is a bug with the TestGenerator which can not be solved: the AWS example schema is not accurate enough to resolve such situation. 

Anyway, 
- :heavy_check_mark: Your test uses `CORSRule` like in the the official documentation
- :heavy_check_mark: The code reads `$xml->CORSRule`
- :heavy_check_mark: You have tested it with the real AWS endpoint

All lights are green for me

jderusse:
Please rebase your PR to take into account last change in code generator

Nyholm (approved):
Yes, I think everything looks good too. 

The generated test code is just to bootstrap the test process. Good thing you fixed it.

sunkan:
Branch rebased

I got this error when trying to re generate the code
```
In GeneratorHelper.php line 54:
                                                             
  No camel case property "CORSRules" is not yet implemented  
                                                             

```

Nyholm:
Good. 

The CodeGenerator does not know how to `strtolower` "CORS". In `GeneratorHelper`, just before row 25 there is a list of exceptions. Could you add `CORS=>Cors` there? Ie `getCors()` is better than `getCORS()`

Nyholm (approved):
Thank yuo

jderusse (approved):

jderusse:
thank you

PR #955, opened 2021-02-17
Make sure we support SF 6

Nyholm:
This will fix the deprecation warnings


jderusse:
I was working on a PR to fix it, but yours is much more better! :heart:

Thank you so much @Nyholm

PR #961, opened 2021-02-27
Don't override PhpCodeGenerator

jderusse:
While trying to fix the PhpStan issue with `Printer::printParameters` method signature, I realized that the original project is moving while we continue using an old variation.

I deleted our custom code to rely on the original implementation, and ran `generate --all` (without cache). And realized that there were no-diff. That's why I think this is not needed anymore.


Nyholm (approved):
Great!

PR #963, opened 2021-03-04
Indicate compatibility with psr/cache 2 and 3

derrabus:
The PSR cache interfaces have been updated with type declarations. Let's welcome the shiny new typed caches by updating our composer.json. 🙂


Nyholm:
Thank you

PR #969, opened 2021-03-18
RequestContext should not be internal

GromNaN:
The `Signer` interface uses the internal class `RequestContext` is its methods signature. Consequently the `SignerV4` cannot be used outside of library whereas it is not explicitly tagged with `@internal` annotation.

My actual use-case is signing an STS request to authenticate against a Vault server.
https://gist.github.com/GromNaN/7477b475a23bd1be682ac98f6767d4bd

### DIFF
--- a/src/Core/src/RequestContext.php
+++ b/src/Core/src/RequestContext.php
@@ -8,8 +8,6 @@ use AsyncAws\Core\Exception\InvalidArgument;
  * Contains contextual information alongside a request.
  *
  * @author Jérémy Derussé <jeremy@derusse.com>
- *
- * @internal
  */
 class RequestContext
 {


Nyholm:
Hm.. As much as I want to keep the `RequestContext` internal, I do agree that we have to make it non-internal since it is an argument to a public interface. 

Do we want to clean up the `RequestContext` before we release this change?

GromNaN:
Only `getCurrentDate` and `getExpirationDate` are actually used by the `SignerV4`. 
I've seen the class `AsyncAws\Core\Request` is internal too.

jderusse:
Hmm.. RequestContext is also internal because it let us free to change signature of methods whenever we want (this class is a mostly a Bag to let Client bring informations to the Signer (ie. S3 need the name of the operation)).

Request, + GetCallerIdentityRequest::request + RequestContext... That's a lot of internal class and method, and I'm not sure we should opened all of theme.

I suggest to either:
- Implement the VaultAuthenticator in AsyncAws and let us maintained it:
- Expose a new public method that return header/body/method (needed by vault if I understand well). Like we did for the `preSignUrl`.

GromNaN:
I need to sign a request. Maybe, a method `StsClient::sign` equivalent to [`presign`](https://github.com/async-aws/aws/blob/master/src/Core/src/AbstractApi.php#L97-L107) could be added.

Otherwise, if the vault supported presigned GET queries, I would simply call `StsClient::presign(GetCallerIdentityRequest::create())`. There is hope https://github.com/hashicorp/vault/pull/10961

GromNaN:
Closing this MR as it does not provide a complete solution.

PR #972, opened 2021-03-22
Fix wrong ContentEncoding for chunked stream

michealmouner:
problem
if I specify `ContentEncoding` for a file bigger than `64kb` .. the `Content-Encoding` option will be ignored
Screenshot 2021-03-22 at 14 02 00

it will appear empty in S3
Screenshot 2021-03-22 at 14 00 34

according to https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html
we can add multiple ContentEncoding so I check the request & add the right `ContentEncoding` if available

Actual Result after this PR
Screenshot 2021-03-22 at 13 54 57

### DIFF
--- a/src/Service/S3/src/Signer/SignerV4ForS3.php
+++ b/src/Service/S3/src/Signer/SignerV4ForS3.php
@@ -118,7 +118,7 @@ class SignerV4ForS3 extends SignerV4
         // Add content-encoding for chunked stream if available
         $customEncoding = $context->getRequest()->getHeader('content-encoding');
         $contentEncoding = 'aws-chunked';
-        if($customEncoding){
+        if ($customEncoding) {
             $contentEncoding .= sprintf(', %s', $customEncoding);
         }
 
--- a/src/Service/S3/src/Signer/SignerV4ForS3.php
+++ b/src/Service/S3/src/Signer/SignerV4ForS3.php
@@ -116,14 +116,10 @@ class SignerV4ForS3 extends SignerV4
         }
 
         // Add content-encoding for chunked stream if available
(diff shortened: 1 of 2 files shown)


Nyholm:
Thank you. 

Could you add a small test to this?

michealmouner:
> Thank you.
> 
> Could you add a small test to this?

I added small tests .. thank you in advance

src/Service/S3/src/Signer/SignerV4ForS3.php
```diff
@@ -115,8 +115,15 @@ protected function convertBodyToStream(SigningContext $context): void
             return;
         }
 
+        // Add content-encoding for chunked stream if available
+        $customEncoding = $context->getRequest()->getHeader('content-encoding');
+        $contentEncoding = 'aws-chunked';
+        if ($customEncoding) {
+            $contentEncoding .= sprintf(', %s', $customEncoding);
+        }
```
jderusse, line 123: `sprintf` looks overkill to me here, it can be replaced by `', '. $customEncoding` or string interpolation?

Or even remove the entire if.
```suggestion
        $contentEncoding = $customEncoding ? "aws-chunked, $customEncoding" : 'aws-chunked';
```

jderusse (approved):
LGTM, Did you tested it with a real call to S3?

I've just a small comment

michealmouner:
> LGTM, Did you tested it with a real call to S3?

yes I tested the fix with a JS library file > 64kb and it works for me (images attached above from real example before and after my PR)

Nyholm:
Awesome. Thank you for your work!

Nyholm (approved):

PR #977, opened 2021-03-29
[Lambda] Add operation to list functions, list function version and to delete functions

Nyholm:
Im creating a small app to delete layer functions not used.

### DIFF
--- a/src/Service/Lambda/src/Enum/Runtime.php
+++ b/src/Service/Lambda/src/Enum/Runtime.php
@@ -3,7 +3,7 @@
 namespace AsyncAws\Lambda\Enum;
 
 /**
- * A runtime identifier. For example, `go1.x`.
+ * The runtime environment for the Lambda function.
  */
 final class Runtime
 {
--- a/src/Service/Lambda/tests/Integration/LambdaClientTest.php
+++ b/src/Service/Lambda/tests/Integration/LambdaClientTest.php
@@ -4,14 +4,17 @@ namespace AsyncAws\Lambda\Tests\Integration;
 
 use AsyncAws\Core\Credentials\NullProvider;
 use AsyncAws\Core\Result;
+use AsyncAws\Core\Test\TestCase;
 use AsyncAws\Lambda\Input\AddLayerVersionPermissionRequest;
+use AsyncAws\Lambda\Input\DeleteFunctionRequest;
 use AsyncAws\Lambda\Input\InvocationRequest;
+use AsyncAws\Lambda\Input\ListFunctionsRequest;
 use AsyncAws\Lambda\Input\ListLayerVersionsRequest;
+use AsyncAws\Lambda\Input\ListVersionsByFunctionRequest;
 use AsyncAws\Lambda\Input\PublishLayerVersionRequest;
 use AsyncAws\Lambda\LambdaClient;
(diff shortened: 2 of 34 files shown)


src/Service/Lambda/src/Enum/Runtime.php
```diff
@@ -3,7 +3,7 @@
 namespace AsyncAws\Lambda\Enum;
 
 /**
- * A runtime identifier. For example, `go1.x`.
+ * The runtime environment for the Lambda function.
```
jderusse, line 6: why isn't it updated by the bot?

Nyholm: Maybe it is. Let's see tomorrow when the bot runs next time.

Nyholm: hm.. No update by the bot..

Strange.

Nyholm:
The PR is updated and ready for review. 

I generated 3 operations and added tests. I also added a changelog as I prepare a release after merging this.

src/Service/Lambda/src/Enum/Runtime.php
jderusse: Ok, I know why.
We don't have documentation for Shapes. We only have documentation when a Shape is referenced by an attribute.

For the Runtime Shape, we have differents documentations:
```
"Runtime": {
      "base": null,
      "refs": {
        "CompatibleRuntimes$member": null,
        "CreateFunctionRequest$Runtime": "<p>The identifier of the function's <a href=\"https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html\">runtime</a>.</p>",
        "FunctionConfiguration$Runtime": "<p>The runtime environment for the Lambda function.</p>",
        "ListLayerVersionsRequest$CompatibleRuntime": "<p>A runtime identifier. For example, <code>go1.x</code>.</p>",
        "ListLayersRequest$CompatibleRuntime": "<p>A runtime identifier. For example, <code>go1.x</code>.</p>",
        "UpdateFunctionConfigurationRequest$Runtime": "<p>The identifier of the function's <a href=\"https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html\">runtime</a>.</p>"
      }
    },
```
That's why: before your PR, the object was only used by `FunctionConfiguration`.
Now it's used by `FunctionConfiguration` and `ListLayerVersionsRequest` and the second definition is used.

src/Service/Lambda/tests/Integration/LambdaClientTest.php
```diff
@@ -38,6 +41,20 @@ public function testAddLayerVersionPermission(): void
         self::assertStringContainsString('change it', $result->getRevisionId());
     }
 
+    public function testDeleteFunction(): void
+    {
+        self::markTestSkipped('The Lambda Docker image does not implement DeleteFunction.');
```
jderusse, line 46: ho. Really? :(

jderusse (approved):

src/Service/Lambda/tests/Integration/LambdaClientTest.php
Nyholm: I got a 405

PR #981, opened 2021-04-01
Fix forming signature with multiple space

vlastv:
Ref https://github.com/aws/aws-sdk-php/blob/master/src/Signature/SignatureV4.php#L301

Problem example:

```php
$s3->copyObject([
  'Bucket' => 'target',
  'Key' => '1.txt',
  'CopySource' => 'source/1  2.txt', // double space
]);
```


jderusse (approved):
Good catch

Nyholm (approved):
Nice work!

PR #985, opened 2021-04-16
Fix className format

### LINKED ISSUE #984: Template issue leads to documentation having the wrong use statement
https://async-aws.com/clients/sqs.html

`use AsyncAws\SQS\SQSClient;`

wants to be

`use AsyncAws\Sqs\SqsClient;`

under "Usage" it is current. I looked into the source and decided my level of understanding is not high enough to contribute a fix:

`<pre><code class="language-php hljs">{{ clientFqcn|printHLJSUseStatement }}`

jderusse:
Fixes #984 

`SQS` => `Sqs`
`S3` => `S3`
`EC2` => `Ec2`
`FooBar` => `FooBar`
`FOOBar` => `FooBar`
`FooBAR` => `FooBar`

### DIFF
--- a/website/Twig/AsyncAwsTwigExtension.php
+++ b/website/Twig/AsyncAwsTwigExtension.php
@@ -38,6 +38,9 @@ class AsyncAwsTwigExtension extends AbstractExtension
         return [
             new TwigFilter('asset', [$this, 'parseAssetUrls']),
             new TwigFilter('lcfirst', function($str) { return lcfirst($str);}),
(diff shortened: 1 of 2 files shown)


Nyholm:
But the `name` is only used in this file right? And on just a few places. 

I just changed the name ¯\_(ツ)_/¯ 

What solution is better?

jderusse:
Closed in favor of #986 which is more consistent in naming

PR #986, opened 2021-04-16
Use better names of the clients

### LINKED ISSUE #984: Template issue leads to documentation having the wrong use statement

Nyholm:
This will fix #984. 

It is an alternative to #985

### DIFF
--- a/docs/clients/ecr.md
+++ b/docs/clients/ecr.md
@@ -1,7 +1,7 @@
 ---
 layout: client
 category: clients
-name: ECR
+name: Ecr
 package: async-aws/ecr
 ---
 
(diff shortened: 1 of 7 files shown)


jderusse (approved):

Nyholm:
Thank you for the review

PR #996, opened 2021-04-27
Enforce JSON validity

### LINKED ISSUE #993: Handle json_encode failures properly
We had recently a problem with large items that should be saved in DynamoDB. The limitation of 400kb for a DynamoDB item forced us to compress the content with gzip. This leads to a problem in `PutItemInput` class (line 250): the class tries to `json_encode` the payload, which does not work with binary content. Because of that the client tries to send `false` instead of valid json. 

It took me some time to figure out what the actual problem was, because there is no client-side validation of the sent content. It would be good to check if the payload is a valid string, instead of sending invalid data to AWS and let them do the validation. As far as I see there several classes where this missing check could create an error.

jderusse:
Fixes #993

When people provide an invalid input, we could (when PHP >= 7.3) easily throw an exception and let people understand their mistake.

I choose to NOT provide backward compatibility to PHP < 7.3 to avoid unnecessary complex code

### DIFF
--- a/src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
+++ b/src/CodeGenerator/src/Generator/RequestSerializer/RestJsonSerializer.php
@@ -53,7 +53,7 @@ class RestJsonSerializer implements Serializer
             ]), false];
         }
 
-        return ['$bodyPayload = $this->requestBody(); $body = empty($bodyPayload) ? "{}" : json_encode($bodyPayload);', true];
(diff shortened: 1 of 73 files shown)


Nyholm (approved):
Thank you

Nyholm:
Should we add a changelog entry for a patch release too?

jderusse:
CHANGELOGs updated

Nyholm (approved):
Thank you

PR #1000, opened 2021-04-30
Remove rawurlencode

Nyholm:
The header will already be "UrlEncoded". 

This was found by @derrabus in the flysystem 2 adapter. This is a fix for flysystem 1.

### DIFF
--- a/src/Integration/Flysystem/S3/tests/Unit/AsyncAwsS3AdapterTest.php
+++ b/src/Integration/Flysystem/S3/tests/Unit/AsyncAwsS3AdapterTest.php
@@ -27,7 +27,7 @@ use PHPUnit\Framework\TestCase;
 
 class AsyncAwsS3AdapterTest extends TestCase
 {
-    private const BUCKCET = 'my_bucket';
+    private const BUCKET = 'my_bucket';
     private const PREFIX = 'all-files';
 
     public static function setUpBeforeClass(): void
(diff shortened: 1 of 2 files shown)


derrabus (approved):

Nyholm:
Hm. So when you copy from AWS, you need to include the bucket in the `CopySource`. For the Flysystem abstraction, we should not care, we treat a bucket as a filesystem. 

This will break existing applications, but it will be consistent with what the interface specifies and how v2 works.

GrahamCampbell:
Looks like the right thing to do to me. Just need to change the tests. :P

Nyholm:
Thank you. The CI is green now.

src/Integration/Flysystem/S3/tests/Unit/AsyncAwsS3AdapterTest.php
```diff
@@ -631,7 +631,7 @@ public function testCopy()
                     return false;
                 }
 
-                if ($input['CopySource'] !== rawurlencode('/' . self::PREFIX . '/' . $path)) {
+                if ($input['CopySource'] !== self::BUCKCET . '/' . self::PREFIX . '/' . $path) {
```
jderusse, line 634: is there a typo in `BUCKCET`?

jderusse (approved):

src/Integration/Flysystem/S3/tests/Unit/AsyncAwsS3AdapterTest.php
Nyholm: Yes it is =)

GrahamCampbell (approved):
:shipit:

jderusse (approved):

Nyholm:
Thank you

PR #1001, opened 2021-05-04
Bump php-cs-fixer

jderusse:
- Bump to php-cs-fixer 3.0
- change `.gitignore` rule for cache files
- apply rule `native_function_invocation` by letting the new default option `strict=true`

more information in https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/UPGRADE-v3.md)


chalasr (approved):

jderusse:
thank you @chalasr for the review

PR #1002, opened 2021-05-12
Fix dynamo laravel cache store conditional check failed checks

GrahamCampbell:


src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
```diff
@@ -233,13 +234,8 @@ public function add($key, $value, $seconds)
...
 
             return true;
-        } catch (HttpException $e) {
-            $type = $e->getAwsType();
-            if (null !== $type && Str::contains($type, 'ConditionalCheckFailedException')) {
-                return false;
-            }
-
-            throw $e;
+        } catch (ConditionalCheckFailedException $e) {
```
jderusse, line 237: You need to bump `"async-aws/dynamo-db": "^1.1"` in composer.json to use this exception
(or let previous code to keep backward compatibility)

jderusse (commented):
Thank you very much for working on it.
Just a small comment about BC

GrahamCampbell:
Did the old code ever work? It isn't working for me with the latest package, hence this PR. :)

src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php
GrahamCampbell: Done.

Nyholm:
Yes, it did work when it was first created. I guess we are missing some tests for this. That is why we accidentally broke it when updating the exceptions.

jderusse (approved):
thank you.

Nyholm (approved):
Thank you

PR #999, opened 2021-04-30
Route53 client

misantron:
#997

PR #1004, opened 2021-05-14
Setup Route53 service

jderusse:
In order to help #999 this PR adds a new Route53 service (with a single operation `ListHostedZones`)

Once merge, It should ease integration of new operations.


Nyholm:
Why did the tests fail on 7.2 only?

jderusse:
> Why did the tests fail on 7.2 only?

Because of `prefer-lowest`

couscous.yml
```diff
@@ -67,6 +67,9 @@ menu:
             rekognition:
                 text: Rekognition
                 url: /clients/rekognition.html
+            route53:
+                text: Lambda
```
misantron, line 71:
```suggestion
                text: Route53
```
typo?

jderusse: Good catch!!!!! thank you

Nyholm (approved):
Thank you

PR #1006, opened 2021-05-15
Fix lowerCase method in pagniator and waiter

jderusse:


jderusse:
This should fix PhpStan and Psalm errors reported in https://github.com/async-aws/aws/pull/999/checks?check_run_id=2590374136

I'm not sure why these tools did not complain earlier...

Nyholm (approved):
Thank you

PR #1010, opened 2021-05-17
Fix invalid operation names

jderusse:
While working on fixing the pagination I discovered that services like `CloudFront` have invalid operation names in the `api2.json` file.


Nyholm (approved):
Thank you

PR #1011, opened 2021-05-17
Fix pagination when path contains `.`

jderusse:
Properties in pagination metdata (nextMarker, resultKeys, ...) are not result's members but Path to the member.
For Kinesis::DescribeStream, we have
```json
"DescribeStream": {
  "more_results": "StreamDescription.HasMoreShards",
  "output_token": "StreamDescription.Shards[-1].ShardId",
  "result_key": "StreamDescription.Shards"
},
```

This PR add support for such case.

BEWARE, paginated result with such nested attributes are not 100% identical:
In all others case, we can iterator over the result itself or over the getter (async-aws handle pagination in both case).
ie
```
foreach($dynamoClient->listTable() as $tableName) {}

// identical to

foreach($dynamoClient->listTable()->getTableNames() as $tableName) {}
```
BUT with nested object handled by the PRm, the pagination would only work when iterating over the result
```
foreach($kinesisClient->describeStream() as $shard) {}
```
if people want to  iterate over the getter they'll only iterate over the first page
```
// FIRST PAGE ONLY!!
foreach($kinesisClient->describeStream()->getDescription()->getShard() as $shard) {}
```

I hesitated to add a fake getter in the result to provide the same behavior
```
// function `getShard` should not existsm but the Paginator added it
foreach($kinesisClient->describeStream()->getShard() as $shard) {}
```
What do you think?

### DIFF
--- a/src/CodeGenerator/src/Generator/PaginationGenerator.php
+++ b/src/CodeGenerator/src/Generator/PaginationGenerator.php
@@ -79,30 +79,47 @@ class PaginationGenerator
         $outputClass = $this->resultGenerator->generate($operation);
 
         $classBuilder = $this->classRegistry->register($outputClass->getFqdn(), true);
-
         $classBuilder->addImplement(\IteratorAggregate::class);
-        $resultKeys = $pagination->getResultkey();
-        if (empty($resultKeys)) {
-            foreach ($operation->getOutput()->getMembers() as $member) {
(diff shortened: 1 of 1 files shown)


Nyholm:
I think it is fine as it is. There is a small difference in plural/singular. And this is the API specification, let's strictly follow it.

Nyholm (approved):

PR #1012, opened 2021-05-17
Remove dupplicated code in paginated response

jderusse:
When the pagination iterator over only one result (in opposite to S3::listObject that iterate over Contents and CommonPrefixes)
We can reduce code by removing the duplicate code in `getIterator` to rely on the getter instead.


Nyholm (approved):
Great idea!

PR #1027, opened 2021-06-05
Fixed lowest build

Nyholm:
This will also fix the wrong changelog from #1026

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
         "symfony/filesystem": "^5.0 || ^6.0",
         "symfony/finder": "^4.4 || ^5.0 || ^6.0",
         "symfony/http-kernel": "^4.4 || ^5.0 || ^6.0",
-        "symfony/phpunit-bridge": "^5.0"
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -40,7 +40,7 @@
         "symfony/filesystem": "^5.0 || ^6.0",
         "symfony/finder": "^4.4 || ^5.0 || ^6.0",
         "symfony/http-kernel": "^4.4 || ^5.0 || ^6.0",
-        "symfony/phpunit-bridge": "^5.0"
+        "symfony/phpunit-bridge": "^5.3 || ^6.0"
```
jderusse, line 43: I don't think we need to support 6.0.

For the same reason we use fixes versions for phpstan and psalm, we should use only one version for phpunit

jderusse (approved):

composer.json
Nyholm: Sure. You are correct

jderusse (approved):

Nyholm:
Thank you for the review

PR #1033, opened 2021-06-25
Fix rewinded s3 signer

### LINKED ISSUE #950: Generator passed to yield from was aborted without proper return and is unable to continue
When waiting on multiple Results over `Result::wait()` we randomly got following error:
```
Fatal error: Uncaught Error: Generator passed to yield from was aborted without proper return and is unable to continue in /var/www/html/vendor/async-aws/core/src/Stream/IterableStream.php:51
Stack trace:
#0 /var/www/html/vendor/symfony/http-client/HttpClientTrait.php(310): AsyncAws\Core\Stream\IterableStream->getIterator()
#1 [internal function]: Symfony\Component\HttpClient\CurlHttpClient::Symfony\Component\HttpClient\{closure}(Object(AsyncAws\Core\Stream\IterableStream))
#2 /var/www/html/vendor/symfony/http-client/HttpClientTrait.php(292): Generator->valid()
#3 /var/www/html/vendor/symfony/http-client/CurlHttpClient.php(462): Symfony\Component\HttpClient\CurlHttpClient::Symfony\Component\HttpClient\{closure}(65536)
#4 /var/www/html/vendor/symfony/http-client/CurlHttpClient.php(247): Symfony\Component\HttpClient\CurlHttpClient::readRequestBody(65536, Object(Closure), '', false)
#5 [internal function]: Symfony\Component\HttpClient\CurlHttpClient::Symfony\Component\HttpClient\{closure}(Resource id #8604, NULL, 65536)
#6 / in /var/www/html/vendor/async-aws/core/src/Stream/IterableStream.php on line 51
```
The results that are being waited on are a list of `PutObjectOutput`.

We aren't able to reproduce the error reliably and don't know what may trigger that error.

jderusse:
fixes #950

The Body sent to the HttpClient MUST be rewindable, because, if the HttpClient decides to retry a request, it needs to rewind the stream.

### DIFF
--- a/src/Service/S3/CHANGELOG.md
+++ b/src/Service/S3/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Fixed
+
+- Fix issue when a request to upload a file is retried
+
 ### Added
 
 - AWS enhancement: Documentation updates for Amazon S3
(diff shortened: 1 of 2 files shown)


Nyholm (approved):

Nyholm:
Thank you

PR #1054, opened 2021-07-22
Make CI run on PHP 8.1

Nyholm:


Nyholm:
CI was not running on 8.1 because we failed to get composer cache dir. It failed because the command output a lot of deprecation warnings too. 

This will not fix the build failure but it will make sure CI is running.

jderusse (approved):

Nyholm:
Thank you

PR #1071, opened 2021-08-21
Added support for idempotencyToken

### LINKED ISSUE #1069: Optional value required when creating secret in SecretsManager
When creating/updating a secret in the secrets manager your required to specify a `ClientRequestToken` but the generated client says that the value is optional. 

And if you don't add it the api will throw `InvalidRequestException "You must provide a ClientRequestToken value. We recommend a UUID-type value."`

So is there any way to modify the specification before we generate the code so that we can either generate the `ClientRequestToken` or mark it as required?

The [aws documentation](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_CreateSecret.html#SecretsManager-CreateSecret-request-ClientRequestToken)  does say that it's optional because they automatically generate the value in there SDK

jderusse:
Fixes #1069

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -22,7 +22,7 @@ return (new PhpCsFixer\Config())
         'magic_method_casing' => true,
         'modernize_types_casting' => true,
         'multiline_comment_opening_closing' => true,
-        'native_constant_invocation' => ['include' => ['ReturnTypeWillChange']],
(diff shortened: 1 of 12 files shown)


sunkan (approved):
Thanks for implementing this.

I've tested this in my code base and it works as expected

Nyholm (approved):
Thank you

PR #1073, opened 2021-08-21
Add requirements in composer.json

jderusse:
Automatically update composer to add requirements

This will be very valuable for https://github.com/async-aws/aws/pull/1071 and avoid forgetting polyfill when generating new operation

### DIFF
--- a/src/Service/Route53/composer.json
+++ b/src/Service/Route53/composer.json
@@ -12,7 +12,9 @@
     "license": "MIT",
     "require": {
         "php": "^7.2.5 || ^8.0",
-        "ext-json": "*",
+        "ext-SimpleXML": "*",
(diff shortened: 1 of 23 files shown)


src/Service/Route53/composer.json
```diff
@@ -12,7 +12,9 @@
     "license": "MIT",
     "require": {
         "php": "^7.2.5 || ^8.0",
-        "ext-json": "*",
+        "ext-SimpleXML": "*",
+        "ext-dom": "*",
+        "ext-filter": "*",
```
jderusse, line 17: This is the kind of issue, this PR will avoid

Nyholm (approved):
This is a great idea. Thank you

PR #1079, opened 2021-08-28
Do not convert null to array

Nyholm:
Stop using `empty()` and `!$foo`. Instead we should check if the input is null or an array and then keep the type.


src/Service/CloudFormation/src/Result/DescribeStacksOutput.php
```diff
@@ -100,7 +100,7 @@ protected function populateResult(Response $response): void
         $data = new \SimpleXMLElement($response->getContent());
         $data = $data->DescribeStacksResult;
 
-        $this->stacks = !$data->Stacks ? [] : $this->populateResultStacks($data->Stacks);
+        $this->stacks = null === $data->Stacks ? null : ([] === $data->Stacks ? [] : $this->populateResultStacks($data->Stacks));
```
jderusse: This change looks wrong.
- stacks is not nullable
- PHPStan is write, we can't use `=== []` with an XML node

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
```diff
@@ -255,7 +255,7 @@ private function parseResponseList(ListShape $shape, string $input, bool $requir
             ]), $shape);
         }
 
-        return strtr($required ? '$this->FUNCTION_NAME(INPUT)' : 'empty(INPUT) ? [] : $this->FUNCTION_NAME(INPUT)', [
+        return strtr($required ? '$this->FUNCTION_NAME(INPUT)' : 'null === INPUT ? null : ([] === INPUT ? [] : $this->FUNCTION_NAME(INPUT))', [
```
jderusse, line 258:
```suggestion
        return strtr($required ? '$this->FUNCTION_NAME(INPUT)' : 'null === INPUT || [] === INPUT ? INPUT : $this->FUNCTION_NAME(INPUT)', [
```

src/Service/EventBridge/src/Result/PutEventsResponse.php
```diff
@@ -41,7 +41,7 @@ protected function populateResult(Response $response): void
         $data = $response->toArray();
 
         $this->failedEntryCount = isset($data['FailedEntryCount']) ? (int) $data['FailedEntryCount'] : null;
-        $this->entries = empty($data['Entries']) ? [] : $this->populateResultPutEventsResultEntryList($data['Entries']);
+        $this->entries = null === $data['Entries'] ? null : ([] === $data['Entries'] ? [] : $this->populateResultPutEventsResultEntryList($data['Entries']));
```
jderusse, line 44: this is not compatible with `getEntries` signature

jderusse (commented):
I'm not sure what problem this PR solves?

arrays are initialized to `[]`, why to we want to switch them to `null`?

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
Nyholm: Thank you!

src/Service/DynamoDb/src/Result/BatchGetItemOutput.php
```diff
@@ -191,7 +191,7 @@ private function populateResultItemList(array $json): array
     {
         $items = [];
         foreach ($json as $item) {
-            $a = empty($item) ? [] : $this->populateResultAttributeMap($item);
+            $a = empty($item) ? null : $this->populateResultAttributeMap($item);
             if (null !== $a) {
                 $items[] = $a;
```
Nyholm, line 196: This if-statement was always true before.

Nyholm:
Current master CI is [red](https://github.com/async-aws/aws/actions/runs/1176195375). 

This is because we compare two objects. One created with a constructor and one created from a response. Since objects created from a response never have a list/map property stored as `null`, the test fails. 

This PR is now ready for a review.

src/Service/CloudFormation/src/Result/DescribeStacksOutput.php
Nyholm: Yepp. You are correct. The PR is updated.

src/Service/CloudFormation/src/Result/DescribeStacksOutput.php
```diff
@@ -200,23 +200,23 @@ private function populateResultStacks(\SimpleXMLElement $xml): array
                 'StackName' => (string) $item->StackName,
                 'ChangeSetId' => ($v = $item->ChangeSetId) ? (string) $v : null,
                 'Description' => ($v = $item->Description) ? (string) $v : null,
-                'Parameters' => !$item->Parameters ? [] : $this->populateResultParameters($item->Parameters),
```
jderusse, line 203: this is incompatible with the `stack::getParameter(): array` signature

src/Service/DynamoDb/src/Result/BatchGetItemOutput.php
jderusse: maybe the fix is to remove this `if`.

All our array are initialized with `[]` on purpose. This allows users to use "array methods" or `foreach ...` even when the value is empty.

jderusse:
> Current master CI is [red](https://github.com/async-aws/aws/actions/runs/1176195375).
> 
> This is because we compare two objects. One created with a constructor and one created from a response. Since objects created from a response never have a list/map property stored as `null`, the test fails.

Hmmm . I believe the right fix is: array property in `ObjectValue` should be initialized with `[]` like all other arrays

src/Service/CloudFormation/src/Result/DescribeStacksOutput.php
Nyholm: But have a look at the Stack constructor and in the Stack::getParameters()

This is safe

phpunit.xml.dist
```diff
@@ -11,7 +11,7 @@
     <ini name="error_reporting" value="-1"/>
     <env name="AWS_SHARED_CREDENTIALS_FILE" value="/dev/null" />
     <env name="AWS_CONFIG_FILE" value="/dev/null" />
-    <server name="SYMFONY_DEPRECATIONS_HELPER" value="max[indirect]=3" />
+    <server name="SYMFONY_DEPRECATIONS_HELPER" value="max[direct]=21&amp;max[indirect]=4" />
```
jderusse, line 14: ? I'm not sure we should allow direct deprecation.

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
```diff
@@ -255,7 +255,8 @@ private function parseResponseList(ListShape $shape, string $input, bool $requir
             ]), $shape);
         }
 
-        return strtr($required ? '$this->FUNCTION_NAME(INPUT)' : 'empty(INPUT) ? [] : $this->FUNCTION_NAME(INPUT)', [
+        return strtr($required ? '$this->FUNCTION_NAME(INPUT)' : 'empty(INPUT) ? EMPTY : $this->FUNCTION_NAME(INPUT)', [
+            'EMPTY' => '$data' === substr($input, 0, 5) ? '[]' : 'null',
```
jderusse, line 259: this test on `$data` sounds really dubious

src/Service/CloudFormation/src/Result/DescribeStacksOutput.php
jderusse: Indeed. LGTM

jderusse: hmm.. after thinking about this.
I truly believe we should fix the constror of the ObjectValue instead. like we do for results:

https://github.com/async-aws/aws/blob/8b698e287967fa35a99246e99e69d45b6a47930f/src/Service/CognitoIdentityProvider/src/Result/AdminGetUserResponse.php#L144

It would have make sens to differentiate `null`/`[]` if the class exposed setters. 
But this is not the cases, and the getter converts null to [] anyway.

Working with array everywhere will be simpler IMHO

phpunit.xml.dist
Nyholm: They are deprecations for flysystem. "these method will be changed in version 2". We will never implement version 2.

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
Nyholm: Indeed.

Nyholm:
I'll try to move deprecations to a new PR. 

I'll also try fix `ObjectValue` instead. https://github.com/async-aws/aws/pull/1079#discussion_r697853859

I'll leave this PR open in the meantime.

phpunit.xml.dist
jderusse: Is there another way to silent deprecation @nicolas-grekas ?

Like an a annotation saying? "This class will never be compatible with next major version, ignore this deprecation"

Nyholm:
We could mark the tests as “group legacy”. But I’m not sure that is better.

jderusse:
Ok... I tried to initialize `ObjectValue` and I understand better the error now and this PR.

When used in `Inputs`, properties can be null: this means => "do not send the property in the Request Payload". 

That's why **We can't replace null by `[]`.**

But when parsing the response, I don't know what is the best implementation. Our test is falling because PhpUnit compares Object, but if we had called the `getter`, the test would have been green, because while deal with null values.

I don't have strong opinion about:
- fixing the test
- this PR that prevent creating `[]` in ObjectValues.

Anyway, I've open #1079 that is similar to your PR but fix the `$data == input`. Feel free to close my PR and backport this change in yours.

PR #1082, opened 2021-08-29
Do not initialize array property

jderusse:
remove redundant property initialization (ie. `private $attributes = [];`) because the property is over-hidden in object populator


Nyholm (approved):
Thank you

PR #1083, opened 2021-08-29
Create List with `[]` instead of `null`

jderusse:
replace `?? null` by `?? []` in constructor and remove `return $this->property ?? []` in getters


jderusse:
ok... I understand better the error...

PR #1084, opened 2021-08-29
Fix type check on list of map

jderusse:
In a list (and map), objects are always added to the list even when they are null.

In the same way, map and list should also be added even when they are empty


Nyholm (approved):
Thank you. I agree with this.

PR #1096, opened 2021-09-10
Fix deprecation and add baseline

jderusse:
This PR fix few deprecations 

And also uses deprecation baseline to track and ignore deprecation we won't fix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,9 +154,12 @@ jobs:
       - name: Run tests
         env:
           SYMFONY_PHPUNIT_VERSION: ${{ matrix.phpunit_version }}
-          SYMFONY_DEPRECATIONS_HELPER: ${{ matrix.strategy == 'lowest' && '9999' || '1' }}
+          OVERRIDE_SYMFONY_DEPRECATIONS_HELPER: ${{ matrix.strategy == 'lowest' && 'max[self]=9999' || '' }}
         run: |
           ok=0
+          if [ -n "$OVERRIDE_SYMFONY_DEPRECATIONS_HELPER" ]; then
+            export SYMFONY_DEPRECATIONS_HELPER=$OVERRIDE_SYMFONY_DEPRECATIONS_HELPER
+          fi
           for COMPONENT in $(find src -maxdepth 4 -type f -name phpunit.xml.dist -printf '%h\n' | sort)
           do
             echo ::group::$COMPONENT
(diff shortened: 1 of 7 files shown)


src/Integration/Laravel/Queue/tests/baseline.json
```diff
@@ -0,0 +1,12 @@
+[
+    {
+        "location": "AsyncAws\\Illuminate\\Queue\\Tests\\Unit\\AsyncAwsSqsQueueTest::testPush",
+        "message": "Ramsey\\Uuid\\Uuid implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)",
```
jderusse, line 4: Hm... I wonder if we should really ignore this deprecation...
This is a true deprecation that should/will be fixed.  But this is an indirect deprecation (triggered by a 3rd party).

If we ignore this deprecation, then we'll better have to ignore all the indirect deprecation and only use `max[self]=0` instead of `max[total]=0`

TL;DR = should we care about indirect deprecations or just ignoring them?

.github/workflows/ci.yml
```diff
@@ -154,14 +154,21 @@ jobs:
       - name: Run tests
         env:
           SYMFONY_PHPUNIT_VERSION: ${{ matrix.phpunit_version }}
-          SYMFONY_DEPRECATIONS_HELPER: ${{ matrix.strategy == 'lowest' && '9999' || '1' }}
+          OVERRIDE_SYMFONY_DEPRECATIONS_HELPER: ${{ matrix.strategy == 'lowest' && 'max[total]=9999' || '' }}
         run: |
           ok=0
+          if [ -n "$OVERRIDE_SYMFONY_DEPRECATIONS_HELPER" ]; then
+            export SYMFONY_DEPRECATIONS_HELPER=$OVERRIDE_SYMFONY_DEPRECATIONS_HELPER
+          fi
```
jderusse, line 162: This avoid defining the `SYMFONY_DEPRECATIONS_HELPER` env variable and let the value defined in `phpunit.xml.dist`

src/Integration/Laravel/Queue/tests/baseline.json
Nyholm: I think we should just ignore all indirect deprecations.

.github/workflows/ci.yml
Nyholm: Oh. This is what I was missing before. Thank you!

Nyholm (approved):

Issue #1101, opened 2021-09-20
Support for AppSync


jordihendrix:
I should be able to get most of this done myself, might need a little bit of instruction on how to implement the tests.

Will add a (Draft/WIP) PR with what I've got so far.

jordihendrix:
Still need to fix one or two things and add some documentation, but I could use a little bit of help on the "Generated output not changed" check. I don't really understand why it's failing as I'm pretty sure I just followed the instructions on the site. Also, if I run the `./generate --all` (as suggested by the check output) I end up with changes in tons of files in every service, and it also adds AppSync operations I don't need. Am I doing something wrong here?

Nyholm:
Are you happy with the [AppSync](https://github.com/async-aws/app-sync) client? Should we tag a release?

PR #1102, opened 2021-09-20
Implement AppSync

jordihendrix:
#1101


jderusse:
If you add new operations in the manifest, you have to run again `./generate --all`. Don't forget to rebase on master and run `composer update` before

The `Root tests (stable) check` tell you that some documentation is missing (https://github.com/async-aws/aws/pull/1102/checks?check_run_id=3768841795) in the `doc/client/index` and in the Readme.

I can help you with finalizing this PR if you add the missing tests for the new operations you added.

jordihendrix:
I added some docs and I think I figured out the generated code thing now, it seems some operations got added to the manifest accidentally. I'm not sure if/how I can fix the Psalm failure, I may be missing something but to me it seems to be failing on something that is generated and also identical to the other services so I'm not sure what to do there. Could you help me out with that? And could you let me know if I'm still missing anything else?

Also, don't think it's a problem here but just FYI: if I run `./generate --all` it still starts changing a lot of stuff (mainly order of types in comments and things like that) in all services, if I just do `./generate <someType>` it seems to be fine though.

jderusse:
Thank you. It looks good to me, I just tried to fix psalm error, but can't push it to your repository? 
Could you please allow the maintainers to edit your PR 

Screenshot from 2021-10-04 19-12-45
?

jordihendrix:
Seems like that feature doesn't work since I created the repo in an organization. I've added you to the repo, hope that works too?

PR #1103, opened 2021-09-21
Allow psr/log v2 & 3

nicolas-grekas:


jderusse:
Do we have tests that cover these versions?

nicolas-grekas:
I'm not sure. But the code only consumes loggers, so this should be fine.

jderusse:
Thank you

PR #1109, opened 2021-10-01
Added support for task status step function actions

GrahamCampbell:


jderusse (approved):

docs/clients/step-functions.md
```diff
@@ -22,3 +22,45 @@ $result = $stepFunctions->startExecution(new StartExecutionInput([
 
 echo $result->getExecutionArn();
 ```
+
+### Send a task heartbeat
```
jderusse, line 26: No need to document all the methods we provide.
We don't pretend to replace the official AWS documentation.

GrahamCampbell: Cool. Maybe just success would have been enough. ;)

PR #1131, opened 2021-11-17
Update static tools

jderusse:


src/Core/tests/Unit/Credentials/DateFromResultTest.php
```diff
@@ -32,23 +32,6 @@ public function testWithValidDate()
...
-    public function testWithNoResponse()
-    {
-        $response = $this->getMockBuilder(Response::class)
-            ->disableOriginalConstructor()
-            ->onlyMethods(['info', 'resolve'])
-            ->getMock();
-
-        $response->expects(self::once())
-            ->method('info')
-            ->willReturn([]);
```
jderusse, line 44: This is not possible:
- current implementation always return an array with a response inside
- the class is internal
- the docbloc says it must return a response

PR #1144, opened 2021-12-16
Fixed Composer Normalize workflow

GrahamCampbell:
Tests are failing, but not (all) for related reasons:

> PHP Fatal error:  Declaration of Nyholm\BundleTest\AppKernel::getProjectDir() must be compatible with Symfony\Component\HttpKernel\Kernel::getProjectDir(): string in /home/runner/work/aws/aws/vendor/nyholm/symfony-bundle-test/src/AppKernel.php on line 94

PR #1145, opened 2021-12-16
Fix broken tests on Symfony 6

GrahamCampbell:
The other failing workflow will be fixed by https://github.com/async-aws/aws/pull/1144.


jderusse:
thank you so much @GrahamCampbell

PR #1149, opened 2021-12-18
Support Laravel 9 on `async-aws/illuminate-cache` and `async-aws/illuminate-queue`

GrahamCampbell:


GrahamCampbell:
No. I wanna have the tests actually run against Laravel 9. I also need to check for changes to interfaces in Laravel 9 that may affect this code.

GrahamCampbell:
Don't wait up for this weekend's releases for this PR. Laravel 9 isn't due to do it's first RC release for another 3 weeks or so.

GrahamCampbell:
OK, so major changes are required to the flysystem and mail packages for them to work on Laravel  9. We'd probably need a new major version and drop earlier Laravel versions. I'm gonna re-scope this to just updating the other 2 packages.

src/Integration/Laravel/Cache/composer.json
```diff
@@ -13,8 +13,8 @@
         "php": "^7.2.5 || ^8.0",
         "ext-json": "*",
         "async-aws/dynamo-db": "^1.1",
-        "illuminate/cache": "^6.18.13 || ^7.10 || ^8.0",
-        "illuminate/support": "^6.18.13 || ^7.10 || ^8.0"
+        "illuminate/cache": "^6.18.13 || ^7.10 || ^8.0 || ^9.0",
```
jderusse, line 16: The CI ran with the version 8 (see https://github.com/async-aws/aws/runs/4914754787?check_suite_focus=true#step:10:172)

Any chance to make the CI testing this new version?

driesvints: @jderusse you'll need the following in your composer.json:

```
    "minimum-stability": "dev",
    "prefer-stable": true
```

jderusse: We have a test with dev flag.

main composer.json file need to be updated too

https://github.com/async-aws/aws/blob/4aafa6bf7f3514c1cff14c94cd3f58cfb43401f3/composer.json#L24

GrahamCampbell:
I'll re-visit this in ~10 days when Laravel 9 is actually released, so we don't need to have non-stable versions.

jderusse:
Close in favor of #1190 and #1189

PR #1140, opened 2021-11-30
Prepare releases

jderusse:
Core => 1.13.0

CloudFormation => 1.2.0
CodeDeploy => 1.3.0
DynamoDb => 1.2.0
kinesis => 1.1.0
rekognition => 0.1.6
Route53 => 0.1.2
s3 => 1.10.0
sns => 1.3.0
Sqs => 1.7

SymfonyBundle => 1.6.0

PR #1150, opened 2021-12-20
[CodeDeploy] Added `getDeployment` method

martinssipenko:
Added `getDeployment()`method to `CodeDeployClient`.

### DIFF
--- a/src/Service/CodeDeploy/tests/Integration/CodeDeployClientTest.php
+++ b/src/Service/CodeDeploy/tests/Integration/CodeDeployClientTest.php
@@ -4,6 +4,7 @@ namespace AsyncAws\CodeDeploy\Tests\Integration;
 
 use AsyncAws\CodeDeploy\CodeDeployClient;
 use AsyncAws\CodeDeploy\Input\CreateDeploymentInput;
+use AsyncAws\CodeDeploy\Input\GetDeploymentInput;
 use AsyncAws\CodeDeploy\Input\PutLifecycleEventHookExecutionStatusInput;
 use AsyncAws\CodeDeploy\ValueObject\AppSpecContent;
 use AsyncAws\CodeDeploy\ValueObject\AutoRollbackConfiguration;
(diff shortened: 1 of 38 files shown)


martinssipenko:
@jderusse `Generated output not changed` fails because the newly added `getDeployment` method reuses the same classes as `createDeployment` method, but the generated output for each method is different, how should I proceed with this?

Also I'm not sure what to do with the issues reported by Psalm.

jderusse:
> @jderusse `Generated output not changed` fails because the newly added `getDeployment` method reuses the same classes as `createDeployment` method, but the generated output for each method is different, how should I proceed with this?
> 
> Also I'm not sure what to do with the issues reported by Psalm.

run `./bin/generate --all` to let the script pick the right version for you

martinssipenko:
@jderusse now only Psalm issue remains, should I add it to baseline?

jderusse:
> @jderusse now only Psalm issue remains, should I add it to baseline?

Yes, it's a false positive.

martinssipenko:
Alright, CI is green now.

src/Service/CodeDeploy/tests/Integration/CodeDeployClientTest.php
```diff
@@ -80,6 +81,20 @@ public function testCreateDeployment(): void
...
+        $client = $this->getClient();
+
+        $input = new GetDeploymentInput([
+            'deploymentId' => 'change me',
+        ]);
+        $result = $client->getDeployment($input);
+
+        $result->resolve();
+
+        // self::assertTODO(expected, $result->getdeploymentInfo());
```
jderusse, line 95: could you add an assertion to this test?

jderusse (approved):
Thank you for this PR. 
A last small comment, and I'll merge the PR

martinssipenko:
@jderusse can we expect a release of this any time soon?

jderusse:
> @jderusse can we expect a release of this any time soon?

I'm currently on it see https://github.com/async-aws/aws/pull/1140

Issue #1152, opened 2021-12-25
[illuminate-mail] merge config from services.ses


ankurk91:
Hi,
Laravel keeps the SES configs in `config/services.php` and the [ServiceProvider](https://github.com/async-aws/aws/blob/master/src/Integration/Laravel/Mail/src/ServiceProvider.php#L25) should merge the configs from this location,

Laravel core is already merging, check

https://github.com/laravel/framework/blob/972538f7071e73dcff429c1cb0ec143d66d82dc1/src/Illuminate/Mail/MailManager.php?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L266-L270

Would you accept a PR for this change?

PR #1153, opened 2021-12-25
illuminate-mail: merge config from services.ses

ankurk91:
Fixes https://github.com/async-aws/aws/issues/1152

Laravel is using MAILER concept since v7.x, merging the config wont break it on v6.x

https://github.com/laravel/framework/blob/972538f7071e73dcff429c1cb0ec143d66d82dc1/src/Illuminate/Mail/MailManager.php?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L266-L270

### DIFF
--- a/src/Integration/Laravel/Mail/src/ServiceProvider.php
+++ b/src/Integration/Laravel/Mail/src/ServiceProvider.php
@@ -22,6 +22,12 @@ class ServiceProvider extends AbstractServiceProvider
     public function createTransport(array $config): AsyncAwsSesTransport
     {
         $clientConfig = [];
+
+        $config = array_merge(
+            $this->app['config']->get('services.ses', []),
+            $config
+        );
+
         if (isset($config['key']) && isset($config['secret'])) {
             $clientConfig['accessKeyId'] = $config['key'] ?? null;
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I'm not sure it's a good idea to conflate this config by default.

ankurk91:
Laravel is merging the config already for ses mailer

jderusse:
The `services.ses` configuration looks dedicated to the official AWS SDK. by merging the 2 configurations, we'll end up with keys like `version: latest` which are not supported by `async-aws`. I'm not sure that is a good idea :/

ankurk91:
The mail service provider is not pulling everything from `services.ses`, it is picking specific keys from array. So there is no need to worry about those extra keys.

ankurk91:
Looks like, I could not impress maintainers with my opinions.
I will send another PR to update the docs, the docs does not say anything about mailer,

https://github.com/async-aws/aws/blob/master/docs/integration/laravel.md#illuminate-mail

PR #1162, opened 2022-01-19
Simplify HttpClient injection

### LINKED ISSUE #1159: Not easy to set timeout on HTTP client
There is no way to set a timeout on the underlying symfony http client. This is technically possible to do, but it involves breaking abstraction and using internal classes. I have to construct a retrying http client in my upstream library or application, and then pass it through into the constructor, but the upstream code has to know the error factory used by the client, somehow.

```php
final class Example
{
    public static function buildSqsClient(array $config, LoggerInterface $logger): SqsClient
    {
        /** @psalm-suppress InternalClass */
        return new SqsClient(
            $config,
            null,
            self::getHttpClient(new XmlAwsErrorFactory(), $logger), // `new XmlAwsErrorFactory()` is the really bad code
            $logger,
        );
    }

    private static function getHttpClient(
        AwsErrorFactoryInterface $errorFactory,
        LoggerInterface $logger
    ): HttpClientInterface {
        return new RetryableHttpClient(
            HttpClient::create(['timeout' => 10.0]),
            // also unfortunate to have to duplicate this logic
            new AwsRetryStrategy(AwsRetryStrategy::DEFAULT_RETRY_STATUS_CODES, 1000, 2.0, 0, 0.1, $errorFactory),
            3,
            $logger,
        );
    }
}
```

jderusse:
This PR fix #1159 by making the API to always decorate the HttpClient with the RetryableHttpClient, even when it has been injected by the user.

Given it does not change the behavior of the component when no error occurred, I think it does no harm.

IHMO the worst cast is people injecting an homemade RetryableHttpClient. In this case, the client will be decorated twice, making the request retried `N*M`. But I think this case is rare enough to make it acceptable.

It also allows people to easily inject a custom HttpClient (let's say TracableHttpClient in dev mode) without handling all the stuff about Retry logic + awsErrorFactory

### DIFF
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Make the injected HttpClient decorated by our  `RetryableHttpClient`
 - Support for KMS
 
 ## 1.13.0
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -74,15 +74,15 @@ abstract class AbstractApi
         $this->awsErrorFactory = $this->getAwsErrorFactory();
         if (!isset($httpClient)) {
             $httpClient = HttpClient::create();
-            if (class_exists(RetryableHttpClient::class)) {
-                /** @psalm-suppress MissingDependency */
-                $httpClient = new RetryableHttpClient(
-                    $httpClient,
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
This will break my code, double wrapping and causing lambda timeouts.

GrahamCampbell:
:shipit:

GrahamCampbell (approved):

GrahamCampbell:
Any chance of a new kms and core release, please?

jderusse:
Will be happy to release this PR as well https://github.com/async-aws/aws/pull/1149

PR #1167, opened 2022-01-25
Add new operation Ssm::deleteParameters

hmazter:


jderusse (commented):
Thank you for the addition.

Could you please add an entry in the SSM's CHANGELOG?

hmazter:
> Could you please add an entry in the SSM's CHANGELOG?

Added that now

jderusse:
thank you

PR #1168, opened 2022-01-25
Add new operation CloudWatchLogs::filterLogEvents

hmazter:
Generated a new operation `FilterLogEvents` for the `CloudWatchLogs` Client.

See https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_FilterLogEvents.html

`ext-filter` was added to `composer.json` by the generate command, don't know why/if that is needed.

Saw also that the [interleaved property is deprecated](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_FilterLogEvents.html#API_FilterLogEvents_RequestSyntax) any special way that should be handled? Its always handled as `true` by the API

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -69,6 +69,7 @@
             "methods": [
                 "CreateLogGroup",
                 "DescribeLogStreams",
+                "FilterLogEvents",
                 "PutLogEvents"
             ]
         },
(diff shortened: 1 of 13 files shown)


jderusse (approved):

jderusse:
thank you

PR #1171, opened 2022-01-27
Fix error when config keys arent included

Jefemy:
Code would previously error due to array keys not existing even though it should go ahead and attempt to create the client and search for the AWS keys in the environment variables/other AWS credential stores

### DIFF
--- a/src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemManager.php
+++ b/src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemManager.php
@@ -17,7 +17,7 @@ class AsyncAwsFilesystemManager extends FilesystemManager
     public function createAsyncAwsS3Driver(array $config)
     {
         $s3Config = [];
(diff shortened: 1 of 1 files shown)


src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemManager.php
```diff
@@ -17,7 +17,7 @@ class AsyncAwsFilesystemManager extends FilesystemManager
     public function createAsyncAwsS3Driver(array $config)
     {
         $s3Config = [];
-        if ($config['key'] && $config['secret']) {
+        if (!empty($config['key']) && !empty($config['secret'])) {
             $s3Config['accessKeyId'] = $config['key'] ?? null;
             $s3Config['accessKeySecret'] = $config['secret'] ?? null;
```
jderusse, line 22: I think you can safetly remove the ` ?? null` part

jderusse (approved):

jderusse:
thank you

PR #1172, opened 2022-01-28
Fix documentation

matthieuwerner:
Add missing assignation.


jderusse (approved):

jderusse:
good catch. Thank you

PR #1204, opened 2022-03-16
Do not enable chunked-body by default

jderusse:
Because AWS has a bug and lets an empty `ContentEncoding` header (which break browser that serve such file).
see https://github.com/aws-amplify/aws-sdk-ios/issues/443, https://github.com/aws/aws-sdk-net/issues/559 

I think we should not enable this behavior by default (like what other SDKs do)


martinssipenko:
Just experience this issue the hard way :) After uploading objects to S3 the CloudFront started to fail because of missing value in content-encoding.

PR #1211, opened 2022-04-07
Update Cloudfront example

j0k3r:
`CallerReference` is defined as a string in `AsyncAws\CloudFront\ValueObject\InvalidationBatch` but `time()` return an integer.


jderusse:
Thank you

PR #1216, opened 2022-04-14
Fix ISO822 dates

### LINKED ISSUE #1213: IfModifiedSince not honored
I'm using this package to perform calls to MinIO in dev, and to a regular AWS otherwise. Both seem to completely ignore `IfModifiedSince` headers I set when I attempt to download a file. I'm migrating from the official SDK, which has a working implementation of this feature.

Digging through the code, I see that they hardcode the timezone to `GMT` for some reason: https://github.com/aws/aws-sdk-php/blob/15f6b434b92826a321191bf5b72820934d970580/src/Api/TimestampShape.php#L40

And sure enough, when I use the following code at this line: https://github.com/async-aws/aws/blob/2981f162628b3bd07325f7df36e3918d154db0b8/src/Service/S3/src/Input/GetObjectRequest.php#L353-L355

When I use `gmdate('D, d M Y H:i:s \G\M\T', $value);`

… it works! I'm not sure why

# Code to reproduce the issue

First, create `some_file` inside `my_bucket`

```php
        $this->s3Client->getObject([
            'Bucket' => 'my_bucket',
            'Key' => 'some_file',
            'IfModifiedSince' => new \DateTimeImmutable('now'),
        ] );
```

## Expected result

`[AsyncAws\Core\Exception\Http\RedirectionException (304)]`

jderusse:
fixes #1213

@greg0ire could you give a try?

### DIFF
--- a/src/CodeGenerator/src/Generator/InputGenerator.php
+++ b/src/CodeGenerator/src/Generator/InputGenerator.php
@@ -474,6 +474,11 @@ PHP
                     throw new \InvalidArgumentException('Constant "\DateTimeInterface::' . $format . '" does not exists.');
                 }
 
+                if ('RFC822' === $format) {
+                    $variable .= '->setTimezone(new \\DateTimeZone("GMT"))';
+                    $format = 'RFC7231';
+                }
+
                 return $variable . '->format(\DateTimeInterface::' . $format . ')';
             case 'boolean':
                 return $variable . ' ? "true" : "false"';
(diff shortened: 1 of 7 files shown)


greg0ire:
Awesome! I should be able to try that today 👍

greg0ire:
@jderusse actually I wasn't sure how to proceed since it a monorepo so I applied your patch to `GetObjectRequest`, and:     

```
[AsyncAws\Core\Exception\Http\RedirectionException (304)]
  HTTP 304 returned for "http://redacted".
```

So it seems to work great, thanks 👍

PR #1229, opened 2022-04-29
Fix #1226 listObjectsV2 breaks bucket path to host

### LINKED ISSUE #1226: S3 client inconsistent behavior
Hi,

I have found a very strange behavior in the S3 client.
getObject adds the bucket as the url host. Meanwhile ListObjectsV2 adds it to the path. 
My configuration for pathStyleEndpoint is false.

Redominus:
S3 listObjectsV2 breaks bucket path to host

### DIFF
--- a/src/Service/S3/src/S3Client.php
+++ b/src/Service/S3/src/S3Client.php
@@ -973,8 +973,8 @@ class S3Client extends AbstractApi
     protected function getEndpoint(string $uri, array $query, ?string $region): string
     {
         $uriParts = explode('/', $uri, 3);
-        $bucket = preg_replace('~\?.*~', '', $uriParts[1] ?? '');
-        $uriWithOutBucket = preg_replace('~^[^?]+~', '', $uriParts[1] ?? '') . ($uriParts[2] ?? '');
+        $bucket = explode('?', $uriParts[1] ?? '', 2)[0];
+        $uriWithOutBucket = substr($uriParts[1] ?? '', strlen($bucket)) . ($uriParts[2] ?? '');
         $bucketLen = \strlen($bucket);
         $configuration = $this->getConfiguration();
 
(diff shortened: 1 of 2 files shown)


Redominus:
I would really appreciate some help with PHPStan analysis as I did not change any method signature

jderusse (changes requested):
You can not change this file.  This file is automatically generated and the URL is part of the AWS spec. 

You have 2 options:
- change the generator (that dump the file) to detect URI containing query string and generating the patch you submitted.
- change the S3Client getEndpoint method to ignore the querystring from the URI when guessing the bucket.

The first option is closer to your patch, but I have some concern:
- the $query property is used in many places, including generating signatures. I fear that changing this behavior might generate authentication issue
- Some S3 operations use queryString that looks like `"/{Bucket}?tagging"`, but the AsyncAws core component expects a key/value to generate URL containing `key=value`. If you change the Generator to move the querystring part to the `$query` parameter, you first need to make the `sync/core` component compatible with options without value `$query['tagging'] = null` must generate `?tagging` instead of `?tagging=`

That's why I believe the solution 2 would be better.

Tell me if you need help

Redominus:
Hope this is fine. It does not changes the query.

src/Service/S3/src/S3Client.php
```diff
@@ -973,7 +973,8 @@ protected function getAwsErrorFactory(): AwsErrorFactoryInterface
     protected function getEndpoint(string $uri, array $query, ?string $region): string
     {
         $uriParts = explode('/', $uri, 3);
-        $bucket = $uriParts[1] ?? '';
+        $bucket = preg_replace('~\?.*~', '', $uriParts[1] ?? '');
+        $uriWithOutBucket = preg_replace('~^[^?]+~', '', $uriParts[1] ?? '') . ($uriParts[2] ?? '');
```
jderusse, line 977: no need to use regexp
```suggestion
        $uriWithOutBucket = substr($uriParts[1] ?? '', strlen($bucket)) . ($uriParts[2] ?? '');
```

Redominus:
It had a bit of influence from being a regex golfer :rofl:

jderusse (approved):

PR #1235, opened 2022-05-09
Added endpoint address support for timestream

### LINKED ISSUE #1232: TimestreamWrite `writeRecords` not working
Putting this here so I can remember to fix it.

```
AsyncAws\Core\Exception\Http\ClientException with message 'HTTP 404 returned for "https://ingest.timestream.us-east-1.amazonaws.com/".

Code:    UnknownOperationException
Message: 
Type:    
Detail:  
'
```

GrahamCampbell:
Closes #1232. Depends on #1236.

### DIFF
--- a/docs/clients/timestream-query.md
+++ b/docs/clients/timestream-query.md
@@ -10,12 +10,16 @@ package: async-aws/timestream-query
 ### Execute a query
 
 ```php
+use AsyncAws\TimestreamQuery\Input\DescribeEndpointsInput;
 use AsyncAws\TimestreamQuery\Input\QueryInput;
 use AsyncAws\TimestreamQuery\TimestreamQueryClient;
 
 $timestreamQuery = new TimestreamQueryClient();
 
+$endpoints = $timestreamQuery->describeEndpoints(new DescribeEndpointsInput())->getEndpoints();
+
 $result = $timestreamQuery->query(new QueryInput([
+    'EndpointAddress' => $endpoints[0]->getAddress(),
     'ClientToken' => 'qwertyuiop',
     'QueryString' => 'SELECT * FROM db.tbl ORDER BY time DESC LIMIT 10',
 ]));
--- a/docs/clients/timestream-query.md
+++ b/docs/clients/timestream-query.md
@@ -19,7 +19,7 @@ $timestreamQuery = new TimestreamQueryClient();
 $endpoints = $timestreamQuery->describeEndpoints(new DescribeEndpointsInput())->getEndpoints();
 
 $result = $timestreamQuery->query(new QueryInput([
-    'EndpointAddress' => $endpoints[0]->getAddress(),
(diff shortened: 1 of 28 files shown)


jderusse:
I would like avoiding as much as possible adding new Magic parameters in the methods (We did it with `@region` to mimic the the SDK)

Looking at the official AWS SDK, (class `EndpointDiscoveryMiddleware`), it looks like the workflow is:
- if the operation contains a `endpointdiscovery.required` 
- check if "active" endpoint is present in cache
- if not, call the `endpointDiscoveryOperation` (the one with `endpointoperation: true` flag) to get the endpoint.
- + fresh/stale cache mecahnism

I think, this can be achieved by something like
- in code generator, when `endpointdiscovery.required` is true
- loops on declared operations, to find the operation with `endpointoperation=true`
- injecting a `endpointOperation: foobar` in the request Context
- in AbstractApi, replace, `$request->setEndpoint($this->getEndpoint($request->getUri(), $request->getQuery(), $input->getRegion()));` by logic to fetch endpoint from cache, and fallback to `$this->${$context['endpointOperation']}->getEndpoint()`

- to prevent unsupported edge case, we should not forget to throw some NotImplementeException:
- when endpointOperation is not generated
- when endpointOperation's input has required parameters (shape has member)
- when endpointOperation's don't have a `getEndpoint()` or `cachePeriodInMinutes()`  methods

I don't have much time this week to work on it, but I could have a looks next week if you want.

GrahamCampbell:
I did consider that approach, but it has a major downside: the first time someone makes an HTTP request to do 100 parallel batchWrite requests, we'd then we end up doing them synchronously. It is important for my use case that I can control that there is only one request made to the DescribeEndpoints action, followed by 100 parallel requests.

jderusse:
I think that could be handled.
Thank to the cache layer, users could warmup the cache by calling the endpointOperation themself.
Or by implementing a smarter logic that takes care of deduplicating multiple identical requests.

cesarkohl-voxie:
Interesting.

GrahamCampbell:
As long as the following code will only result in 1 request followed by 100 in parallel, and we don't have to wait for each of the 2nd requests to complete in sequence, I don't mind.

```php
$results = []
$results[] = $client->writeItems(...);
foreach ($results as $result) { $result->resolve(); }
```

GrahamCampbell:
I guess what you saying is a user would have to do one extra line, to enable that:

```
$results = []
$client->describeEndpoints(); 
$results[] = $client->writeItems(...);
foreach ($results as $result) { $result->resolve(); }
```

GrahamCampbell:
I have to say, I prefer the explicit interface of providing the endpoint, but this would also do what I needed.

GrahamCampbell:
I guess to get prefer-lowest to pass, I need to bump the minimum core version, which would require a new release of core, first.x

jderusse:
closed in favor of #1239

Thank you for investigating into this issue and worked on this patch @GrahamCampbell

PR #1236, opened 2022-05-12
Pin to localstack 0.14.2

GrahamCampbell:
EventBridge and SNS are broken on `latest`.


jderusse (approved):
oh :heart: thank you for taking care of this.

PR #1239, opened 2022-05-14
Add support for endpoint discovery

jderusse:
Alternative to https://github.com/async-aws/aws/pull/1235 to automatically discover endpoints.

AWS data.json files provides required information to automatically fetch endpoint.

This PR:
 - adds a new `endpointCache` that take care of caching and reusing previously fetched endpoints with TTL
 - adds a new `client::discoverEndpoints` protected method to fetch the list of endpoints from the client
 - add glue in Request's Context to ask the AbstractApi to fetch endpoints using the AWS API

this PR is in draft and requires to be split inu multiple PR to fix version constraints
- [ ] the `Core` that contain new Context properties
- [ ] The generator + services

### DIFF
--- a/src/Service/DynamoDb/src/DynamoDbClient.php
+++ b/src/Service/DynamoDb/src/DynamoDbClient.php
@@ -32,6 +32,7 @@ use AsyncAws\DynamoDb\Input\BatchWriteItemInput;
 use AsyncAws\DynamoDb\Input\CreateTableInput;
 use AsyncAws\DynamoDb\Input\DeleteItemInput;
 use AsyncAws\DynamoDb\Input\DeleteTableInput;
+use AsyncAws\DynamoDb\Input\DescribeEndpointsRequest;
 use AsyncAws\DynamoDb\Input\DescribeTableInput;
 use AsyncAws\DynamoDb\Input\ExecuteStatementInput;
 use AsyncAws\DynamoDb\Input\GetItemInput;
@@ -48,6 +49,7 @@ use AsyncAws\DynamoDb\Result\BatchWriteItemOutput;
 use AsyncAws\DynamoDb\Result\CreateTableOutput;
 use AsyncAws\DynamoDb\Result\DeleteItemOutput;
 use AsyncAws\DynamoDb\Result\DeleteTableOutput;
+use AsyncAws\DynamoDb\Result\DescribeEndpointsResponse;
 use AsyncAws\DynamoDb\Result\DescribeTableOutput;
 use AsyncAws\DynamoDb\Result\ExecuteStatementOutput;
 use AsyncAws\DynamoDb\Result\GetItemOutput;
@@ -107,7 +109,7 @@ class DynamoDbClient extends AbstractApi
             'ResourceNotFoundException' => ResourceNotFoundException::class,
             'RequestLimitExceeded' => RequestLimitExceededException::class,
             'InternalServerError' => InternalServerErrorException::class,
-        ]]));
+        ], 'usesEndpointDiscovery' => true]));
 
         return new BatchGetItemOutput($response, $this, $input);
     }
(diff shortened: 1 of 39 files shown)


src/Service/Kinesis/phpunit.stdout
```diff
@@ -0,0 +1,4 @@
+PHPUnit 9.5.10 by Sebastian Bergmann and contributors.
```
GrahamCampbell, line 1: committed in error

src/Service/DynamoDb/src/DynamoDbClient.php
```diff
@@ -98,7 +100,7 @@ public function batchGetItem($input): BatchGetItemOutput
             'ResourceNotFoundException' => ResourceNotFoundException::class,
             'RequestLimitExceeded' => RequestLimitExceededException::class,
             'InternalServerError' => InternalServerErrorException::class,
-        ]]));
+        ], 'usesEndpointDiscovery' => true]));
```
GrahamCampbell, line 103: does the aws cli tool actually use the discovered endpoints for dynamo api calls?

GrahamCampbell (changes requested):

src/Service/DynamoDb/src/DynamoDbClient.php
jderusse: yes, this is an opt-in option. (same for my PR)

the behavior is active when:
`data['endpointdiscovery']['required']` is true

or `data['endpointdiscovery']` exists and client is constructed with `endpointDiscoveryEnabled=true` (or associated env variable)

GrahamCampbell:
This looks like a reasonable approach. Let me know if you need anything from me to get this over the line. I am currently using a private fork with my other implementation, but happy to cut over to this, once it lands.

GrahamCampbell (approved):

GrahamCampbell (changes requested):
This doesn't have a slick way to force priming the cache. Anyone who needs reliable performance when kicking off thousands of requests would be unable to use the cache, and would have to first make a client, manually discover the endpoint, then make a new client with the endpoint address.

GrahamCampbell:
> This doesn't have a slick way to force priming the cache.

I suppose what happens here right now is the first call to an endpoint that needs discovered synchronously discovers, so it's not the end of the world.

GrahamCampbell (approved):

jderusse:
> This doesn't have a slick way to force priming the cache

Yes you're right... endpoints are discovered lazily at the first call. I wonder if calls to `discoverEndpoints` shouldn't warm the cache... :thinking: 



Or, if needed, one could add a `warmEndpoint` method, that looks like

```
public function warmEndpoint(string|null $regiuon): void
{
  $endpoints = $this->discoverEndpoints($region);
  $this->endpointCache->addEndpoints($region, $endpoints);
}
```

GrahamCampbell:
It would be cool if we could have a strict mode, where it would crash if there's no warm endpoint (either because it had expired or was never warmed). That would then bring us to 100% feature parity with my other PR. :)

GrahamCampbell:
Or possibly the ability to ignore the TTL on the endpoint cache. ;)

GrahamCampbell:
> Yes you're right... endpoints are discovered lazily at the first call.

Can we discover, non-lazily?

jderusse:
> > Yes you're right... endpoints are discovered lazily at the first call.
> 
> They are discovered eagerly, right? Only once we know the endpoint, do we return?

That would be the point of providing a `warmEndpoint` method. 

I've to take time to think about this, ideally, we should provide:
- user can warmup the endpoint
- user can refresh the endpoint at any time in a non-blocking mode
- AbstractClient can refresh the endpoint by itself before TTL expires in a non-blocking mode

> It would be cool if we could have a strict mode, where it would crash if there's no warm endpoint

Why would you do that?

GrahamCampbell (approved):

PR #1240, opened 2022-05-17
Do not override HttpClient DI

### LINKED ISSUE #1234: No mechanism to disable HTTP retries
There's no mechanism to instantiate a client without retries. The bug with the symfony retry client is really (https://github.com/async-aws/aws/issues/1128) kicking me hard, with 10s of thousands of failures every day. Moreover, when I'm performing lots of requests in parallel, it is not efficient when there is a failure, for each to synchronously re-try. For that reason, I feel the best approach for my is to turn off retries, and manage that higher up the stack, so I can retry large batches of requests at once.

jderusse:
fix #1234 by reverting #1162

The current implementation decorates the injected HttpClient with the RetriableClient.
I think that's a mistake (the clients should not change the injected dependencies).

This PR reverts that behavior, and lets people inject a non-retryable client if they want.

I'm not sure if this is a BC-break or not :thinking: Code wont break sure... but a behavior (auto-setup retry on top of the injected client) could be disabled for some people.

### DIFF
--- a/src/Core/src/HttpClient/AwsHttpClientFactory.php
+++ b/src/Core/src/HttpClient/AwsHttpClientFactory.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace AsyncAws\Core\HttpClient;
+
+use Psr\Log\LoggerInterface;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Component\HttpClient\RetryableHttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Jérémy Derussé <jeremy@derusse.com>
+ */
+class AwsHttpClientFactory
+{
+    public static function createRetryableClient(HttpClientInterface $httpClient = null, LoggerInterface $logger = null): HttpClientInterface
+    {
+        if (null === $httpClient) {
+            $httpClient = HttpClient::create();
+        }
+        if (class_exists(RetryableHttpClient::class)) {
+            /** @psalm-suppress MissingDependency */
+            $httpClient = new RetryableHttpClient(
+                $httpClient,
+                new AwsRetryStrategy(),
+                3,
+                $logger
+            );
+        }
+
+        return $httpClient;
+    }
+}
--- a/docs/clients/index.md
+++ b/docs/clients/index.md
@@ -21,13 +21,16 @@ All arguments are optional and sensible defaults are used.
 ```php
 use AsyncAws\Core\Configuration;
 use AsyncAws\Core\Credentials\ConfigurationProvider;
+use AsyncAws\Core\HttpClient\AwsHttpClientFactory;
 use AsyncAws\DynamoDb\DynamoDbClient;
 
 $config = Configuration::create([
     'region' => 'eu-central-1',
 ]);
 $credentialProvider = new ConfigurationProvider();
-$httpClient = // ... Instance of Symfony's HttpClientInterface
+$httpClient = HttpClient::create(); // ... Instance of Symfony's HttpClientInterface
+// Or, to enable automatic retries on top of your own HttpClient
+// $httpClient = AwsHttpClientFactory::createRetryableClient($httpClient); T
 $logger = // ... A PSR-3 logger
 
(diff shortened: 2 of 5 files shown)


GrahamCampbell:
I think we originally changed that at my request, and I've still not gotten around to relying on the changes, so this is not breaking for me. If I am the only stakeholder, then we're good to merge this (after appeasing the PHP-CS-Fixer gods). Not gonna cause me any issues.

GrahamCampbell (approved):

jderusse:
Indeed, this has been introduced here #1162 to fix #1159.

I changed this PR to also provide a mechanism to create clients and fix #1159

src/Core/src/HttpClient/AwsHttpClientFactory.php
```diff
@@ -0,0 +1,32 @@
...
+    public static function createRetryableClient(HttpClientInterface $httpClient = null, LoggerInterface $logger = null): HttpClientInterface
+    {
+        if (null === $httpClient) {
+            $httpClient = HttpClient::create();
+        }
+        if (class_exists(RetryableHttpClient::class)) {
+            /** @psalm-suppress MissingDependency */
+            $httpClient = new RetryableHttpClient(
+                $httpClient,
+                new AwsRetryStrategy(),
```
GrahamCampbell, line 24: But how do we know what error factory to use?

GrahamCampbell (changes requested):

src/Core/src/HttpClient/AwsHttpClientFactory.php
jderusse: The default factory uses the chainFactory that iterate over known factories.

It's not as performant as the strategy generated by the client. But given it's only used in case of failure, I think it's enough.

jderusse: Maybe, we could move this HttpClie tFactory in the client... 🤔

GrahamCampbell: Yeh, that would work.

jderusse: I was thinking about this, but I feel uncomfortable with adding a `public static function createHttpClient` method in the abstractApi. It feels very hacky and breaks the separation of concerns...

PLUS, building specialized HttpClient could be an issue for people creating the HttpClient once and injecting it in multiple Services (ie S3Client and DynamoDb does not expect the same AWSErrorFactory)

Finally, I think it's better to
- do not inject anything and let the AbstractApi building it's own client.
- if people want to customize the HttpClient AND want the Retry thing expose a factory in a dedicated service that return a generic RetryStrategy that work with all the services.

GrahamCampbell: I would be fine with just the original PR that was accidentality merged yesterday. I already have the code to create the strategy anyway, for each of the services I need.

GrahamCampbell: > do not inject anything and let the AbstractApi building it's own client.

I definitely do not want that. I need to share the client to avoid blowing through my open file limit of 1000 on aws lambda. curl doesn't reuse - i found that out the hard way...

jderusse: I think we agreed. (I also want letting people optionally injecting their own HttpClient)

See the current diff of this PR (which is closed to the first version of this PR + exposes an optional `AwsHttpClientFactory` if people need help to create a Client.

The AbstractApi takes an optional HttpClientInterface
- People that want the default behavior just use `new S3Client()`, and the AbstractApi creates a `RetryableHttpClient`
- People that want full control can inject their own HttpClient `new S3Clien(.., $myClient)`, and the AbstractApi does nothing
- People that want Retryable thing with their own httpclient can use the new Factory `new S3Clien(.., AwsHttpClientFactory::createRetryableClient($myClient))`

GrahamCampbell: yeh

GrahamCampbell (approved):

PR #1247, opened 2022-05-27
Add support for endpoint discovery in core

jderusse:
Followup #1239

This PR add support of endpoint discovery in core.
It could be leverage by client, thank to new parameter in RequestContext and implementing the underlying protected method

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -102,6 +102,13 @@ An identifier for the assumed role session
 The relative path that is used to fetch credentials inside and ECS instance.
 See [IAM Roles for Tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) for more information.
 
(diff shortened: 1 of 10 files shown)


stof:
@jderusse the timestream Endpoint objects don't actually implement the EndpointInterface of the core. So this interface looks useless.

stof:
And it looks like `getDiscoveredEndpoint` ignores the method arguments when finding an endpoint through discovery. Is it expected ?

jderusse:
> it looks like getDiscoveredEndpoint ignores the method arguments

Which argument?

jderusse:
> `$uri` and `$query`

You're right. 
Hopefully, it did not have any impact on our single client using discovered endpoint (DynamoDb) because all implemented endpoints used `uri = /` uri and `query = []`.

Fixed in https://github.com/async-aws/aws/pull/1415

PR #1253, opened 2022-05-28
[CSS] Increase max height on website sidebar

Nyholm:
We need a max height for the transition. But it needs to be slightly larger than 99rem. Currently we dont see Translate or Xray
Screenshot 2022-05-28 at 15 33 28


jderusse (approved):
I wonder if we should remove this animation :thinking:

PR #1267, opened 2022-06-29
feat: Add Iot client

tyx:
Hello !

Here is a new client to allow us to do some Iot stuff.

To start we mainly need to create and list things.

Hope nothing is missing.


tyx:
I just realized that iot is only supported in pro version for localstack. How am I supposed to do integration test ?

jderusse:
> I just realized that iot is only supported in pro version for localstack. How am I supposed to do integration test ?

In that situation, we don't provide any integration test.

tyx:
So I just need to remove the integration folder ?

tyx:
I also had trouble with psalm but with another Client (Cognito) so unrelated to this PR I guess

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
```diff
@@ -111,7 +111,7 @@ public function generateForPath(StructureShape $shape, string $path, string $out
 
     protected function parseResponseTimestamp(Shape $shape, string $input, bool $required): string
     {
-        $body = 'new \DateTimeImmutable((string) INPUT)';
+        $body = 'new \DateTimeImmutable((string) (is_numeric(INPUT) ? "@{INPUT}" : INPUT))';
```
tyx, line 114: Allow to handle timestamp unix with microseconds

tyx:
I reverted my change in generated code.

I succeed in doing one change in COdeGenerator to handle my timestamp with microseconds but the others PHPStan/Psalm failure looks more weird.

jderusse:
> So I just need to remove the integration folder ?

adds a `markTestIncomplete` on top of the test (You'll find several occurrences in other services)


> But I just see that php 7.2 does not support `DateTime::createFromFormat` disappointed

Hmm... I believe this should work. See https://3v4l.org/DlCbH#v7.2.34

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
```diff
@@ -111,7 +111,7 @@ public function generateForPath(StructureShape $shape, string $path, string $out
 
     protected function parseResponseTimestamp(Shape $shape, string $input, bool $required): string
     {
-        $body = 'new \DateTimeImmutable((string) INPUT)';
+        $body = 'is_numeric(INPUT) ? \Datetime::createFromFormat("U.u", sprintf("%.6F", INPUT)) : new \DateTimeImmutable((string) INPUT)';
```
jderusse, line 114: What is the prupose of the check on `is_numeric` do you have example of non numeric data?

tyx: Well I did not want to break anything on other Client as I don't know how other timestamp are returned by aws !

But if we are sure we have real timestamp everywhere I can remove it no problem !

jderusse: I think the `sprintf` thing would fix the input. https://3v4l.org/lXAQc#v7.2.34

jderusse: hmmm... looks like this is already what we are doing in `JsonRpcParser`

tyx:
Thanks @jderusse I succeeded in fixing the whole CI

src/Core/src/AwsClientFactory.php
```diff
@@ -292,6 +293,19 @@ public function iam(): IamClient
         return $this->serviceCache[__METHOD__];
     }
 
+    public function iot(): IotClient
```
jderusse, line 296: Please add an entry in CHANGELOG

src/Integration/Symfony/Bundle/src/DependencyInjection/AwsPackagesProvider.php
```diff
@@ -73,6 +73,10 @@ public static function getAllServices(): array
                 'class' => \AsyncAws\Iam\IamClient::class,
                 'package' => 'async-aws/iam',
             ],
+            'iot' => [
```
jderusse, line 76: Please add an entry in CHANGELOG

jderusse (commented):
Thank you for this contribution. 

I've just 2 minor comments about adding references in related CHANGELOG files

src/Core/src/AwsClientFactory.php
tyx: done

PR #1286, opened 2022-08-22
Update DynamoDbClient.php

jusufazer:
add endpoint configuration for local dynamodb environment


GrahamCampbell:
This file cannot be edited in this way - it is auto-generated. You'll need to update the code generator. You can find more information in the docs and in the code.

jderusse (commented):
I don't think this change is needed.
You should be able to use the endpoint parameter in the current version of async-aws/dynamodb

jderusse:
closing this PR. AsyncAws already expose a parameter to configure the endpoint

Issue #1264, opened 2022-06-23
Timestream error 500 can trigger `ErrorException` when building `InternalServerException`


GrahamCampbell:
```
ErrorException Undefined array key "message" 
    vendor/voxie/async-aws-timestream-write/src/Exception/InternalServerException.php:17 Illuminate\Foundation\Bootstrap\HandleExceptions::handleError
    vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php:257 Illuminate\Foundation\Bootstrap\HandleExceptions::Illuminate\Foundation\Bootstrap\{closure}
    vendor/voxie/async-aws-timestream-write/src/Exception/InternalServerException.php:17 AsyncAws\TimestreamWrite\Exception\InternalServerException::populateResult
    vendor/voxie/async-aws-core/src/Exception/Http/HttpExceptionTrait.php:50 AsyncAws\Core\Exception\Http\ClientException::__construct
    vendor/voxie/async-aws-core/src/Response.php:406 AsyncAws\Core\Response::AsyncAws\Core\{closure}
    vendor/voxie/async-aws-core/src/Response.php:423 AsyncAws\Core\Response::getResolveStatus
    vendor/voxie/async-aws-core/src/Response.php:123 AsyncAws\Core\Response::resolve
    vendor/voxie/async-aws-core/src/Response.php:105 AsyncAws\Core\Response::__destruct
```

PR #1313, opened 2022-10-11
[Timestream] Set message attribute only if message is present during InternalServerException

bikalbasnet:
Fixes https://github.com/async-aws/aws/issues/1264


GrahamCampbell:
This can't be achieved by editing the generated code, unfortunately. The actual code generator needs to be changed.

PR #1329, opened 2022-11-13
Fixed issue with exception generation

### LINKED ISSUE #1264: Timestream error 500 can trigger `ErrorException` when building `InternalServerException`

GrahamCampbell:
Strictly speaking, this is a spec compliance bug with the server-side implementations, but getting Amazon to anything is impossible. Fixes #1264.

### DIFF
--- a/src/CodeGenerator/src/Definition/StructureMember.php
+++ b/src/CodeGenerator/src/Definition/StructureMember.php
@@ -22,6 +22,10 @@ class StructureMember extends Member
 
     public function isRequired(): bool
     {
+        if ($this->getOwnerShape() instanceof ExceptionShape && \in_array(strtolower($this->data['_name']), ['code', 'message'], true)) {
+            return false;
+        }
+
         return $this->data['_required'];
     }
 
--- a/src/Service/Sns/src/Exception/ValidationException.php
+++ b/src/Service/Sns/src/Exception/ValidationException.php
@@ -16,6 +16,8 @@ final class ValidationException extends ClientException
         if (0 < $data->Error->count()) {
             $data = $data->Error;
         }
-        $this->message = (string) $data->message;
+        if (null !== $v = (($v = $data->message) ? (string) $v : null)) {
+            $this->message = $v;
+        }
     }
 }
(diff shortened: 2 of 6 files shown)


src/CodeGenerator/src/Definition/StructureMember.php
```diff
@@ -22,6 +22,10 @@ public function getOwnerShape(): StructureShape
 
     public function isRequired(): bool
     {
+        if ($this->getOwnerShape() instanceof ExceptionShape && \in_array(strtolower($this->data['_name']), ['code', 'message'], true)) {
```
jderusse, line 25: Looking at generated code, it looks like all occurrences are `messages` (code is never required)

jderusse (approved):
I think thats the best move we can do :(

Is it a BC break to have a null value for message? Shouldn't we fallback to something else (ie. `?? ''`) ?

src/CodeGenerator/src/Definition/StructureMember.php
GrahamCampbell: I think this is only because of what we have currently generated. I think in principal newly generated options could use code. I copied what was done for `isNullable`.

GrahamCampbell:
> Is it a BC break to have a null value for message? Shouldn't we fallback to something else (ie. ?? '') ?

I don't think we're breaking anything. If we decide to skip setting message, it is already set to `''`.

jderusse:
thank you

PR #1336, opened 2022-11-24
Implemented some additional S3 methods

GrahamCampbell:
WIP - no tests updated yet.


manifest.json
```diff
@@ -448,16 +448,20 @@
...
                 "GetBucketCors",
                 "GetBucketEncryption",
+                "GetBucketTagging",
+                "GetBucketVersioning",
                 "GetObject",
                 "GetObjectAcl",
                 "HeadObject",
                 "ListBuckets",
                 "ListMultipartUploads",
+                "ListObjects",
```
jderusse, line 458: By curiosity, what is the use-case for this method? And why not using `ListObjectsV2`?

Don't forget to update the changelog BTW :wink:

GrahamCampbell:
I don't have capacity to finish this, and no longer have a use for it. I already used it for a one-time script. ;)

PR #1350, opened 2022-12-29
add `putuserpolicy`, `deleteuserpolicy`, `createservicespecificcredential` to `IAM` client

nealio82:
* Add generated `PutUserPolicy` method
* Add generated `DeleteUserPolicy` method
* Add generated `CreateServiceSpecificCredential` method
* Fill in tests
* Add documentation


nealio82:
```
#  One of two things are wrong here:
#   1) You have modified generated code.
#   2) You have modified the code generator and not regenerated the code.
```

Neither of these things are true 🤔 

```
diff --git a/src/Service/Iam/src/Enum/StatusType.php b/src/Service/Iam/src/Enum/StatusType.php
index 54f0223..9cddacc 100644
--- a/src/Service/Iam/src/Enum/StatusType.php
+++ b/src/Service/Iam/src/Enum/StatusType.php
@@ -3,8 +3,7 @@
 namespace AsyncAws\Iam\Enum;
 
 /**
- * The status of the service-specific credential. `Active` means that the key is valid for API calls, while `Inactive`
- * means it is not.
+ * The status of the access key. `Active` means that the key is valid for API calls, while `Inactive` means it is not.
  */
```

Does this mean that the Enum's being re-used and there's a race condition in generating the comment at the top?

jderusse:
> ```
> #  One of two things are wrong here:
> #   1) You have modified generated code.
> #   2) You have modified the code generator and not regenerated the code.
> ```
> 
> Neither of these things are true thinking
> 
> ```
> diff --git a/src/Service/Iam/src/Enum/StatusType.php b/src/Service/Iam/src/Enum/StatusType.php
> index 54f0223..9cddacc 100644
> --- a/src/Service/Iam/src/Enum/StatusType.php
> +++ b/src/Service/Iam/src/Enum/StatusType.php
> @@ -3,8 +3,7 @@
>  namespace AsyncAws\Iam\Enum;
>  
>  /**
> - * The status of the service-specific credential. `Active` means that the key is valid for API calls, while `Inactive`
> - * means it is not.
> + * The status of the access key. `Active` means that the key is valid for API calls, while `Inactive` means it is not.
>   */
> ```
> 
> Does this mean that the Enum's being re-used and there's a race condition in generating the comment at the top?

AWS has a weird way to add comments on the objects.
Actually,m the comment is set on the relation caller -> callee. That's why sometimes the comment changes when adding a new method. 
To fix the issue, run `generate --all` to make the CI happy

nealio82:
@jderusse passing now :)

Kinda frustrating that Localstack doesn't support `CreateServiceSpecificCredential` though :/

jderusse (approved):

jderusse:
thanks you @nealio82

PR #1353, opened 2023-01-02
Update php-cs-fixer

jderusse:


src/Service/S3/src/Input/CompleteMultipartUploadRequest.php
```diff
@@ -251,32 +252,24 @@ public function request(): Request
         $headers = ['content-type' => 'application/xml'];
         if (null !== $this->checksumCrc32) {
             $headers['x-amz-checksum-crc32'] = $this->checksumCrc32;
-        }
-        if (null !== $this->checksumCrc32C) {
+        }if (null !== $this->checksumCrc32C) {
```
stof, line 255: This CS change looks really weird to me.

GrahamCampbell: This is more of an issue of php-cs-fixer not adding the new line, than it removing it. The issue with the code generator was fixed in https://github.com/async-aws/aws/commit/e5724c9165a8ec2f71a84cef08dc7d45f7a4e7fa.

stof: The diff in https://github.com/async-aws/s3/compare/1.13.0...master#diff-c3b87b583e61012cbf9b5726c2b83aa643b57c7bb7fb66d442860f719422159fL777 (current unreleased changes in the S3 client) make me think that this is not fully fixed.

GrahamCampbell: Heh, good spot. I guess there's another newline missing somewhere in the code gen.

jderusse: addressed by https://github.com/async-aws/aws/pull/1382
thanks for the report

PR #1354, opened 2023-01-02
Upgrade PSALM

jderusse:


src/Core/src/EnvVar.php
```diff
@@ -17,10 +17,10 @@ public static function get(string $name): ?string
     {
         if (isset($_ENV[$name])) {
             // variable_order = *E*GPCS
-            return $_ENV[$name];
-        } elseif (isset($_SERVER[$name]) && 0 !== strpos($name, 'HTTP_')) {
+            return (string) $_ENV[$name];
+        } elseif (isset($_SERVER[$name]) && !\is_array($_SERVER[$name]) && 0 !== strpos($name, 'HTTP_')) {
```
GrahamCampbell, line 21: Surely we want to check is string, otherwise strpos will fail?

GrahamCampbell (changes requested):

src/Core/src/EnvVar.php
jderusse: `strpos` is used for `$name` (which is typehinted in method declaration), `is_array` is used for `$_SERVER[$name]`

GrahamCampbell: 👌

PR #1372, opened 2023-02-17
Fixed broken query string encoding

GrahamCampbell:
`http_build_query` defaults to building strings for form request bodies, not URI query strings. We have to put it into that mode using the 4th param.

### DIFF
--- a/src/CodeGenerator/src/Generator/TestGenerator.php
+++ b/src/CodeGenerator/src/Generator/TestGenerator.php
@@ -104,7 +104,7 @@ class TestGenerator
 
                 break;
             case 'query':
-                $stub = substr(var_export($exampleInput ? (\is_array($exampleInput) ? http_build_query($exampleInput, '', '&', \PHP_QUERY_RFC1738) : $exampleInput) : "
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
Oh, I think the encoding in the V4 signer is also wrong, and only worked because the encoding here was wrong also.

GrahamCampbell:
Probably nobody has noticed before since it would be rare to use a space character in a query string, and may AWS services probably accept the wrong encoding, and since it was encoded wrong both when signing and constructing, the signature was good.

GrahamCampbell:
image
image

GrahamCampbell:
Moving to draft until I have had time to try this out for really realz.

GrahamCampbell:
Many AWS services seem to support the incorrect encoding, though I think not all. This is ready for review now.

jderusse:
the AWS SDK use `PHP_QUERY_RFC3986`. What is the difference with `PHP_QUERY_RFC1738`?

GrahamCampbell:
OK, fixed now. This explains why the signing was working. The query encoding was correct when doing the signing, so when AWS normalized the query string on their side after our incorrect encoding, the signature matched because the correct encoding algorithm was used when we calculated the signature.

jderusse (approved):

PR #1375, opened 2023-02-28
Fix composer.json CS

jderusse:


composer.json
```diff
@@ -11,7 +11,7 @@
         "ext-hash": "*",
         "ext-json": "*",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
-        "symfony/http-client": "^4.4.16 || ^5.1.7,!=5.2.0 || ^6.0",
+        "symfony/http-client": "^4.4.16 || ^5.1.7 || ^6.0",
```
GrahamCampbell, line 14:
```
"~4.4.16 || ~5.1.7 || ^5.2.1 || ^6.0"
```

composer.json
```diff
@@ -27,7 +27,7 @@
         "league/mime-type-detection": "^1.0",
         "matthiasnoback/symfony-config-test": "^4.1",
         "monolog/monolog": "^1.1 || ^2.0",
-        "nette/php-generator": "^3.3,!=3.6.3",
+        "nette/php-generator": "^3.3",
```
GrahamCampbell, line 30:
```
"^3.6.4"
```

composer.json
```diff
@@ -43,28 +43,38 @@
         "symfony/phpunit-bridge": "^5.3",
         "symfony/polyfill-uuid": "^1.0"
     },
+    "conflict": {
+        "nette/php-generator": "3.6.3",
```
GrahamCampbell, line 47: unforuntately, this conflict applies when this package is required, but this is only a dev dep, unecessarily preventing the install of this dep.

GrahamCampbell (changes requested):

PR #1380, opened 2023-03-13
implement IMDSv2

### LINKED ISSUE #1163: IMDSv2 support
Not sure if this is already the case in async-aws but it is preferred to use Instance metadata v2 over v1 in ami linux 2. I couldn't find any related issue or PR referencing it.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

We are using symfony messenger sqs v5.4 and receive error when we disable IMDSv1 on ec2 instances.

DavidBadura:
fix #1163

Here is an implementation for IMDSv2 that we have in use. In the end we adjusted the `InstanceProvider` as described in the documentation:

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

~In this implementation I replace the IMDSv1 variant with IMDSv2. I wonder if we really still need IMDSv1 or if IMDSv2 is available everywhere.~
It will first try to do an IMDSv2 authentication, if it doesn't work, then it will fallback to IMDSv1

I look forward to feedback.

### DIFF
--- a/docs/authentication/ec2-metadata.md
+++ b/docs/authentication/ec2-metadata.md
@@ -10,5 +10,5 @@ When you run code within an EC2 instance (or EKS, Lambda), AsyncAws is able to f
 When running a single application on the Server, this is the simplest way to grant permissions to the application. You
 have nothing to configure on the application, you only grant permissions on the Role attached to the instance.
 
-AsyncAWS uses the IMDSv1 protocol to read instance metadata. So this authentication method won't work if the v1
+AsyncAWS uses the IMDSv2 protocol to read instance metadata. So this authentication method won't work if the v2
 protocol is disabled.
--- a/docs/authentication/ec2-metadata.md
+++ b/docs/authentication/ec2-metadata.md
@@ -10,5 +10,4 @@ When you run code within an EC2 instance (or EKS, Lambda), AsyncAws is able to f
 When running a single application on the Server, this is the simplest way to grant permissions to the application. You
 have nothing to configure on the application, you only grant permissions on the Role attached to the instance.
 
-AsyncAWS uses the IMDSv2 protocol to read instance metadata. So this authentication method won't work if the v2
-protocol is disabled.
+AsyncAWS uses the IMDSv2 protocol to read instance metadata and fallback to the IMDSv1 protocol if it isn't supported.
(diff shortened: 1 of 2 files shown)


docs/authentication/ec2-metadata.md
```diff
@@ -10,5 +10,5 @@ When you run code within an EC2 instance (or EKS, Lambda), AsyncAws is able to f
 When running a single application on the Server, this is the simplest way to grant permissions to the application. You
 have nothing to configure on the application, you only grant permissions on the Role attached to the instance.
 
-AsyncAWS uses the IMDSv1 protocol to read instance metadata. So this authentication method won't work if the v1
+AsyncAWS uses the IMDSv2 protocol to read instance metadata. So this authentication method won't work if the v2
```
jderusse, line 13: That's a BC break for people using AsyncAws with `IMDSv1` and disabling `IMDSv2`
Is there a way to try `IMDSv2` and deal with errors (from the call to `TOKEN_ENDPOINT`) to decide if we could fallback to `IMDSv1~

DavidBadura: I adjusted the code. This is also how the AWS documentation describes how their SDKs work:

> The AWS SDKs use IMDSv2 calls by default. If the IMDSv2 call receives no response, the SDK retries the call and, if still unsuccessful, uses IMDSv1. This can result in a delay.

I only left out the retry because the lib uses a retry http client decorator.

TomBerriot:
Hello there, is this an implementation of IMDS v2 only for a specific part of the `async-aws/aws` project or is this intended to be a full support ? 

We just had a bug enabling IMDS v2 when Invoking our function through bref cli and here is the error we received

```
AWS_DEFAULT_REGION=eu-west-3 vendor/bin/bref cli NAME_OF_LAMBDA -- doctrine:migrations:migrate --no-interaction


In Response.php line 422:
                                                                               
  HTTP 403 returned for "https://lambda.eu-west-3.amazonaws.com/2015-03-31/fu  
  nctions/production-backend-php-console/invocations".                         
                                                                               
  Code:    MissingAuthenticationTokenException                                 
  Message: Missing Authentication Token                                        
  Type:                                                                        
  Detail:    
```
I suspect the code making this call has its request built by the `InvocationRequest.php` class

 https://github.com/async-aws/lambda/blob/master/src/Input/InvocationRequest.php#L157

made surely by the `invoke` function of `LambdaClient`

https://github.com/async-aws/lambda/blob/master/src/LambdaClient.php#L193

jderusse:
Thank you @DavidBadura for taking care of this.

Truly sorry for the delay

PR #1385, opened 2023-03-29
minor: update link in docblock

kbond:
I had a bit of confusion when determining if this client used SES v2 or not.


GrahamCampbell:
I think those blocks are auto-generated, so this may not be possible to change (as it will get changed back). Hard to know, since GitHub Actions is experiencing an outage. ;)

kbond:
Ah, ok. Not a big deal - feel free to close if that's the case.

jderusse (approved):
I've pushed a commit to fix the typo in the Manifest (used to generate the links)

PR #1401, opened 2023-05-06
[Rekognition] Add DetectModerationLabels API

t-richard:
This adds the DetectModerationLabels that allows to detect inappropriate images

See: 
- https://docs.aws.amazon.com/rekognition/latest/APIReference/API_DetectModerationLabels.html
- https://docs.aws.amazon.com/rekognition/latest/dg/procedure-moderate-images.html

Will try this for real later this weekend

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -430,6 +430,7 @@
                 "DeleteCollection",
                 "DeleteProject",
                 "DetectFaces",
+                "DetectModerationLabels",
                 "GetCelebrityInfo",
                 "IndexFaces",
                 "ListCollections",
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -47,6 +47,7 @@ parameters:
               - src/Service/Kinesis/src/ValueObject
               - src/Service/Lambda/src/ValueObject
               - src/Service/Iot/src/ValueObject/Tag.php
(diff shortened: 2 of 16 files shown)


t-richard:
Just tested it live and it works as expected for my use case !

manifest.json
```diff
@@ -430,6 +430,7 @@
                 "DeleteCollection",
                 "DeleteProject",
                 "DetectFaces",
+                "DetectModerationLabels",
```
jderusse, line 433: Could you add an entry in the `src/Service/Rekognition/CHANEGELOG.md` file :pray:

jderusse (approved):

manifest.json
t-richard: Done, there was already a `NOT RELEASED` entry in the changelog so I added it there, let me know if you want me to write a specific version (probably 0.1.8)

jderusse (approved):

manifest.json
jderusse: perfect

jderusse:
thank you @t-richard

t-richard:
Thanks for the great work on async-aws :heart: 

Love how easy it was to add a new operation and contribute it !

PR #1402, opened 2023-05-08
Enforce HTTPS with TLS 1.2 or higher

GrahamCampbell:
I have also required curl in order to do this in a simple way. I don't feel this is much of a restriction in practice, since the Symfony HTTP client is only "async" if curl is loaded, or another package is installed. Curl is pretty ubiquitous these days.

### DIFF
--- a/src/Core/src/HttpClient/AwsHttpClientFactory.php
+++ b/src/Core/src/HttpClient/AwsHttpClientFactory.php
@@ -12,11 +12,25 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
  */
 class AwsHttpClientFactory
 {
+    public static function createClient(): HttpClientInterface
+    {
+        return return new CurlHttpClient([
+            'extra' => [
+                'curl' => [
+                    \CURLOPT_PROTOCOLS => \CURLPROTO_HTTPS,
+                    \CURLOPT_REDIR_PROTOCOLS => \CURLPROTO_HTTPS
+                    \CURLOPT_SSLVERSION => \CURL_SSLVERSION_TLSv1_2,
+                ],
+            ]
+        ]);
+    }
+
     public static function createRetryableClient(HttpClientInterface $httpClient = null, LoggerInterface $logger = null): HttpClientInterface
     {
         if (null === $httpClient) {
-            $httpClient = HttpClient::create();
+            $httpClient = self::createClient();
         }
+
         if (class_exists(RetryableHttpClient::class)) {
             /** @psalm-suppress MissingDependency */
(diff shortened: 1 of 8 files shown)


src/Core/src/HttpClient/AwsHttpClientFactory.php
```diff
@@ -12,11 +12,23 @@
...
 {
+    public static function createClient(): HttpClientInterface
+    {
+        return new CurlHttpClient([
+            'extra' => [
+                'curl' => [
+                    \CURLOPT_SSLVERSION => \CURL_SSLVERSION_TLSv1_2,
+                ],
+            ],
+        ]);
```
jderusse, line 23: I'm not sure we should force people to use an extension
what about making curl a soft dependency with something like (not tested):

```suggestion
        $options = [];
        if (\extension_loaded('curl')) {
           $options = [
                'extra' => [
                    'curl' => [
                        \CURLOPT_SSLVERSION => \CURL_SSLVERSION_TLSv1_2,
                    ],
                ],
            ];
        }

        return HttpClient::create($options);
        
```

What do you think @nicolas-grekas

GrahamCampbell: If we do this, then it doesn't enforce TLS 1.2. I'd rather have it fail loud if curl is not installed, than silently fallback. If we were to have it silently fallback, this would not solve the issue I created the PR to solve.

GrahamCampbell: > I'm not sure we should force people to use an extension

If people are not using curl, they can explicitly opt out by passing a client themselves. That seems like a more secure and robust default than silently not being async or not requiring TLS 1.2.

nicolas-grekas:
I think this should be improved at the Symfony client level.

jderusse:
> We will need to drop support for PHP older than 8.1 if we are to rely on symfony to do this.

IMHO, this is not our duty to force people to install curl or upgrade PHP in order to use a higher protocol version. (because by adding `ext-curl` in the composer.json file, we require people to install the extension, even if they want to inject their own HttpClient implementation)

At best, we could add the `crypto_method` option (if Nicolas' PR is merged) when they use the Symfony 6.3 Client, or the `CURLOPT_SSLVERSION` option when the curl extension is loaded, and add a "suggest" section in composer.json.

GrahamCampbell:
The symfony http client will default to TLS 1.2 or higher, from 6.3.0 onwards, if that hits 6.3, otherwise, 6.4.0/7.0.0.

PR #1407, opened 2023-05-22
Corrected athena change log

GrahamCampbell:
Removed two unreleased entries that cancel each other out.


jderusse (approved):
good catch :pray:

PR #1415, opened 2023-05-30
Add missing request information in discovered endpoint

jderusse:


stof (approved):

src/Core/src/AbstractApi.php
```diff
@@ -303,7 +303,12 @@ private function getDiscoveredEndpoint(string $uri, array $query, ?string $regio
             }
         }
 
-        return $endpoint;
+        $endpoint .= $uri;
+        if (empty($query)) {
+            return $endpoint;
+        }
+
+        return $endpoint . (false === strpos($endpoint, '?') ? '?' : '&') . http_build_query($query);
```
GrahamCampbell, line 311: This is incorrect use of `http_build_query`. The default params build with syntax for a form request body, not a query string for a URI.

PR #1416, opened 2023-05-30
Update the GitHub Actions configuration

stof:
This makes 2 changes:

- use `actions/checkout@v3` to avoid using the deprecated node12 runtime
- updates the runner to `ubuntu-latest` instead of `ubuntu-20.04` so that it uses the 22.04 runner (and updates to new ones when available)


GrahamCampbell:
👎 for using ubuntu-latest. It prevents us from controlling when upgrades happen. I'd highly recommend switching to ubuntu-22.04.

PR #1418, opened 2023-05-30
Generates proper documentation for classes and properties

jderusse:
The AWS SDK provides 2 sections for each doc entries:
- the 'main' section that is used to describe the Object
- the 'refs' section is a list that contains one entry by property using the object.

In the previous implementation, we used the refs section when available to define the documentation of the `Class`. And a shorter version of the same documentation for each usage.

This is "prettier". But has 2 drawbacks:
- 1. we lose some information (ie, when 2 different properties reference the same object, they might have 2 different descriptions)
- 2. the Class's description vary with the first property that use the class.

This PR change the behavior to use the `main` description for the class. And the **full** `refs` description for the properties.


src/CodeGenerator/src/Generator/GeneratorHelper.php
```diff
@@ -111,6 +113,7 @@ public static function parseDocumentation(string $documentation, bool $short = t
         $spaceNext = false;
 
         // converts <li> into `- ` AND handle multi-level list
+        $counter = null;
```
stof, line 116: this way of handling the counter does not seem to be compatible with nested lists (while the comment on the previous line says we support it)

jderusse: good catch.

stof (approved):

src/Core/src/Sts/Input/AssumeRoleRequest.php
```diff
@@ -22,6 +22,13 @@ final class AssumeRoleRequest extends Input
 
     /**
      * An identifier for the assumed role session.
+     * Use the role session name to uniquely identify a session when the same role is assumed by different principals or for
```
GrahamCampbell, line 25: I would prefer it if we could separate the first line from the rest with a blank line.

```
    /**
     * An identifier for the assumed role session.
     *
     * Use the role session name to uniquely identify a session when the same role is assumed by different principals or for
     *
```

In the original psr draft phpdoc spec, the first paragraph has special meaning as the summary description.

GrahamCampbell (commented):
Is this really necessary? Doesn't this add a lot of extra bytes to the packages.

jderusse:
> Is this really necessary? Doesn't this add a lot of extra bytes to the packages.

I have the same feeling.

I wonder if we shouldn't just use the first line, and let people browse the official documentation if they want more details...

stof:
@jderusse what is your decision on that ? I would like to know if I need to improve the doc generation in #1417 or if it will be unnecessary once this work is done if the doc is removed anyway (and so I should just wait instead).

src/Core/src/Sts/Input/AssumeRoleRequest.php
Nyholm: This should be done by php-cs-fixer. 

They usually help us so we don’t need to add any logic for formatting.

Nyholm (approved):
I appreciate the extra comments. 

But I do hear the arguments about extra size. Can we try it out? How much more size is s3 or sqs client taking with this pr?

Nyholm:
Lovely. 

I suggest to merge this pr

jderusse:
let's move forward.
I've added 3 changes to this PR:
- `<a href>` tags in the doc are now converted to [footnote](https://www.markdownguide.org/extended-syntax/#footnotes) (previously rendered in `@see ... ` tags)
- content is html_entity_decoded (code snippets are encoded in the original doc)
- extra HTML detection now use `</` or `/>` instead of `<`

src/Core/src/Sts/Input/AssumeRoleRequest.php
stof: The GeneratorHelper removes empty lines in `$s = preg_replace('/\n+/', "\n", $s);`. Maybe removing this would be enough.

stof (approved):

PR #1419, opened 2023-05-30
Ignore vendor's deprecations

jderusse:
Fixes error

```
1x: Method "Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface::process()" might add "void" as a native return type declaration in the future. Do the same in implementation "Nyholm\BundleTest\CompilerPass\PublicServicePass" now to avoid errors or add an explicit @return annotation to suppress this message.
      1x in BundleInitializationTest::setUp from AsyncAws\Symfony\Bundle\Tests\Functional
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "monolog/monolog": "^1.1 || ^2.0",
         "nette/php-generator": "^3.6.4",
         "nette/utils": "^3.0",
-        "nyholm/symfony-bundle-test": "1.x-dev#0555e4898e1b05655a4bdf52ba10c7f56611a514 as 1.9.0",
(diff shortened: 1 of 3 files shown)


stof:
alternatively, you could send a PR to @Nyholm's repository :smile:

stof:
Actually, the proper fix is to upgrade to v2 of `nyholm/symfony-bundle-test` as those deprecations are already fixed.

stof:
@jderusse I have the migration in #1420 (and now, I see that you pushed a similar change without changing your PR title, but your change is incomplete and makes the CI fail)

jderusse:
closing in favor of  #1420

PR #1420, opened 2023-05-31
Upgrade to nyholm/symfony-bundle-test v2

stof:
This avoids getting deprecations when running against Symfony 6.3 (the unmaintained 1.x branch did not get patched for it).

Note that I needed to copy the `PublicServicePass` that got removed from the bundle. This is because the kind of assertions being done in those tests would actually better fit with https://github.com/SymfonyTest/SymfonyDependencyInjectionTest


jderusse:
thank you

PR #1423, opened 2023-06-02
Clean outdated errors from the psalm baseline

stof:
This regenerates the Psalm baseline to get rid of errors that are not relevant anymore. This has 2 benefits:

- if a PR regenerates the baseline to add new errors in it, it won't have a huge diff removing those
- if a PR changes code and introduces new issues, there is no risk to ignore them silently if we are unlucky that it matches an old baseline entry


jderusse (approved):

jderusse:
thanks

Issue #1414, opened 2023-05-30
Correct approach for endpoint discovery for the MediaConvert service


stof:
I'm looking into adding support for the MediaConvert service.
This service requires account-specific endpoints. However, unlike DynamoDB and Timestream, those endpoints are not expiring ones and the official SDK does not implement automatic endpoint discovery for them.

Here is the suggested code for the official SDK:

```php
$client = new Aws\MediaConvert\MediaConvertClient([
    'profile' => 'default',
    'version' => '2017-08-29',
    'region' => 'us-east-2'
]);

//retrieve endpoint
try {
    $result = $client->describeEndpoints([]);
} catch (AwsException $e) {
    // handle error
}

$single_endpoint_url = $result['Endpoints'][0]['Url'];

print("Your endpoint is " . $single_endpoint_url);

//Create an AWSMediaConvert client object with the endpoint URL that you retrieved: 
$mediaConvertClient = new MediaConvertClient([
    'version' => '2017-08-29',
    'region' => 'us-east-2',
    'profile' => 'default',
    'endpoint' => $single_endpoint_url
]);

// Use this second client for other API call
```

The retrieval of the endpoint URL is meant to be cached: https://docs.aws.amazon.com/mediaconvert/latest/apireference/getting-started.html

As this is cumbersome to use, I thought that maybe the automatic discovery could be applied to that service as well. The difference with other services using the automatic ones is the structure of the `DescribeEndpoints` response:
- the URL is in a field called `Url`, not `Address`
- there is no cache period in minutes, as this is supposed to have an infinite TTL

To reuse the existing infrastructure, I suggest that the CodeGenerator could insert additional methods in the MediaConvert Endpoint value object looking like that:

```php
class Endpoint
{
    // ...

    public function getAddress(): string
    {
        return $this->getUrl();
    }

    public function getCachePeriodInMinutes(): int
    {
        // 365 days of caching, as it actually does not expire. 
        // Not using PHP_INT_MAX avoids integer overflowing when computing the expiration timestamp
        return 525600;
    }
}
```

However, I don't know whether this would be accepted or what would be the best way to configure the code generator for that.
What do you think about that ?

jderusse:
I agree with you: it would be a nice addition.

But to reuse the current code, it requires some changes:

[...]

PR #1424, opened 2023-06-02
Fix the OperationGenerator for services with a custom endpoint discovery

stof:
Such services will not have an operation marked as being the `endpointoperation` but will instead implement the `discoverEndpoint` manually.
The `findEndpointOperationName` method returns `Operation|null` but the code was not properly handling `null` values.


stof:
Note that this is the kind of mistake that could be detected by phpstan or psalm, but they don't run at the full capacity right now, and the null checks are part of their disabled rules.

stof:
This is necessary for MediaConvert, as #1414 decided to use a custom discovery implementation.

jderusse (approved):

PR #1425, opened 2023-06-02
Add support for generating code for discovered endpoint using the URI

stof:
The core package has been updated to support those, so the code generator can rely on them.
This makes the RequirementRegistry smarter by merging requirements for the same package added in multiple places by keeping the requirement with the highest lower bound. For now, this supports only the `*` requirement and requirements using the `^` operator as they are the only requirements used by the code generators.

This leverages #1415 in the code generator (this PR assumes that the next release of Core including this patch will be 1.19 as this deserves a minor version IMO as it brings a new feature). The changelog entry added in Core actually corresponds to the change in #1415.


stof:
This is needed for the usage of endpoint discovery in MediaConvert, which uses parameters in the URL path.

jderusse (approved):
much better!

PR #1440, opened 2023-06-13
Upgrade to PHP-CS-Fixer 3.17

jderusse:
Also add a rule to trim blank line inside `{}`, `[]` and `()`


src/CodeGenerator/src/Generator/CodeGenerator/PopulatorGenerator.php
```diff
@@ -56,7 +56,7 @@ class PopulatorGenerator
      */
     private $parserProvider;
 
-    public function __construct(ClassRegistry $classRegistry, NamespaceRegistry $namespaceRegistry, RequirementsRegistry $requirementsRegistry, ObjectGenerator $objectGenerator, ?TypeGenerator $typeGenerator = null, ?EnumGenerator $enumGenerator = null, ?ParserProvider $parserProvider = null)
+    public function __construct(ClassRegistry $classRegistry, NamespaceRegistry $namespaceRegistry, RequirementsRegistry $requirementsRegistry, ObjectGenerator $objectGenerator, TypeGenerator $typeGenerator = null, EnumGenerator $enumGenerator = null, ParserProvider $parserProvider = null)
```
GrahamCampbell, line 59: I'd kind rather we always enforce that the type should be nullable. I d'no why symfony has gone with this choice. Seems kinda wrong IMO.

GrahamCampbell (changes requested):

src/CodeGenerator/src/Generator/CodeGenerator/PopulatorGenerator.php
jderusse: let's make it explicit

PR #1443, opened 2023-06-13
feat: Sns Client ListTopics

Ferror:


jderusse:
Do you need help to finish this RP?

Ferror:
> Do you need help to finish this RP?

I'm learning how everything around here works. For now it's mostly time issue 😄

src/Service/Sns/tests/Integration/SnsClientTest.php
```diff
@@ -135,8 +135,8 @@ public function testListTopics(): void
 
         $result->resolve();
 
-        // self::assertTODO(expected, $result->getTopics());
-        self::assertSame('change me', $result->getNextToken());
+        self::assertNull($result->getNextToken());
+        self::assertCount(2, $result->getTopics());
```
GrahamCampbell: we should also look at the actual topics in the result

GrahamCampbell (changes requested):

src/Service/Sns/tests/Integration/SnsClientTest.php
Ferror: Is it worth checking out, as the https://github.com/async-aws/aws/blob/b7cbfcbcc7258ad0bef10c7899bf830b8c3a7e2e/src/Service/Sns/tests/Unit/Result/ListTopicsResponseTest.php test already testing the response?

stof: Integration tests and unit tests provide a different level of guarantees

stof:
@Ferror you should rebase your branch on the latest master and re-run the code generator, to make sure that our new generated benefits from all the improvements done recently in the code generator (and anyway, the CI won't be green without that).

Ferror:
> @Ferror looks like you rebased but without re-running the code generator

Just regenerated the ListTopic action. Btw I think it's breaking the phpdocs

Screenshot 2023-07-04 at 17 30 32

stof:
@Ferror phpstorm has a bug, not liking string literals containing a `@` in array shapes. But the phpdoc is valid.

stof:
For reference, I reported the PhpStorm bug at https://youtrack.jetbrains.com/issue/WI-73316/Array-shapes-in-phpdoc-with-a-key-containing-a-report-weird-error-saying-Expected-DOCCOMMENTEND

You can upvote it if you want.

stof (approved):

stof:
@Ferror looks like the last assertion you added fails

Ferror:
> @Ferror looks like the last assertion you added fails

Seems I had eu-west-1 region locally. Changed it

src/Service/Sns/src/SnsClient.php
```diff
@@ -265,8 +264,7 @@ public function listSubscriptionsByTopic($input): ListSubscriptionsByTopicRespon
      *
      * @param array{
      *   NextToken?: string,
-     *
-     *   @region?: string,
+     *   '@region'?: string|null.
```
GrahamCampbell: period should be removed

GrahamCampbell (changes requested):

GrahamCampbell (approved):

PR #1454, opened 2023-06-22
Fix some issues in the code generator

stof:
- the generated client test was missing a use statement for the client class, making it a bad DX when adding a new client (after that, the use statement would probably have been added manually to make test pass without a fatal error, but we could avoid the step of the fatal error)
- the generated test assertions could use a wrong case for the getter name as they were not using the same logic than the PopulatorGenerator (this was working due to PHP method names being case insensitive but it was not nice)
- when loading a manifest which does not define variables (as part of my experiment generating an out-of-tree component), the GenerateCommand was failing with an array-to-string conversion due to using a bad fallback value.


jderusse (approved):

jderusse:
thank you

PR #1457, opened 2023-06-22
Generate valid phpdoc param tags

stof:
When an array shape key starts with a `@` like `@region`, it must be quoted in the phpdoc as the unquoted one would be interpreted as the next phpdoc tag.

Due to the invalid tag, static analysis tools would not have any type for those arguments, defeating the type-safety benefit documented in the readme (we would have it for value objects, but not for the input objects that are the main entry point as they all have a `@region` option).

This now removes the ignore rule allowing invalid tags in phpdoc to avoid hiding such issues.


stof:
OK, phpstan 1.6.0 is too old to understand the `array{'@region'?: ?string,...}` syntax needed by Psalm 5.x. So I will first work on updating phpstan (after #1456 is merged) before finishing this PR.

jderusse (approved):

jderusse:
thank you :heart:

PR #1462, opened 2023-06-23
Upgrade psalm

stof:


.github/workflows/static.yml
```diff
@@ -105,7 +105,7 @@ jobs:
           composer update --no-interaction --prefer-dist --optimize-autoloader
 
       - name: Psalm
-        uses: docker://vimeo/psalm-github-actions:5.1.0
+        uses: docker://ghcr.io/psalm/psalm-github-actions:5.11.0
```
GrahamCampbell, line 108: can we install the psalm phar, similar to the way we do phpstan?

```
"psalm/phar": "5.11.0"
```

GrahamCampbell (changes requested):

jderusse (commented):
Is there a reason for not using composer like we did for phpstan and BC-break-checker?

stof:
@jderusse I kept it closer to the way it was done right now. But I'm strongly in favor of using composer

jderusse:
> @jderusse I kept it closer to the way it was done right now. But I'm strongly in favor of using composer

I'm also in favor of migrating to composer: Docker images are not always maintained, and also prevent contributor from running tools on their local environments

stof:
Well, running Psalm locally is still a pain because it requires installing dependencies with an old PHP version as target (due to the fact that Psalm defaults to analyzing the code with the lowest bound of the composer requirement for PHP but also insists that the code scanned in the vendor folder can be analyzed with this PHP version)

stof:
I have no idea why the CI fails. Running the same Psalm version locally does not report the error.

GrahamCampbell:
Possibly a bad local psalm cache. psalm is pretty bad at getting itself in a situation with an incorrect cache if things are moving around while it is running.

stof:
I'm getting this difference even with the psalm cache deleted and the `--no-cache` option

GrahamCampbell:
Hmm. 🤷

stof:
I still don't know why my local setup was not detecting this error. But as it is a legitimate one, I fixed it in the PR.

jderusse (approved):

PR #1464, opened 2023-06-23
Generate the phpdoc for the static create method of objects

stof:


src/Core/src/Sts/Input/AssumeRoleRequest.php
```diff
@@ -260,6 +260,22 @@ public function __construct(array $input = [])
...
+     *   Tags?: Tag[],
+     *   TransitiveTagKeys?: string[],
+     *   ExternalId?: string,
+     *   SerialNumber?: string,
+     *   TokenCode?: string,
+     *   SourceIdentity?: string,
+     *   '@region'?: string|null,
+     * }|AssumeRoleRequest $input
+     */
     public static function create($input): self
```
jderusse, line 279: This method is just a proxy for `--construct`.
I'm not sure duplicating this docbloc provides any value.

If the user enters invalid content, the STA should report the issue when calling `new self($input);` right?

jderusse (approved):

PR #1467, opened 2023-06-26
Generate the property type for generated objects

stof:


jderusse (approved):

stof:
I'll wait for my other PRs to be merged before fixing static analysis issues here, as they will contribute to it.

stof:
For the 3 new errors added in the psalm baseline due to having types for properties, 1 would be fixed by #1489 and the other 2 are related to #1481 (for the part I haven't fixed yet). I used to have many more reports before my other PRs related to #1481.

jderusse (approved):

Issue #1412, opened 2023-05-29
Uncaught SQS Exception when symfony/messenger worker stopped


armetiz:
Hi there.

I'm using `symfony/messenger` with SQS transport configuration; `async-aws/sqs` 1.7.0

I have two recurring uncaught exceptions from SQS Connection.
I'm not sure, but I think the problem is coming from `aws/async`, not `symfony/messenger`. 

Every ~3600 sec, symfony messenger worker is stopped because of time-limit option and is restarted automatically by supervisord.

When worker is stopping (I guess), SQS throw one of the two following uncaught exceptions.

After reading #1218, I tried to increase polling_timeout from default (0.1) to 0.5. I'll keep you informed.

Here the [issue on symfony project](https://github.com/symfony/symfony/issues/50342)

----

```
thrown in /home/ec2-user/acme/vendor/async-aws/sqs/src/Result/ReceiveMessageResult.php on line 33
#7 {main}
#6 [internal function]: Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\Connection->__destruct()
#5 /home/ec2-user/acme/vendor/symfony/amazon-sqs-messenger/Transport/Connection.php(82): Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\Connection->reset()
#4 /home/ec2-user/acme/vendor/symfony/amazon-sqs-messenger/Transport/Connection.php(371): Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\Connection->fetchMessage()
#3 /home/ec2-user/acme/vendor/symfony/amazon-sqs-messenger/Transport/Connection.php(236): AsyncAws\Sqs\Result\ReceiveMessageResult->getMessages()
#2 /home/ec2-user/acme/vendor/async-aws/sqs/src/Result/ReceiveMessageResult.php(26): AsyncAws\Core\Result->initialize()
#1 /home/ec2-user/acme/vendor/async-aws/core/src/Result.php(133): AsyncAws\Sqs\Result\ReceiveMessageResult->populateResult(Object(AsyncAws\Core\Response))
#0 /home/ec2-user/acme/vendor/async-aws/sqs/src/Result/ReceiveMessageResult.php(33): SimpleXMLElement->__construct('')
Stack trace:
PHP Fatal error:  Uncaught Exception: String could not be parsed as XML in /home/ec2-user/acme/vendor/async-aws/sqs/src/Result/ReceiveMessageResult.php:33
```

```
[...]

PR #1476, opened 2023-06-28
Fix the cancellation of prefetched pagination results

stof:


jderusse (approved):
could it be related to #1412 :thinking:

stof:
Not sure about it. I don't think a paginator is involved in those issues.

PR #1478, opened 2023-06-28
Fix the parsing of long values in headers

stof:
I missed this in #1471 as I looked for places using `long` in the code, but this type was going through the default case.


stof:
this was caught by static analysis in #1467 as it reported a mismatch between int and string.

jderusse (approved):

PR #1479, opened 2023-06-28
Add types for most of the code generator

stof:
The remaining missing types are all in the `Definition` namespace. I started the work on them but they are harder to fix (and they trigger new static analysis errors due to existing type inconsistencies) so I kept them separate.


src/CodeGenerator/src/Generator/PhpGenerator/ClassBuilder.php
```diff
@@ -57,13 +70,6 @@ public function addExtend(string $name): self
         return $this;
     }
 
-    public function setImplement($names): self
```
stof, line 60: This method is not used anyway, and it calls a non-existent method on the Nette ClassType object.

jderusse (approved):

PR #1482, opened 2023-06-30
Remove unused value object

stof:
When introducing the AppSync client in https://github.com/async-aws/aws/pull/1102, one of the operation was removed from the PR in one of the commits. But the value object used by that operation was not properly removed.

As this shape is not referenced by any managed operation, it is not regenerated when running the generator, which means that it is not in sync with the code generator output (for instance, it does not have phpdoc on its `create` method) and might not be in sync with AWS anymore (in case they added new fields for instance).

For now, I decided to delete the class, even though it is technically a BC break if someone uses the class directly. However, I don't see how such usage would be useful as there is no operation that can use this object. The only other alternative I see would be to add an operation relying on that shape so that the value object gets generated again.


stof:
@jderusse what do you think about this ?

GrahamCampbell:
Kinda wondering if our codegen should look for unused code and remove it.

stof:
If the decision is to remove it for such case, we could indeed make the code generator clean classes in the managed namespaces (`Enum`, `Exception`, `Input`, `Result` and `ValueObject`) when running in `--all` mode. But I'll wait for the decision on the removal first.

jderusse (approved):

stof:
@jderusse I will implement the cleaning in the generator. But I'll wait for the merge of https://github.com/async-aws/aws/pull/1484 to be able to reuse the DirectoryResolver instead of duplicating the logic once more.

jderusse (commented):
Could we add a BC break note in the change log?
I think I'll release a new major version.

stof:
@jderusse I added a changelog entry. The next version for this component will be a major one anyway due to the migration of long fields from string to int as AppSync was impacted.

Issue #1481, opened 2023-06-29
Strategy for dealing with invalid data


stof:
Currently, the way invalid data is handled in value objects and results is quite messy:

- getters are typed with the expected type of the value
- when creating the object, missing required fields are initialized to `null`, which will then throw a TypeError when calling the getter
- for required timestamp fields, passing a wrong date string will assign `false` to the property, which also leads to a TypeError when calling the getter
- for optional timestamp fields, passing a wrong date string will assign `null` to the property, which means it removes the date
- for integer/long fields transmitted in headers (only happening in the S3 service among the services we support currently), an invalid integer string assigns `false` to the property (which also leads to a TypeError when calling the getter)

Additionally, for JSON response parsers, we use `filter_var(..., FILTER_VALIDATE_BOOLEAN)`, which means we first cast the boolean value from the json to string before validating it again. This is quite inefficient way to ensure we have a boolean in the property.

This inconsistency between the expected property type and the assigned value gets caught by static analysis tools in https://github.com/async-aws/aws/pull/1467 once I actually define the expected property type in phpdoc.

Note: input objects are already clean in that regard. They define all fields as nullable in a consistent way (getters are also nullable) and validate things when creating the request body.

What should be the strategy to handle those values ?

stof:
Here is my proposal:

- for json parsers, use a `(bool)` cast instead of `FILTER_VALIDATE_BOOLEAN`. This still provides safety in case someone passes an unexpected value but makes it a no-op when we receive the AWS response (which is expected to return a value of the right type)
- for integer and long fields in headers, use a `(int)` cast (as done when getting an integer field in XML)
- add an helper in Core to convert date strings to DateTimeImmutable (with one static method per supported timestamp format of AWS) which would throw an exception in case of invalid value instead of returning `false` and use it in generated code (both for optional and required fields)
- for missing required fields, throw an exception instead of assigning null.

what do you think about this proposal ?

jderusse:
> when creating the object, missing required fields are initialized to null, which will then throw a TypeError when calling the getter

That's a wrong assumption. Let's take the `CreateBucketRequest::$bucket` property.
This property is `required` AND is type hinted `string|null`. The getter is type hinted `?string`. That means, users are allowed to create an instance of `CreateBucketRequest` without that property and they can call `getBucket()` without error.
[...]

PR #1483, opened 2023-06-30
Improve the parsing of values coming from headers

stof:
- integer values are casted instead of using `FILTER_VALIDATE_INT`, as done when reading from JSON or XML bodies
- required fields are read without assigning `null` as fallback in the Result object (consistent with the behavior in JSON or XML bodies)

Relates to #1481


stof:
Unfortunately, the Route53 implementation in localstack is broken: https://github.com/localstack/localstack/pull/5639#issuecomment-1064527918

stof:
OK, I removed the commit fixing the handling of required parameters coming from headers for now. I will re-submit it separately to be able to handle the case of the route53 integration tests separately.

PR #1486, opened 2023-06-30
Remove the internal tag from Core classes used in other packages

stof:
Classes and interfaces from Core used in other packages should not be marked as `@internal` as they need to be covered by BC.

For a bunch of them, I made them final (when they were not already final) so that we don't need to support the use case of inheritance.

### DIFF
--- a/src/Core/src/AwsError/AwsErrorFactoryFromResponseTrait.php
+++ b/src/Core/src/AwsError/AwsErrorFactoryFromResponseTrait.php
@@ -4,9 +4,6 @@ namespace AsyncAws\Core\AwsError;
 
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
-/**
- * @internal
- */
 trait AwsErrorFactoryFromResponseTrait
 {
     public function createFromResponse(ResponseInterface $response): AwsError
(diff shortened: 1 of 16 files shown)


src/Core/src/AwsError/AwsErrorFactoryFromResponseTrait.php
```diff
@@ -4,9 +4,6 @@
 
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
-/**
- * @internal
```
stof, line 8: I made the trait non-final because it is the only implementation of `createFromResponse` and so it might be useful for #1485 when implementing an error factory internally.

Alternatively, we could remove `createFromResponse` from the interface, inlining the extraction of the `createFromContent` arguments. A class with a custom logic in `createFromContent` is incompatible with the chain factory anyway because the chain factory performs this extraction on its own and then calls only `createFromContent`

src/Core/src/Stream/IterableStream.php
```diff
@@ -8,8 +8,6 @@
  * Convert an iterator into a Stream.
  *
  * @author Jérémy Derussé <jeremy@derusse.com>
- *
- * @internal
```
stof, line 12: Of all these stream classes, this IterableStream could stay internal if the S3 signer uses the factory instead. but other special streams can only be used directly.

PR #1488, opened 2023-07-01
Validate value object required fields in the constructor

stof:
Relates to #1481

### DIFF
--- a/src/CodeGenerator/src/Generator/ObjectGenerator.php
+++ b/src/CodeGenerator/src/Generator/ObjectGenerator.php
@@ -191,6 +191,9 @@ class ObjectGenerator
 
         $constructor->addParameter('input')->setType('array');
 
+        // To throw an exception in an expression, we need to use a method to support PHP 7.x
+        $needsThrowMethod = false;
+
         $constructorBody = '';
         foreach ($shape->getMembers() as $member) {
             $memberShape = $member->getShape();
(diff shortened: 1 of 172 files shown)


src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -191,36 +191,72 @@ private function namedConstructor(StructureShape $shape, ClassBuilder $classBuil
...
+                    $needsThrowMethod = true;
+                } else {
+                    $fallback = 'null';
+                }
+
+                $constructorBody .= strtr('$this->PROPERTY = isset($input["NAME"]) ? CLASS::create($input["NAME"]) : FALLBACK;' . "\n", [
+                    'PROPERTY' => GeneratorHelper::normalizeName($member->getName()),
+                    'NAME' => $member->getName(),
+                    'CLASS' => $objectClass->getName(),
+                    'FALLBACK' => $fallback,
```
jderusse, line 215: only when the class is used in Input? :thinking: 

Do we need to assert the content returned by the provider. Or we can blindly trust the data, and remove all this boilerplate like we do in output generated code.

stof: For now, I will keep the `isset` check for all value objects instead of changing depending on whether it is used in input. It makes the code simpler and your suggestion would not always avoid checking that for responses anyway given that some value objects are used both in input and responses.

jderusse (approved):

PR #1489, opened 2023-07-01
Fix the handling of required fields populated from headers

stof:


stof:
I'm keeping this as a draft PR for now as the integration tests of Route53 are failing right now due to an invalid implementation of the operation in `moto` (used by localstack). This has been reported at https://github.com/getmoto/moto/issues/6464

PR #1498, opened 2023-07-03
Add signature for locators in the code generator

stof:
This also provides better failure mode in case of invalid metadata instead of failing with a TypeError (static analysis tools don't like code paths that can fail with a type error).

### DIFF
--- a/src/CodeGenerator/src/Definition/Operation.php
+++ b/src/CodeGenerator/src/Definition/Operation.php
@@ -35,7 +35,7 @@ class Operation
     private $example;
 
     /**
-     * @var \Closure
+     * @var \Closure(string, Member|null=, array<string, mixed>=): Shape
      */
     private $shapeLocator;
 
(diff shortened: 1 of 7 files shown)


src/CodeGenerator/src/Definition/Operation.php
```diff
@@ -35,14 +35,17 @@ class Operation
...
+     * @var \Closure(string, Member|null=, array<string, mixed>=): Shape
      */
     private $shapeLocator;
 
     private function __construct()
     {
     }
 
+    /**
+     * @param \Closure(string, Member|null=, array<string, mixed>=): Shape $shapeLocator
```
GrahamCampbell, line 47: What does the tailing equals do?

stof: it marks the argument as optional in the callable signature (without it, you cannot omit the parameter when calling the callable).

jderusse (approved):

PR #1509, opened 2023-07-04
Fix Changelog generator

jderusse:


stof (approved):

stof:
Does this mean that a bunch of previous changelogs are also wrong ?

stof:
OK, apparently, they were fixed manually by you.

jderusse:
> OK, apparently, they were fixed manually by you.

yeah, changelogs have almost always been wrong (mostly because some entries are related to a not-supported operation).

But these last days, I've noticed that diffs were getting longer and longer :sweat_smile:

Issue #1511, opened 2023-07-05
[LocationService] Support hostPrefix for endpoints


smoench:
Sorry, I missed following with the LocationService Client generation. AWS has for each endpoint a `hostPrefix` definition:

Example:
```json
"CalculateRoute":{
      "name":"CalculateRoute",
      "http":{
        "method":"POST",
        "requestUri":"/routes/v0/calculators/{CalculatorName}/calculate/route",
        "responseCode":200
      },
      "input":{"shape":"CalculateRouteRequest"},
      "output":{"shape":"CalculateRouteResponse"},
      "errors":[
        {"shape":"InternalServerException"},
        {"shape":"ResourceNotFoundException"},
        {"shape":"AccessDeniedException"},
        {"shape":"ValidationException"},
        {"shape":"ThrottlingException"}
      ],
      "endpoint":{"hostPrefix":"routes."}
    },
```

The host needs to be constructed as followed:
`protocol://<service-code>.geo.<region-code>.amazonaws.com`

Service | Service code
-- | --
Amazon Location Maps | maps
Amazon Location Places | places
Amazon Location Routes | routes
Amazon Location Geofences | geofencing
Amazon Location Trackers | tracking

Any ideas how I can implement the functionality to respect the `hostPrefix` in the CodeGenerator?

PR #1514, opened 2023-07-07
Implement support for the host prefix in operations

stof:
This is needed to have a working implementation of LocationService.

Closes https://github.com/async-aws/aws/issues/1511

### DIFF
--- a/src/Core/src/Request.php
+++ b/src/Core/src/Request.php
@@ -44,6 +44,11 @@ final class Request
     private $endpoint;
 
     /**
+     * @var string
+     */
+    private $hostPrefix;
+
+    /**
      * @var array{scheme: string, host: string, port: int|null}|null
      */
     private $parsed;
@@ -52,7 +57,7 @@ final class Request
      * @param array<string, string> $query
      * @param array<string, string> $headers
      */
-    public function __construct(string $method, string $uri, array $query, array $headers, RequestStream $body)
+    public function __construct(string $method, string $uri, array $query, array $headers, RequestStream $body, string $hostPrefix = '')
     {
         $this->method = $method;
         $this->uri = $uri;
@@ -62,6 +67,7 @@ final class Request
         }
         $this->body = $body;
         $this->query = $query;
+        $this->hostPrefix = $hostPrefix;
         $this->endpoint = '';
     }
 
@@ -148,6 +154,17 @@ final class Request
         return $this->query;
     }
 
+    public function getHostPrefix(): string
+    {
+        return $this->hostPrefix;
+    }
+
+    public function setHostPrefix(string $hostPrefix): void
+    {
+        $this->hostPrefix = $hostPrefix;
+        $this->endpoint = '';
+    }
+
     public function getEndpoint(): string
     {
         if (empty($this->endpoint)) {
@@ -155,7 +172,7 @@ final class Request
                 throw new LogicException('Request::$endpoint must be set before using it.');
             }
 
-            $this->endpoint = $this->parsed['scheme'] . '://' . $this->parsed['host'] . (isset($this->parsed['port']) ? ':' . $this->parsed['port'] : '') . $this->uri . ($this->query ? (false === strpos($this->uri, '?') ? '?' : '&') . http_build_query($this->query, '', '&', \PHP_QUERY_RFC3986) : '');
+            $this->endpoint = $this->parsed['scheme'] . '://' . $this->hostPrefix . $this->parsed['host'] . (isset($this->parsed['port']) ? ':' . $this->parsed['port'] : '') . $this->uri . ($this->query ? (false === strpos($this->uri, '?') ? '?' : '&') . http_build_query($this->query, '', '&', \PHP_QUERY_RFC3986) : '');
         }
 
         return $this->endpoint;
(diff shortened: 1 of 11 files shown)


src/Core/src/Request.php
```diff
@@ -52,7 +57,7 @@ final class Request
      * @param array<string, string> $query
      * @param array<string, string> $headers
      */
-    public function __construct(string $method, string $uri, array $query, array $headers, RequestStream $body)
+    public function __construct(string $method, string $uri, array $query, array $headers, RequestStream $body, string $hostPrefix = '')
```
GrahamCampbell, line 60: Hmmm, this feels kinda wrong. Could we set the endpoint instead?

src/Core/src/Request.php
```diff
@@ -148,27 +154,37 @@ public function getQuery(): array
...
+            $this->endpoint = $this->parsed['scheme'] . '://' . $this->hostPrefix . $this->parsed['host'] . (isset($this->parsed['port']) ? ':' . $this->parsed['port'] : '') . $this->uri . ($this->query ? (false === strpos($this->uri, '?') ? '?' : '&') . http_build_query($this->query, '', '&', \PHP_QUERY_RFC3986) : '');
         }
 
         return $this->endpoint;
     }
 
     public function setEndpoint(string $endpoint): void
     {
-        if (!empty($this->endpoint)) {
+        if (null !== $this->parsed) {
```
GrahamCampbell, line 183: this looks wrong, because endpoint is initialized to empty string.

GrahamCampbell (changes requested):

stof:
Looking at the official SDK, there is a bunch of services that rely on this (having different subdomains for different endpoints). I found 3 kinds of prefixes:

- subdomains of the service endpoint (like for the location service, where host prefixes are `routes.` or `places.`)
- prefixes of the service endpoint changing the service subdomain (for instance, datasync has a few operations using `discovery-` as prefix
- s3 and s3control have some operations where the host prefix contains parameters coming from the input (this case is not supported by the code generator in this PR, but is properly detected to fail the code generation instead of generating invalid code)

src/Core/src/Request.php
stof: We cannot. The Request object does not know the info that is injected by the client later

src/Core/src/Request.php
stof: This is checking the `parsed` property, which is assigned in this method, not the `endpoint` one.

I removed the assignment of `$endpoint` to `$this->endpoint` in the current method, because it removed the usage of the logic inside `getEndpoint` where I injected the host prefix before the parsed host, thus ignoring the host prefix.

stof:
The CI failure for the "components lowest" job is because Composer installs the current master branch of the async-aws/core subtree split instead of the patched branch from the PR. And the subtree split does not yet have the new code targeting the 1.20.

The only solution is I see to make the required CI job green would be to split the PR in 2, with a first one implementing the new code in Core and the second one updating the code generator to use it (so that when this PR gets passed on CI, the subtree split for Core is already updated)

jderusse (approved):
LGTM.
Could you give a try @smoench ?

stof:
@jderusse what is the way to handle the CI failure ?

jderusse:
> @jderusse what is the way to handle the CI failure ?

It looks like the issue is related to the change in the queryString parameter:

```diff
-http://localhost:4570/flysystem-bucket-v1?delete
+http://localhost:4570/flysystem-bucket-v1?delete=
```

The generated code builds an URL without the `=` sign:

https://github.com/async-aws/aws/blob/78eeca6b54e72441d5fceec59d28c323b6b09b7f/src/Service/S3/src/Input/DeleteObjectsRequest.php#L214

But because of parseStr and query string stuff, the original URL is altered.

stof:
@jderusse this failure for S3 is actually an issue in the fakes3 gem used by https://github.com/async-aws/testing-s3. Instead of checking whether the query string has a parameter named `delete`, it checks whether the whole query string is equal to `delete`, without caring about the URI semantic of a query string.
AWS does not have this issue. `delete=` is accepted properly. How should we handle that ?

jderusse:
I agree that fakeS3 could/should have respected the URI semantic.
But this is not the case (and maybe not the case for "S3 like" services), and this is not an issue with aws-cli, nor official AWS SDKs.

I would suggest finding a way to keep the original query string format.

stof:
@jderusse I have an idea to preserve that.
however, note that the existing code in master already suffers from this issue when generating a presigned request, and this cannot be solved (for presigned requests, the signer needs to modify the query string, so we have no other way the rebuilding the query string from the parsed data).

stof:
@jderusse how should I handle the failure for the lowest deps job ?

stof:
@jderusse this return of the psalm issue on CI drives me mad. Would it make sense to make the CI run Psalm without keeping the cache, so that builds are isolated ? I'm less and less trusting Psalm to properly invalidate its cache when the codebase has changed.

jderusse (approved):
I've cleared the cache for psalm builds.
Hop it won't happen again

Failure for the `lowest` job would required to split this PR to merge the core first.
But I think it's OK to merge it like this.

stof:
@jderusse for the case of the lowest job, I opened https://github.com/async-aws/aws/pull/1520 to improve the setup.

PR #1515, opened 2023-07-07
Fix the dynamodb store

stof:
When recreating the table during cache clearing, the call was missing the required `AttributeDefinitions` key to define the type of the attributes of the key.
And the `KeySchemaElement` was hardcoding the attribute name as `key` instead of accounting for the fact that the controller allows to customize it.

This is an error that was actually detected by static analysis. I was working on checking the content of our baseline files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "symfony/service-contracts": "^1.0 || ^2.0 || ^3.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "3.17.*",
+        "friendsofphp/php-cs-fixer": "~3.17.0",
         "illuminate/cache": "^6.18.13 || ^7.10 || ^8.0 || ^9.0",
         "illuminate/filesystem": "^6.18.13 || ^7.10 || ^8.0",
         "illuminate/mail": "^6.18.13 || ^7.10 || ^8.0",
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -15,6 +15,5 @@ parameters:
         - src/Integration/Laravel/Filesystem/src/AsyncAwsFilesystemAdapter.php
 
     ignoreErrors:
-        - '#^Parameter \#1 \$input of method AsyncAws\\DynamoDb\\DynamoDbClient::\w+\(\) expects array\{[^\)]+\}\|AsyncAws\\DynamoDb\\Input\\\w+Input, array\{[^\)]+\} given\.#'
         - '#PHPDoc tag @throws with type Psr\\Cache\\CacheException is not subtype of Throwable$#'
         - '#^Dead catch - JsonException is never thrown in the try block\.$#'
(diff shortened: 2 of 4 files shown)


GrahamCampbell:
Ha, not sure who is actually using that flush. It also discards provisioned capacity, or if for on demand, the table ends up in a scaled down state when re-built.

GrahamCampbell:
Kinda too late now, but I'd rather we raised a not implemented exception or did a noop on flush.

stof:
Well, not sure it is too late actually. I'm 90% sure that the existing release actually throws an exception in this method due to the invalid data for the API call.

GrahamCampbell:
I guess nobody is using it then, lol.

jderusse (approved):

GrahamCampbell:
I'd rather we change this code to a noop than make it delete and re-create the table. The re-create also fails to apply the correct tags.

stof:
@jderusse what do you prefer for that ? Should I change the method to a noop ?

jderusse:
I don't understand the issue here.
The purpose of the flush method is to [remove all items from the cache](https://laravel.com/api/9.x/Illuminate/Contracts/Cache/Store.html#method_flush)
Why should we use a noop?

GrahamCampbell:
Because deleting the table is a really bad way of "removing all items", for many reasons:

1. We don't restore any tags after we're done.
2. We don't restore any stream consumers after we're done.
3. We don't remember if the table was on demand or provisioned and restore that correctly.
4. For on demand tables, it is not possible to restore with the same capacity. By default, on demand tables have a low internal capacity, and only gradually scale up over many hours or days, so if we flush an on demand table that is used a lot, the effect is the cache goes down for hours or days while it scales up.
5. The cache is unavailable while we're deleting and re-creating the table. This is not expected behaviour from a flush. Usually one expects a cache to prefer availability of reads over persistence of writes.

jderusse:
is there a way to empty a dynamodb table?

jderusse:
> With the above in mind, and the note about caches not usually providing durability, doing nothing on a flush request is not unreasonable here, and I would argue, the least worst thing we can do.

I agree that cache does not provide durability. But doing nothing when calling `flush` is quite the opposite: After flushing the cache, I expect not to get old entries.

use cases in mind: 
- flush the cache after a new release => I expect my app will not contain half data generated with the previous version.

GrahamCampbell:
> is there a way to empty a dynamodb table?

Not in a way that will be fast. It would be highly dependent on size of table and configuration.

GrahamCampbell:
By the way, the official laravel/framework implementation in the dynmoadb store looks like this:

```
    public function flush()
    {
        throw new RuntimeException('DynamoDb does not support flushing an entire table. Please create a new table.');
    }
```

jderusse:
It looks like there is no "fast way" to empty a DynamoDB... 
I'm for throwing an exception instead of a noop. If I understand well, the current implementation already fails and the user already had an exception.

jderusse (approved):

PR #1518, opened 2023-07-10
Use quoted literal in phpdoc when psalm needs them

stof:
See https://github.com/vimeo/psalm/issues/10008

Psalm treats array key shapes that match their keywords as keywords and then makes them lowercase, breaking the array shape (as array keys are case sensitive).

This updates the code generator to generate string literals for the keys for this case, to ensure compatibility with both phpstan and psalm.

### DIFF
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -53,11 +53,6 @@
       <code><![CDATA[array_merge($options, ['Bucket' => $bucket, 'Key' => $key])]]></code>
     </InvalidArgument>
   </file>
-  <file src="src/Integration/Laravel/Cache/src/AsyncAwsDynamoDbStore.php">
-    <RedundantCast>
-      <code><![CDATA[(int) $e->getCode()]]></code>
-    </RedundantCast>
-  </file>
   <file src="src/Integration/Laravel/Cache/src/ServiceProvider.php">
     <InvalidArgument>
       <code>$config</code>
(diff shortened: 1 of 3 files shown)


GrahamCampbell (approved):

jderusse (changes requested):
I'm against messing our code base because of buggy tools.

GrahamCampbell:
I don't think this is a tooling bug - this is happening because bool and null are primitive literals.

GrahamCampbell:
We should not expect `'foo'` to be treated as a literal in a type, but then not `bool`. We are also very used to `null` being treated as a literal.

GrahamCampbell:
The fact that null cannot be in an array key is a semantic rule, and we cannot expect the syntax to encode that.

jderusse:
> this is happening because bool and null are primitive literals.


the dockbloc syntax `array{x: Y}` means `literally "x"`.
If I wanted to express a type of value for the key, I would have used `array<typeOfX, Y>` (ie. `array<string, Y>` or `array<bool, Y>`.

PHP does not allow `type` as an array key. `$array[bool];` [is syntaxly incorrect in PHP.](https://3v4l.org/HYl7M) 

Therefore the docbloc `array{bool: type}` treated in something else than `$array["bool"]` is meaningless.

stof:
Well, regarding `@region`, PHPStorm was already failing the parsing without quotes around it (but in a less abrupt way). And they recognized that the issue with shape keys starting with `@` was an issue with their phpdoc parser.

However, I'm fine with waiting for the resolution on the Psalm side.

GrahamCampbell (approved):
I'd be really helpful to have this merged, even though this is not really an issue on our side, as such. It is not uncommon for projects to work around bugs in 3rd party code. ;)

stof:
As one of the benefits of async-aws over the official SDK is type safety, it would be great to have this merged to avoid Psalm users getting SA errors on valid code hinting them to make their code invalid. The maintenance cost should be minimal as the affected types are generated ones. And this only takes a small amount of code in the TypeGenerator.

I added a TODO comment referencing the Psalm issue so that we know what this code is about and we could remove it in the future if Psalm fixes its issue.

GrahamCampbell:
Any chance we can get this merged? This causes me problems most days.

jderusse (approved):
Even if I'm really upset by these buggies SAT (that asks everyone to add workarounds to make them work), I hear your pain.
Let's go ahead and move it forward. 

@stof would you mind to rebase this PR?

stof:
@jderusse rebase is done

PR #1519, opened 2023-07-10
adds support for SSO credentials

smoench:
This PR adds SSO (legacy) credentials support. The official SDK supports it and we are using this.


smoench:
@jderusse Would you accept this functionality? The official SDK supports it. In case yes I would make this PR ready to be merged.

docs/clients/sso.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Usage
+
+### Retrieve role credentials
+
+```php
+use AsyncAws\Core\Sso\Input\GetRoleCredentialsRequest;
+use AsyncAws\Core\Sso\SsoClient;
+
+$client = new StsClient();
```
stof, line 17: this is the wrong class name

jderusse:
I'm on the fence here..

From one side: You have a use case and we state that we accept all PR that add new operation if people have a use case.

On the other side: This adds a new Client inside the `core` project (downloaded by everyone), for a legacy feature.

I think we should create a "normal" (not inside core) client, and use `if class_exists(SsoClient) ...` in core

stof:
I tend to agree with @jderusse here

docs/clients/sso.md
smoench: done

smoench:
I tested it locally and it worked :-)

How should I deal with the psalm failure?

stof:
@jderusse here is the return of this dreaded Psalm caching issue...
Should we really keep caching the Psalm cache between CI runs ?

smoench:
Is there anything left I can do or support you with? I would love to get this merged and released :-)

src/Core/src/Credentials/IniFileProvider.php
```diff
@@ -92,6 +93,15 @@ private function getCredentialsFromProfile(array $profilesData, string $profile,
             return $this->getCredentialsFromRole($profilesData, $profileData, $profile, $circularCollector);
         }
 
+        $existsSsoClient = class_exists(SsoClient::class);
+        $hasProfileSsoLegacyConfig = isset($profileData[IniFileLoader::KEY_SSO_START_URL]);
+
+        if (!$existsSsoClient && $hasProfileSsoLegacyConfig) {
```
stof, line 99: I suggest doing the class_exists check only when needed:

```php
if (isset($profileData[IniFileLoader::KEY_SSO_START_URL])) {
    if (class_exists(SsoClient::class)) {
        return $this->getCredentialsFromLegacySso($profileData, $profile);
    }

    $this->logger->warning('The profile "{profile}" contains SSO (legacy) config but the "async-aws/sso" package is not installed.', ['profile' => $profile]);

    return null;
}
```

I'm also returning `null` to avoid reaching the info message saying no credentials are found. And I updated slightly the warning message.

smoench: Looks better. Thx, I adjusted it

stof (approved):

src/Integration/Symfony/Bundle/src/DependencyInjection/AwsPackagesProvider.php
```diff
@@ -154,6 +154,10 @@ public static function getAllServices(): array
                 'class' => \AsyncAws\Ssm\SsmClient::class,
                 'package' => 'async-aws/ssm',
             ],
+            'sso' => [
```
jderusse, line 157: missing entry in changelog

src/Core/src/Credentials/IniFileProvider.php
```diff
@@ -92,6 +93,16 @@ private function getCredentialsFromProfile(array $profilesData, string $profile,
             return $this->getCredentialsFromRole($profilesData, $profileData, $profile, $circularCollector);
         }
 
+        if (isset($profileData[IniFileLoader::KEY_SSO_START_URL])) {
+            if (class_exists(SsoClient::class)) {
+                return $this->getCredentialsFromLegacySso($profileData, $profile);
+            }
+
+            $this->logger->warning('The profile "{profile}" contains SSO (legacy) config but the "async-aws/sso" package is not installed.', ['profile' => $profile]);
```
jderusse, line 101: For a better DX, this deserve a `Run "composer require async-aws/sso" to add the missing dependency` (or something like that

src/Core/src/Credentials/IniFileProvider.php
```diff
@@ -146,4 +157,68 @@ private function getCredentialsFromRole(array $profilesData, array $profileData,
...
+            $profileData[IniFileLoader::KEY_SSO_ACCOUNT_ID],
+            $profileData[IniFileLoader::KEY_SSO_ROLE_NAME]
+        )) {
+            $this->logger->warning('Profile "{profile}" does not contains require legacy SSO config.', ['profile' => $profile]);
+
+            return null;
+        }
+
+        $ssoCacheFileLoader = new SsoCacheFileLoader($this->logger);
+        $tokenData = $ssoCacheFileLoader->loadSsoCacheFile($profileData[IniFileLoader::KEY_SSO_START_URL]);
```
jderusse, line 178: I think a piece of the puzzle is missing:
- This code fetch a token form the cached file
- Then use the client to get a new Role/Crendential

But what is filling the cache?

In the official AwsSdk, there is a code in charge of generating the token and storing it in the cache (see SsoTokenProvider, especially the method `ssoOidcClient->createToken`)

I feel that:
- the cache location is not a AWS standard, but only a file uses by the official PHP-SDK
- this PR is a bridge for the official AWS-SDK, and we still need the SDK to generate the token.

src/Integration/Symfony/Bundle/src/DependencyInjection/AwsPackagesProvider.php
smoench: I added the entry

src/Core/src/Credentials/IniFileProvider.php
smoench: I adjusted the log message

src/Core/src/Credentials/IniFileProvider.php
smoench: There is also a official documentation for SDKs and SSO: https://docs.aws.amazon.com/sdkref/latest/guide/access-sso.html :-)

smoench: All good with explanation and the linked documentation or are there any open question I might can answer?

src/Core/src/Credentials/SsoCacheFileLoader.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * @return array<string, string>
+     */
+    public function loadSsoCacheFile(string $ssoStartUrl): array
+    {
+        $filepath = sprintf('%s/.aws/sso/cache/%s.json', $this->getHomeDir(), sha1($ssoStartUrl));
+
+        if (!@is_readable($filepath)) {
+            $this->logger->warning('The sso cache file {path} is not readable.', ['path' => $filepath]);
```
GrahamCampbell, line 37: I don't think this is the right place for this check, and opens up a race condition between here and line 42. `file_get_contents` could still fail, but we don't properly handle that.

GrahamCampbell (changes requested):

src/Core/src/Credentials/SsoCacheFileLoader.php
GrahamCampbell: Example robust ways to open the file, then read it's contents can be found in the Guzzle Utils class (`tryFopen`, `tryGetContents`): https://github.com/guzzle/psr7/blob/2.5.0/src/Utils.php#L340-L434.

smoench: What do you think about this? Would it be enough?
```php
if (!@is_readable($filepath) || false === ($content = file_get_contents($filepath))) {
```

GrahamCampbell: `file_get_contents` can also raise errors. The linked code handles them and surfaces they back to the user as an exception.

GrahamCampbell: Checking is_readable is largely not useful, because the file's readability could change in between that call and then the file_get_contents call.

smoench: FYI the `IniFileLoader` has the same race condition:
- https://github.com/async-aws/aws/blob/master/src/Core/src/Credentials/IniFileLoader.php#L54
- https://github.com/async-aws/aws/blob/master/src/Core/src/Credentials/IniFileLoader.php#L95

GrahamCampbell: Great - this means we are justified to add a little internal class in the Core namespace to get the contents of a file or throw an exception.

smoench: @GrahamCampbell Since the [referenced code](https://github.com/guzzle/psr7/blob/2.5.0/src/Utils.php#L340-L434) is written by yourself, would you consider to contribute it here?

smoench: Since I didn't get an answer, I will go with following:
```php
if (false === ($contents = @file_get_contents($filepath))) {
```
I don't fell well to copy the referenced code when the original author is involved here 😉

src/Core/src/Credentials/IniFileProvider.php
```diff
@@ -146,4 +157,68 @@ private function getCredentialsFromRole(array $profilesData, array $profileData,
...
+     */
+    private function getCredentialsFromLegacySso(array $profileData, string $profile): ?Credentials
+    {
+        if (!isset(
+            $profileData[IniFileLoader::KEY_SSO_START_URL],
+            $profileData[IniFileLoader::KEY_SSO_REGION],
+            $profileData[IniFileLoader::KEY_SSO_ACCOUNT_ID],
+            $profileData[IniFileLoader::KEY_SSO_ROLE_NAME]
+        )) {
+            $this->logger->warning('Profile "{profile}" does not contains require legacy SSO config.', ['profile' => $profile]);
```
stof, line 172:
```suggestion
            $this->logger->warning('Profile "{profile}" does not contains required legacy SSO config.', ['profile' => $profile]);
```

stof (changes requested):

src/Core/src/Credentials/SsoCacheFileLoader.php
jderusse: are you OK with this change @GrahamCampbell

GrahamCampbell: Yeh, we could also just "fix it in post", since this is an existing bug, and nobody seems to mind it so far. Are you OK if we make an issue for it, assign me to it, then I'll get to it when I have some time.

jderusse (approved):

PR #1520, opened 2023-07-10
Allow using the patched dev-master branch of core in lowest jobs

stof:
Thanks to marking the path repository non-canonical for the lowest job, it will only be used when we don't have a stable release matching the constraint. But it will be used when the requirement enforces the usage of the dev version instead of using the version from the existing subtree split which does not have the in-progress patch of the dev version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,6 @@ jobs:
           echo ::endgroup::
 
       - name: Modify composer.json - default
-        if: matrix.strategy != 'lowest'
         run: |
           CURRENT_DIR=$(pwd)
           for COMPONENT in $(find src/Service -maxdepth 2 -type f -name phpunit.xml.dist -printf '%h\n' | sort)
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
dev-master is kinda a hack. why can't we just jam on `@dev` or set minimum stability to dev with perfer stable set to true?

GrahamCampbell:
The other PR should be using .20 instead of .19 for the core version, no?

stof:
@GrahamCampbell I'm using `dev-master` on purpose because we need the branch alias (I could do much more magic by making it guess as a different `dev-*` version and adding a branch alias for it) and because I want to shadow the `dev-master` version of https://github.com/async-aws/core which won't have the PR changes yet.
We already set the minimum stability to dev. But `dev-c11d28cddaf0b04245efc37764ec92df3ffa67f1` does not satisfy a `^1.9` requirement.
Previously, the code was replacement the `async-aws/core` requirement with `*@dev` to make any version accepted. This works fine for the normal job (especially given that we have a single candidate version thanks to the canonical repository). But that would totally break the intent of the `prefer-lowest` job which is meant to check the lowest bound.

> The other PR should be using .20 instead of .19 for the core version, no?

It already required `1.20` as minimum version:
https://github.com/async-aws/aws/pull/1514/files#diff-b564474f040ef86ed0644ef2db00b2c6d78b712e5cab1748eda17798bace08e9R17

The issue is that the 1.20.x-dev version exists in 2 places at that point:
- the master branch in https://github.com/async-aws/core which is known by packagist but does not have the PR changes yet.
- the patched version in the PR which updates this 1.20.x-dev in-progress version with more changes

We want our CI to simulate what would happen once the PR is merged and we run tests on the new master branch.

jderusse (approved):

PR #1525, opened 2023-07-23
CS and doc fixes

GrahamCampbell:
1. Upgrade to `php-cs-fixer` 3.22.
2. Remove redundant `.php-cs-fixer` config and sort fixers by name.
3. Update phpdoc generator to remove double space characters.


src/CodeGenerator/src/Definition/Shape.php
```diff
@@ -35,7 +35,7 @@ private function __construct()
 
     /**
      * @param \Closure(string, Member|null=, array<string, mixed>=): Shape $shapeLocator
-     * @param \Closure(): ServiceDefinition $serviceLocator
+     * @param \Closure(): ServiceDefinition                                $serviceLocator
```
jderusse, line 38: Why this change

GrahamCampbell: php-cs-fixer made it - it is a bugfix with the alignment when there's a complex type

jderusse (approved):

PR #1541, opened 2023-08-08
Build Doc only when necessary

jderusse:


.github/workflows/website.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+    paths:
+      - 'docs/**'
+      - 'website/**'
+      - 'couscous.yml'
```
stof, line 10: this should also include `.github/workflows/website.yml` so that changing the workflow file also triggers a run on next push to master.

stof (changes requested):

.github/workflows/website.yml
jderusse: added

stof (approved):

PR #1542, opened 2023-08-08
Allow passing explicit null values for optional input members

stof:
This makes it easier to define the array literal for input objects as PHP does not have a syntax sugar for conditional keys in an array literal. This is consistent with the generated code for value objects.

Even though input objects allow to omit required members in the constructor shape (as they can be set later by using the setter), the phpdoc type still does not allow passing null explicitly (even though the code would deal with it until the validation run) so that static analysis tools can catch mistakes there. Passing a required member explicitly is intended to pass a valid value for it and not a potentially missing one.


jderusse (approved):
Could you add a `### Changed` entry in CHANGELOGs ?

I'll soon create a script that uses the changelog to "propose" a new version and automate the release process (It took me ~2 hours to release all the components yesterday)

stof:
@jderusse Done

jderusse:
> @jderusse Done

did you pushed it ?

stof:
arf. I forgot to push. I'll open a new PR.

PR #1565, opened 2023-09-04
Fix properties not base64 decoded

### LINKED ISSUE #1564: DynamoDB binary value encoding inconsistency
In requests, one passes the raw binary string `['B' => 'foo bar baz']`, but in responses, `->getB()` is base64 encoded. This feels like a bug to me (though will need a major version bump to "fix", so as to not break stuff for people).

jderusse:
fix #1564

### DIFF
--- a/src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
+++ b/src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
@@ -355,14 +355,14 @@ class RestJsonParser implements Parser
                     $body = '
                         $items = [];
                         foreach ($json as $name => $value) {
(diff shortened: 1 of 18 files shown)


GrahamCampbell:
Oooh, it looks like this possibly fixes another issue where by nesting of attributes such as in maps are not transformed?

GrahamCampbell:
Oh, yeh, you mention this in the other issue. ;)

GrahamCampbell:
I had forgotten that I'd also seen this happening with maps before in dynamodb, which was quite annoying, in particular for request retrying using the output of a partial failure from batch write items.

hassaan-dev:
Thank you guys

PR #1566, opened 2023-09-06
Update async-aws/simple-s3 to use async-aws/s3 2.0

stof:


jderusse (approved):
good catch

stof:
@jderusse any chance to tag the 2.0 release of the simple-s3 package to unlock the upgrade of the s3 package ?

Issue #1575, opened 2023-09-27
S3 Retain Date Incompatibility DateTime ISO8601


ngourle:
The use of DateTimeInterface::ISO8601 is incompatible with the ISO 8601 date format according to the PHP documentation. As a result, I was unable to apply a retention delay when uploading an object.

PR #1576, opened 2023-09-27
DateTimeInterface::ISO8601 Incompatibility with ISO8601

ngourle:
Related #1575

### DIFF
--- a/src/CodeGenerator/src/Generator/InputGenerator.php
+++ b/src/CodeGenerator/src/Generator/InputGenerator.php
@@ -562,6 +562,10 @@ PHP
         switch ($shape->getType()) {
             case 'timestamp':
                 $format = strtoupper($shape->get('timestampFormat') ?? ('header' === $part ? 'rfc822' : 'atom'));
+                if ('ISO8601' === $format) {
+                    $format = 'ATOM';
+                }
+
                 if (!\defined('\DateTimeInterface::' . $format)) {
                     throw new \InvalidArgumentException('Constant "\DateTimeInterface::' . $format . '" does not exists.');
                 }
--- a/src/CodeGenerator/src/Generator/InputGenerator.php
+++ b/src/CodeGenerator/src/Generator/InputGenerator.php
@@ -561,7 +561,7 @@ PHP
         $shape = $member->getShape();
         switch ($shape->getType()) {
             case 'timestamp':
-                $format = strtoupper($shape->get('timestampFormat') ?? ('header' === $part ? 'rfc822' : 'atom'));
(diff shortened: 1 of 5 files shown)


jderusse (commented):
Tests are red because it need to be fixed in `src/CodeGenerator/src/Generator/InputGenerator.php`

src/CodeGenerator/src/Generator/InputGenerator.php
```diff
@@ -561,7 +561,7 @@ private function stringify(string $variable, Member $member, string $part): stri
         $shape = $member->getShape();
         switch ($shape->getType()) {
             case 'timestamp':
-                $format = strtoupper($shape->get('timestampFormat') ?? ('header' === $part ? 'rfc822' : 'iso8601'));
+                $format = strtoupper($shape->get('timestampFormat') ?? ('header' === $part ? 'rfc822' : 'atom'));
```
jderusse, line 564: `shape->get('timestampFormat')` might return `iso8601`.

```suggestion
                $format = strtoupper($shape->get('timestampFormat') ?? ('header' === $part ? 'rfc822' : 'atom'));
                if ($format === 'ISO8601') {
                    $format = 'ATOM'
                }
```

ngourle: You are right, thank you

jderusse (commented):
Last request. Could you please add an entry in the `src/Service/S3/CHANGELOG.md` file?

src/CodeGenerator/src/Generator/InputGenerator.php
```diff
@@ -561,7 +561,11 @@ private function stringify(string $variable, Member $member, string $part): stri
...
```
stof, line 564: the fallback here should still be `iso8601`, to correspond to values used in `timestampFormat` IMO. The rewriting will be done below anyway

src/Service/S3/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Fixed
+
+- According PHP documentation, format datetime with `ATOM` to provide a workaround for the incompatibility with `ISO8601` format.
```
stof, line 7: the fix should be described what is getting fixed in the library, not _how_  (as users of the library don't care about that).

So this should be ```Fix the format of the `objectLockRetainUntilDate` field in requests``` IMO

src/CodeGenerator/src/Generator/InputGenerator.php
ngourle: That's a good point, I'll change that quickly.

jderusse:
thank you @ngourle

PR #1582, opened 2023-10-04
Add test to assert CHANGELOG format

jderusse:
This PR adds 2 tests:
- Assert that a CHANGELOG entry is present
- Assert the CHANGELOG files content are consistent

all files MUST follow the following format (inspired by https://keepachangelog.com/en/1.1.0/)
```
# Change Log

## NOT RELEASED

## 1.2.0

### Added|Changed|Fixed|Deprecated|Removed|Security

- single line entry
- multiline
  entry

## 1.1.0

- Empty release

## 0.1

First version

```

It also fixes all previous change log files to make the baseline.

### DIFF
--- a/tests/Unit/ChangelogTest.php
+++ b/tests/Unit/ChangelogTest.php
@@ -1,0 +1,276 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Test\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Finder\Finder;
+
+/**
+ * Verify CHANGELOG entries are present.
+ */
+class ChangelogTest extends TestCase
+{
+    /**
+     * @dataProvider provideChangedlogFiles
+     */
+    public function testChangelogFormat(string $changelogPath)
+    {
+        $lines = explode("\n", file_get_contents($changelogPath));
+
+        self::assertGreaterThan(3, \count($lines), 'CHANGELOG must not be empty');
+        self::assertSame('# Change Log', $lines[0], 'CHANGELOG MUST start with "# Change Log"');
+        self::assertSame('', $lines[1], 'CHANGELOG MUST have a second line empty');
+        self::assertSame('## NOT RELEASED', $lines[2], 'CHANGELOG MUST have a "## NOT RELEASED" section');
+
+        $inTitle1 = false;
+
+        $inTitle2 = false;
+        $title2Empty = false;
+
+        $inTitle3 = false;
+        $title3Empty = false;
+
+        $title3Choices = [
+            '### BC-BREAK', '### Removed', // Major
+            '### Added', '### Deprecated', // Minor
+            '### Changed', '### Fixed', '### Security', // Patch
+        ];
+        $lastTitle3 = -1;
+        $needEmptyLine = false;
+        $wasEmptyLine = false;
+        $finalWords = false;
+        foreach ($lines as $index => $line) {
+            if ($needEmptyLine) {
+                self::assertSame('', $line, 'CHANGELOG line ' . $index . ' MUST be empty');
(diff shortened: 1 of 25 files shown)


src/Integration/Laravel/Queue/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 ## 0.2.1
 
-### Changed
+### Fixed
```
stof, line 7: Why not allowing `Changed` anymore ? Not all changes are bug fixes.

tests/Unit/ChangelogTest.php
```diff
@@ -0,0 +1,268 @@
...
+class ChangelogTest extends TestCase
+{
+    /**
+     * @dataProvider provideChangedlogFiles
+     */
+    public function testChangelogFormat(string $changelogPath)
+    {
+        $lines = explode("\n", file_get_contents($changelogPath));
+
+        self::assertGreaterThan(3, \count($lines), 'CHANGEOG must not be empty');
```
stof, line 22:
```suggestion
        self::assertGreaterThan(3, \count($lines), 'CHANGELOG must not be empty');
```

tests/Unit/ChangelogTest.php
```diff
@@ -0,0 +1,268 @@
...
+                $this->expectNotToPerformAssertions();
+
+                return;
+            }
+        }
+
+        self::fail('Missing CHANGELOG entry for package ' . $service);
+    }
+
+    public function provideChangedlogFiles(): iterable
```
stof, line 162: Data providers should be static methods. Non-static ones are deprecated in latest PHPUnit.

tests/Unit/ChangelogTest.php
```diff
@@ -0,0 +1,269 @@
...
+
+            if (0 === strpos($line, '## ')) {
+                self::assertTrue($wasEmptyLine, 'CHANGELOG line ' . ($index - 1) . ' MUST be empty');
+                self::assertTrue($inTitle1, 'CHANGELOG line ' . $index . ' MUST have title2 inside a title1');
+                self::assertFalse($title2Empty, 'CHANGELOG line ' . $index . ' MUST not have empty title2');
+                self::assertFalse($title3Empty, 'CHANGELOG line ' . $index . ' MUST not have empty title3');
+                self::assertMatchesRegularExpression('/^## (NOT RELEASED|\d+\.\d+\.\d+)$/', $line, 'CHANGELOG line ' . $index . ' MUST be "## X.Y.Z"');
+                $inTitle2 = true;
+                $inTitle3 = false;
+                $title3s = [];
```
stof, line 71: Before resetting the list of title 3, should we assert the order of title 3 of the previous section to ensure consistent order.

jderusse: Yeah, I hesitated about this. But fear the diff to fix the existing changelogs...


let's do it once for all

tests/Unit/ChangelogTest.php
jderusse: thanks

src/Integration/Laravel/Queue/CHANGELOG.md
jderusse: I would like to have as few "words" as possible

- `BC-Break` and `Added` works well for Major and Minor.
- 'Fixed` works well to.
- for Patches, do you think `Changes` is not ambiguous? (like "I changed the method signatures")

Should we allow `Chores`, `Perf`, `Refactor`, ... like in https://www.conventionalcommits.org/en/v1.0.0/ ?

jderusse: Let's go with https://keepachangelog.com/en/1.1.0/

Issue #1589, opened 2023-10-17
Support UploadPartCopy in async-aws/s3


nmuntyanov:
Hello guys!
I've not found how i can use UploadPartCopy (https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html)
functionality in async-aws/s3
Will be nice to have it to be able to copy files more than 5GB

jderusse:
Hi, Indeed, AsyncAws does not YET support this operation, but you can it by following this documentation https://async-aws.com/contribute/generate.html#creating-a-new-client-operation we'll be happy to review and approve your PR :hugs:

nmuntyanov:
@jderusse may you take a look, please?
https://github.com/async-aws/aws/pull/1590

and why this test is failed?
https://github.com/async-aws/aws/actions/runs/6547020143/job/17778665168?pr=1590

PR #1590, opened 2023-10-17
Multipart copy method

nmuntyanov:
fix https://github.com/async-aws/aws/issues/1589

### DIFF
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -102,7 +102,7 @@ class SimpleS3Client extends S3Client
         for ($i = 1; $bytePosition < $contentLength; ++$i) {
             $startByte = $bytePosition;
             $endByte = $bytePosition + $partSize - 1 >= $contentLength ? $contentLength - 1 : $bytePosition + $partSize - 1;
-            $parts[] = $this->doMultipartCopy($destBucket, $destKey, $uploadId, $i, "{$srcBucket}/{$srcKey}", $startByte, $endByte);
+            $parts[] = $this->doMultipartCopy($destBucket, $destKey, $uploadId, $i, sprintf('%s/%s', $srcBucket, $srcKey), $startByte, $endByte);
             $bytePosition += $partSize;
         }
         $this->completeMultipartUpload(
@@ -273,7 +273,7 @@ class SimpleS3Client extends S3Client
                     'Key' => $key,
                     'UploadId' => $uploadId,
                     'CopySource' => $copySource,
-                    'CopySourceRange' => "bytes={$startByte}-{$endByte}",
+                    'CopySourceRange' => sprintf('bytes=%d-%d', $startByte, $endByte),
                     'PartNumber' => $partNumber,
                 ])
             );
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -279,7 +279,7 @@ class SimpleS3Client extends S3Client
                     'PartNumber' => $partNumber,
                 ])
             );
-            /** @var  CopyPartResult $copyPartResult */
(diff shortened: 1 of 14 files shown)


src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +52,71 @@ public function has(string $bucket, string $key): bool
...
+                array_merge($options, ['Bucket' => $destBucket, 'Key' => $destKey])
+            )
+        )->getUploadId();
+
+        $bytePosition = 0;
+        $parts = [];
+        for ($i = 1; $bytePosition < $contentLength; ++$i) {
+            $startByte = $bytePosition;
+            $endByte = $bytePosition + $partSize - 1 >= $contentLength ? $contentLength - 1 : $bytePosition + $partSize - 1;
+            $parts[] = $this->doMultipartCopy($destBucket, $destKey, $uploadId, $i, sprintf('%s/%s', $srcBucket, $srcKey), $startByte, $endByte);
```
jderusse, line 107: would it make sens to run this in parallel?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -195,4 +265,28 @@ private function doSmallFileUpload(array $options, string $bucket, string $key,
             'Body' => $object,
         ]));
     }
+
+    private function doMultipartCopy(string $bucket, string $key, string $uploadId, int $partNumber, string $copySource, int $startByte, int $endByte): CompletedPart
+    {
+        try {
+            $response = $this->uploadPartCopy(
```
jderusse, line 272: this operation requires the (not yet release) next version of S3.
You need to either upgrade composer.json `"async-aws/s3": "^2.1"`

jderusse (commented):
You should split this PR in 2:
- adding the `uploadPartCopy` operation in S3
- adding a new method in SimpleS3 that use the new method of S3

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
nmuntyanov: makes sense
but i have no idea how to implement this))

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
nmuntyanov: https://github.com/async-aws/aws/pull/1591
First one

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
jderusse: Avoid consuming the response: When you access a property (calling a getter) the response blocks until the response is fully processed.

once again, I don't know if it's better to run theses requests in sequence or parallel.
(Maybe run too many requests in parallel have worse performance) => you need to check AWS recommendations for this.


to process requests in parallel you should do something like
```
for (...) {
  $responses[] = $client->uploadPartCopy(...)
}

$success = true;
  try {
    $copyPartResult = $response->getCopyPartResult();
    $parts[] = new CompletedPart(['ETag' => $copyPartResult->getEtag(), 'PartNumber' => $partNumber]);
  } catch (\Throwable $e) {
    $success = false;
    break'
  }
}

if (!$success) {
  $this->abortMultipartUpload(['Bucket' => $bucket, 'Key' => $key, 'UploadId' => $uploadId]);
    try {
      $response->cancel();
    } catch (\Throwable $e) {
      // ...
    }
  }

   throw ...;
}
```

nmuntyanov: Got it!
Seems like the only limit is 10 000 connections (equal to parts)
https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
Anyway, think that 10k connections, even without body, it is too much.

AWS SDK has concurrency option for such copy
Will see how to implement it here

PR #1591, opened 2023-10-17
s3Client::uploadPartCopy method added

nmuntyanov:
Fix https://github.com/async-aws/aws/issues/1589

s3Client::uploadPartCopy method added to be able to copy large files between buckets


nmuntyanov:
Parent PR 
https://github.com/async-aws/aws/pull/1590

jderusse (approved):

PR #1592, opened 2023-10-17
Issue 1589/s3 simple client

nmuntyanov:
SimpleS3Client::copy method added

Method resolve which api it should use by itself (copyObject as atomic or copy by parts)

### DIFF
--- a/docs/integration/simple-s3.md
+++ b/docs/integration/simple-s3.md
@@ -32,6 +32,9 @@ $resource = \fopen('/path/to/cat/image.jpg', 'r');
 $s3->upload('my-image-bucket', 'photos/cat_2.jpg', $resource);
 $s3->upload('my-image-bucket', 'photos/cat_2.txt', 'I like this cat');
 
+// Copy objects between buckets
+$s3->copy('source-bucket', 'source-key', 'destination-bucket', 'destination-key');
+
 // Check if a file exists
 $s3->has('my-image-bucket', 'photos/cat_2.jpg'); // true
 
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -7,10 +7,15 @@ namespace AsyncAws\SimpleS3;
 use AsyncAws\Core\Stream\FixedSizeStream;
 use AsyncAws\Core\Stream\ResultStream;
 use AsyncAws\Core\Stream\StreamFactory;
+use AsyncAws\S3\Input\CompleteMultipartUploadRequest;
+use AsyncAws\S3\Input\CopyObjectRequest;
+use AsyncAws\S3\Input\CreateMultipartUploadRequest;
 use AsyncAws\S3\Input\GetObjectRequest;
+use AsyncAws\S3\Input\UploadPartCopyRequest;
 use AsyncAws\S3\S3Client;
 use AsyncAws\S3\ValueObject\CompletedMultipartUpload;
 use AsyncAws\S3\ValueObject\CompletedPart;
+use AsyncAws\S3\ValueObject\CopyPartResult;
 
 /**
  * A simplified S3 client that hides some of the complexity of working with S3.
@@ -45,6 +50,73 @@ class SimpleS3Client extends S3Client
     public function has(string $bucket, string $key): bool
     {
         return $this->objectExists(['Bucket' => $bucket, 'Key' => $key])->isSuccess();
+    }
+
+    /**
+     * @param array{
+     *   ACL?: \AsyncAws\S3\Enum\ObjectCannedACL::*,
+     *   CacheControl?: string,
+     *   ContentLength?: int,
+     *   ContentType?: string,
+     *   Metadata?: array<string, string>,
+     *   PartSize?: int,
+     * } $options
+     */
+    public function copy(string $srcBucket, string $srcKey, string $destBucket, string $destKey, array $options = []): void
+    {
+        $megabyte = 1024 * 1024;
+        if (!empty($options['ContentLength'])) {
+            $contentLength = (int) $options['ContentLength'];
+            unset($options['ContentLength']);
+        } else {
+            $contentLength = (int) $this->headObject(['Bucket' => $srcBucket, 'Key' => $srcKey])->getContentLength();
+        }
+
+        /*
+         * The maximum number of parts is 10.000. The partSize must be a power of 2.
+         * We default this to 64MB per part. That means that we only support to copy
+         * files smaller than 64 * 10 000 = 640GB. If you are coping larger files,
+         * please set PartSize to a higher number, like 128, 256 or 512. (Max 4096).
+         */
+        $partSize = ($options['PartSize'] ?? 64) * $megabyte;
+        unset($options['PartSize']);
+
+        // If file is less than 5GB, use normal atomic copy
+        if ($contentLength < 5120 * $megabyte) {
+            $this->copyObject(
+                CopyObjectRequest::create(
+                    array_merge($options, ['Bucket' => $destBucket, 'Key' => $destKey, 'CopySource' => "{$srcBucket}/{$srcKey}"])
+                )
+            );
+
+            return;
+        }
+
+        /** @var string $uploadId */
+        $uploadId = $this->createMultipartUpload(
+            CreateMultipartUploadRequest::create(
+                array_merge($options, ['Bucket' => $destBucket, 'Key' => $destKey])
+            )
+        )->getUploadId();
+
+        $partIndex = 1;
+        $startByte = 0;
+        while ($startByte < $contentLength) {
+            $endByte = min($startByte + $partSize, $contentLength) - 1;
+
+            $parts[] = $this->doMultipartCopy($destBucket, $destKey, $uploadId, $partIndex, "{$srcBucket}/{$srcKey}", $startByte, $endByte);
+
+            $startByte += $partSize;
+            $partIndex ++;
+        }
+        $this->completeMultipartUpload(
+            CompleteMultipartUploadRequest::create([
+                'Bucket' => $destBucket,
+                'Key' => $destKey,
+                'UploadId' => $uploadId,
+                'MultipartUpload' => new CompletedMultipartUpload(['Parts' => $parts]),
+            ])
+        );
     }
 
     /**
@@ -194,5 +266,29 @@ class SimpleS3Client extends S3Client
             'Key' => $key,
             'Body' => $object,
         ]));
+    }
+
+    private function doMultipartCopy(string $bucket, string $key, string $uploadId, int $partNumber, string $copySource, int $startByte, int $endByte): CompletedPart
+    {
+        try {
+            $response = $this->uploadPartCopy(
+                UploadPartCopyRequest::create([
+                    'Bucket' => $bucket,
+                    'Key' => $key,
+                    'UploadId' => $uploadId,
+                    'CopySource' => $copySource,
+                    'CopySourceRange' => sprintf('bytes=%d-%d', $startByte, $endByte),
+                    'PartNumber' => $partNumber,
+                ])
+            );
+            /** @var CopyPartResult $copyPartResult */
+            $copyPartResult = $response->getCopyPartResult();
+
+            return new CompletedPart(['ETag' => $copyPartResult->getEtag(), 'PartNumber' => $partNumber]);
+        } catch (\Throwable $e) {
+            $this->abortMultipartUpload(['Bucket' => $bucket, 'Key' => $key, 'UploadId' => $uploadId]);
+
+            throw $e;
+        }
     }
 }
--- a/manifest.json
+++ b/manifest.json
@@ -577,7 +577,8 @@
                 "PutObject",
                 "PutObjectAcl",
                 "PutObjectTagging",
-                "UploadPart"
+                "UploadPart",
+                "UploadPartCopy"
             ]
         },
         "Scheduler": {
--- a/src/Integration/Aws/SimpleS3/CHANGELOG.md
+++ b/src/Integration/Aws/SimpleS3/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 - Upgrade to `async-aws/s3` 2.0
 
+### Added
+
+- Added `SimpleS3Client::copy()` method
+
 ## 1.1.1
 
 ### Changed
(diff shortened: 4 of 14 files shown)


src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +53,119 @@ public function has(string $bucket, string $key): bool
...
+            $this->copyObject(
+                CopyObjectRequest::create(
+                    array_merge($options, ['Bucket' => $destBucket, 'Key' => $destKey, 'CopySource' => "{$srcBucket}/{$srcKey}"])
+                )
+            );
+
+            return;
+        }
+
+        /** @var string $uploadId */
```
stof, line 103: this is useless. `getUploadId` already define its proper return type.

stof (changes requested):

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
nmuntyanov: Hello.
How can I handle this errors 
https://github.com/async-aws/aws/actions/runs/6642786472/job/18048337300?pr=1592
without doc blocs?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +53,119 @@ public function has(string $bucket, string $key): bool
...
+        $partNumber = 1;
+        $startByte = 0;
+        $parts = [];
+        while ($startByte < $contentLength) {
+            $parallelChunks = $concurrency;
+            $responses = [];
+            while ($startByte < $contentLength && $parallelChunks > 0) {
+                $endByte = min($startByte + $partSize, $contentLength) - 1;
+                $responses[$partNumber] = $this->uploadPartCopy(
+                    UploadPartCopyRequest::create([
```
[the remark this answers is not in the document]
nmuntyanov: Where should I change this version?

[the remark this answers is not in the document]
stof: in the composer.json of this package

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
stof: Well, if the AWS API can indeed return `string|null` for the upload id, you must actually handle the case of `null` properly instead of pretending that the type inference is wrong and forcing the type checkers to consider the variable as having the type `string`.
If the AWS API actually returns `null`, the code would still be broken at runtime and the type override added there only prevents the type checker for detecting the broken code. It does not make it less broken.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +53,117 @@ public function has(string $bucket, string $key): bool
...
+
+                $startByte += $partSize;
+                ++$partNumber;
+                --$parallelChunks;
+            }
+            $error = null;
+            foreach ($responses as $idx => $response) {
+                try {
+                    $copyPartResult = $response->getCopyPartResult();
+                    $parts[] = new CompletedPart(['ETag' => $copyPartResult->getEtag(), 'PartNumber' => $idx]);
```
stof, line 136: The weird thing about the PossiblyNullReference error reported by Psalm here is that https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html#API_UploadPartCopy_ResponseElements documents CopyPartResult as required in the output. So it is weird that the operation was generated with a nullable type. Maybe the AWS SDK has metadata that does not match the documentation there.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
nmuntyanov: As I understand from AWS DOC (https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html)
`If the action is successful, the service sends back an HTTP 200 response.`
Upload Id will be always string, otherwise exception would be thrown

underlying client handle case with uploadId == null and throws an exception.
I don't see why this code should be broken, exception will be thown any way

the solution was taken from here
https://github.com/async-aws/aws/blob/master/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php#L104

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +53,119 @@ public function has(string $bucket, string $key): bool
...
+                );
+
+                $startByte += $partSize;
+                ++$partNumber;
+                --$parallelChunks;
+            }
+            $error = null;
+            foreach ($responses as $idx => $response) {
+                try {
+                    /** @var CopyPartResult $copyPartResult */
```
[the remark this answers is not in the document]
nmuntyanov: here is the same situation, 
exception is handled in the same place

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
nmuntyanov: It seems there are no "required" fields for outputs

stof: I suggest reporting this to AWS as it looks like a mismatch between the human-readable documentation and the machine-readable data used to generate the SDKs.

nmuntyanov: https://repost.aws/questions/QU6D3R02ZGSgG5C2-yPYw-qw/aws-php-sdk-api-2-json-is-not-synchronised-with-docs-aws-amazon-com

nmuntyanov: what should I do?
Add additional checking? Or will wait for AWS fix?
We waiting for this functionality

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
nmuntyanov: @stof any updates on this?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
nmuntyanov: @stof any updates on this?

nmuntyanov:
ping

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +55,114 @@ public function has(string $bucket, string $key): bool
         return $this->objectExists(['Bucket' => $bucket, 'Key' => $key])->isSuccess();
     }
 
+    /**
+     * @param array{
+     *   ACL?: \AsyncAws\S3\Enum\ObjectCannedACL::*,
+     *   CacheControl?: string,
+     *   Metadata?: array<string, string>,
+     *   PartSize?: positive-int,
+     *   Concurrency?: positive-int,
```
jderusse, line 64: please use `int`, this syntax is not a standard, therefore not understood by all IDE.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +55,114 @@ public function has(string $bucket, string $key): bool
...
+     *   mupThreshold?: positive-int,
+     * } $options
+     */
+    public function copy(string $srcBucket, string $srcKey, string $destBucket, string $destKey, array $options = []): void
+    {
+        $megabyte = 1024 * 1024;
+        $sourceHead = $this->headObject(['Bucket' => $srcBucket, 'Key' => $srcKey]);
+        $contentLength = (int) $sourceHead->getContentLength();
+        $options['ContentType'] = $sourceHead->getContentType();
+        $concurrency = (int) ($options['Concurrency'] ?? 10);
```
jderusse, line 74: no need to cast, as it's an int by contract. Let people deal with PHP runtime errors if they misuse the method.

docs/integration/simple-s3.md
```diff
@@ -32,6 +32,9 @@ $resource = \fopen('/path/to/cat/image.jpg', 'r');
 $s3->upload('my-image-bucket', 'photos/cat_2.jpg', $resource);
 $s3->upload('my-image-bucket', 'photos/cat_2.txt', 'I like this cat');
 
+// Copy objects between buckets
+$s3->copy('source-bucket', 'source-key', 'destination-bucket', 'destination-key');
```
jderusse, line 36: Options parameters deserve a small documentation IMHO

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +55,114 @@ public function has(string $bucket, string $key): bool
...
     }
 
+    /**
+     * @param array{
+     *   ACL?: \AsyncAws\S3\Enum\ObjectCannedACL::*,
+     *   CacheControl?: string,
+     *   Metadata?: array<string, string>,
+     *   PartSize?: positive-int,
+     *   Concurrency?: positive-int,
+     *   mupThreshold?: positive-int,
```
jderusse, line 65: For consistency

```suggestion
     *   MupThreshold?: positive-int,
```

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +55,114 @@ public function has(string $bucket, string $key): bool
...
+         * The maximum number of parts is 10.000. The partSize must be a power of 2.
+         * We default this to 64MB per part. That means that we only support to copy
+         * files smaller than 64 * 10 000 = 640GB. If you are coping larger files,
+         * please set PartSize to a higher number, like 128, 256 or 512. (Max 4096).
+         */
+        $partSize = ($options['PartSize'] ?? 64) * $megabyte;
+        unset($options['PartSize']);
+
+        // If file is less than multipart upload threshold, use normal atomic copy
+        if ($contentLength < $mupThreshold) {
```
jderusse, line 87: do we really need a parameter for this? 
Is it an issue to split the file even if it is smaller than the default 2GB?

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +55,114 @@ public function has(string $bucket, string $key): bool
...
+        $concurrency = (int) ($options['Concurrency'] ?? 10);
+        $mupThreshold = ((int) ($options['mupThreshold'] ?? 2 * 1024)) * $megabyte;
+        unset($options['Concurrency'], $options['mupThreshold']);
+        /*
+         * The maximum number of parts is 10.000. The partSize must be a power of 2.
+         * We default this to 64MB per part. That means that we only support to copy
+         * files smaller than 64 * 10 000 = 640GB. If you are coping larger files,
+         * please set PartSize to a higher number, like 128, 256 or 512. (Max 4096).
+         */
+        $partSize = ($options['PartSize'] ?? 64) * $megabyte;
```
jderusse, line 83: If PartSize is not defined, I'd rather default to `max(64 * $megabyte , 2 ** ceil(log($contentLength / 10000, 2)))`;

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +55,114 @@ public function has(string $bucket, string $key): bool
...
+            foreach ($responses as $idx => $response) {
+                try {
+                    /** @var CopyPartResult $copyPartResult */
+                    $copyPartResult = $response->getCopyPartResult();
+                    $parts[] = new CompletedPart(['ETag' => $copyPartResult->getEtag(), 'PartNumber' => $idx]);
+                } catch (\Throwable $e) {
+                    $error = $e;
+
+                    break;
+                }
```
jderusse, line 140: you could move the try/catch outside the loop for more readability, and removing the error variable.

```
try {
    foreach ($responses as $idx => $response) {
        $parts[] = new CompletedPart(['ETag' => $response->getCopyPartResult()->getEtag(), 'PartNumber' => $idx]);
} catch (\Throwable $e) {
       try {
           $response->cancel();
       } catch (\Throwable $e) {
           continue;
       }
    }
    $this->abortMultipartUpload(AbortMultipartUploadRequest::create(['Bucket' => $destBucket, 'Key' => $destKey, 'UploadId' => $uploadId]));
}

```

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -47,6 +55,114 @@ public function has(string $bucket, string $key): bool
...
+                        'PartNumber' => $partNumber,
+                    ])
+                );
+
+                $startByte += $partSize;
+                ++$partNumber;
+                --$parallelChunks;
+            }
+            $error = null;
+            foreach ($responses as $idx => $response) {
```
jderusse, line 131: this is not efficient. We define 10 concurents upload, but in fact, we are uploading 10, waiting for the 10 to finish, and uploading 10 more, ...
That means:
- we are waiting for the 10th to finish before starting the 9 previous.
- if 1 upload takes time to finish, we don't leverage the 9 others.

A Beter implementation would be to use a pool (or buffer), to fill the pool with 10 responses, and once 1 response is over, add a new one before checking the other responses.

Better: leverage async processing, by not waiting for the end of a response, but checking if the response is over, if not check the next one.

```
while not all part uploaded
  while pool not full
    => start new upload
  while pool full
    foreach response in pull
      if response not finished
        continue
      else
        => process response
        => remove response from pool

foreach response in pull
  => wait for response to finish
  => process response
  => remove response from pool
```

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
stof, line 87: this threshold should not be configurable IMO. `upload` does not let the caller choose the threshold for switching to a simple upload instead of a multipart one.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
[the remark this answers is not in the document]
stof: My comment was not about the error handling. It was about the `@var` comment

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
stof: Given that AWS has not yet clarified what is right between their online documentation and the metadata used to generate all their official SDKs (and then reused to generate our async-aws SDK), I think we should not blindly assume that the online doc is the right one. Instead, we should handle `null` values properly as required by static analysis (you're also doing it for the upload id above btw).
Note that we don't have any comparison point in official SDKs that are properly typed (I'm excluding aws-sdk-php from that definition) to see how they handle `CopyPartResult` being absent in the result:

- the TransferManager in the official Java SDK has a `copy` method but it is not actually implemented: https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/S3TransferManager.java#L685-L687
- equivalent features in other SDKs (the `@aws-sdk/lib-storage` of the JS SDK or the TransferManager classes of some other languages) don't have a copy function at all.

PR #1595, opened 2023-10-25
Improve the changelog test

stof:
This avoids requiring a changelog entry when changing scaffolding files as those only impact contributors (who can read the git history if they care about them) rather than users of the package.

### DIFF
--- a/tests/Unit/ChangelogTest.php
+++ b/tests/Unit/ChangelogTest.php
@@ -225,10 +225,19 @@ class ChangelogTest extends TestCase
             } else {
                 continue;
             }
+
+            $subPath = substr($file, \strlen($base));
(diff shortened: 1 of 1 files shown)


stof:
I discovered that when doing some changes in `Makefile` files (a PR should follow soon, but its CI will be red without this change).

One thing I'm wondering is whether I should also allow changing `tests` folders of the sub-packages without requiring a changelog or no (for no, this PR still consider them as meaningful changes).

jderusse:
> I would now vote for allowing test-only changes without changelog entries

I agree. Can you add the test folder in your PR?

stof:
@jderusse done

jderusse (approved):

PR #1596, opened 2023-10-25
Upgrade localstack

stof:
This upgrades the localstack image used for integration test to the current latest version instead of a very old version.

The goal is to try fixing https://github.com/async-aws/aws/pull/1489 which had a failed CI due to a parity mismatch between localstack and AWS for Route53. The bug has been fixed in moto (used by localstack) and I don't know yet whether this is enough to fix the issue entirely in localstack or whether it requires additional changes there (the latest localstack version already uses a version of moto that includes the fix).

Future changes might involve investigating whether our S3 and SQS clients might rely on localstack to run integration tests instead of the testing-s3 and testing-sqs docker images for which we (don't) maintain a fork in this organization (and which rely on tools that have no activity since 2019)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,9 +17,9 @@ $(SUBINITIALIZE): %.initialize:
 
 start-docker: start-docker-s3 start-docker-kms start-docker-localstack
 start-docker-localstack:
-	docker pull localstack/localstack:0.14.2
+	docker pull localstack/localstack:2.3.2
 	docker start async_aws_localstack && exit 0 || \
-	docker run -d -p 4566:4566 -p 4567:4566 -p 4568:4566 -p 4571:4566 -p 4572:4566 -p 4573:4566 -p 4574:4566 -p 4575:4566 -p 4576:4566 -p 4577:4566 -p 4578:4566 -e SERVICES=sts,cloudformation,logs,events,iam,sns,ssm,dynamodb,route53,kinesis,secretsmanager -v /var/run/docker.sock:/var/run/docker.sock --name async_aws_localstack localstack/localstack:0.14.2 && \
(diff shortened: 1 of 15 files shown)


stof:
And now, I need to debug why some integration tests failed in CI and not locally...

GrahamCampbell:
Yeh, I pinned this a while ago because localstack introduced some bugs.

stof:
@GrahamCampbell the weird thing in my case is that things worked locally (unless something is broken in `make clean` making it not clean things properly and so 0.14.2 would still be used when starting containers again)

stof:
And this is precisely an attempt at pinning a newer version because the old version also has bugs.

stof:
The failure in the test for `CreateHostedZone` might actually indicate an actual bug in the way empty lists are populated when parsing XML responses. I will check it in more details.

I also need to investigate the failure in the test for `ChangeResourceRecordSets` to see whether it is a bug in localstack or an issue in our testsuite (maybe our tests are not properly isolated)

stof:
OK, I found why I was not seeing failures locally: the clean commands of the `Makefile` setup were doing a partial clean. So initializing it again would reuse the previous container by starting it again, ignoring any update of the image being pulled.
I submitted #1597 to fix this.

The failure from the test for `ChangeResourceRecordSets` has been fixed by updating the test. The new version of localstack emulates Route53, which implies that a created zone automatically has NS and SOA records which are mandatory in a zone. And so the assertion on the number of listed records needs to account for them.
For the remaining failure, I reported it at https://github.com/localstack/localstack/issues/9486. When trying to track it down, localstack 1.0 was already affected. This might have been the issue that made us pin to 0.14 initially (which would be a bad news if it was not reported upstream at that time).

GrahamCampbell:
I vote that we skip the failing tests and then merge this. I am happy to make the code changes to do that, if anyone else is pushed for time. ❤️

GrahamCampbell:
FYI, v3.0.0 has been tagged on their GitHub repo, so possibly that will be formally released in the next few hours or days. We could do that upgrade here, or do it on the SQS JSON PR.

stof:
As said above, I started doing them. I will finish them either this evening or Saturday.

PR #1605, opened 2023-11-22
Add a script to automate releases

jderusse:


release
```diff
@@ -0,0 +1,190 @@
+#!/usr/bin/env php
+<?php
+
+/**
+ * This script aims to simplify the release process of AsyncAws components
+ * Unless you are a maintainer of AsyncAws, their is nothing interesting in there.
+ *
+ * This script will search for components to release (based on change log and will)
+ * - Find the next version
+ * - Update "branch-alias" section of the composer.json (for minor updates)
```
GrahamCampbell, line 10: Updating the branch alias at release time is wrong and too late. The minor version should be bumped only when new features are actually merged, and right at the same time, and the similarly for the major version, at the point where breaking changes are merged, otherwise there is little point in the branch aliases, and we may be better to remove them if we are not wanting to have them accurately reflect the in progress version?

GrahamCampbell (changes requested):

release
```diff
@@ -0,0 +1,190 @@
...
+    }
+    if (!$lastVersion) {
+        throw new \Exception(sprintf('The "%s" changelog file does not contain a previous version', $changelogPath));
+    }
+    if (!preg_match('/^\d+\.\d+\.\d+$/', $lastVersion)) {
+        throw new \Exception(sprintf('The "%s" changelog file does not contain a valid previous version', $changelogPath));
+    }
+
+    $parts = explode('.', $lastVersion);
+    if (0 !== (int) $parts[0]) {
```
jderusse, line 111:
```suggestion
    if (0 === (int) $parts[0]) {
```

GrahamCampbell: ?

jderusse: this code handles `0.x` version which behaves differently:
- BC-BREAK in minor
- Feature in patch

That's the reason for `$level = min(2, $level + 1);` (the number that should be increased moves to the right)

GrahamCampbell (approved):

PR #1606, opened 2023-11-23
Add Accept header on JSON services

jderusse:


jderusse:
this is not the right move

PR #1607, opened 2023-11-23
Fix incorrect branch alias

GrahamCampbell:
So people on dev stability are not pulling in breaking changes.

### DIFF
--- a/src/Service/Sqs/composer.json
+++ b/src/Service/Sqs/composer.json
@@ -28,7 +28,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.10-dev"
+            "dev-master": "2.0-dev"
         }
     }
 }


GrahamCampbell:
Rebased.

jderusse:
The github action [action](https://github.com/async-aws/aws/blob/master/src/Service/Sqs/.github/workflows/branch_alias.yml) will override your change.

I believe this is not the right way to fix it.

We should either remove it (as you suggested in the other PR) or ask people to set it with the right value (as we do with the phpunit on the changelog)
=> if a changelog contains `BC-BREACK` or `removed` keyword => guess the next major version => assert the composer.json contains the right `branch-alias`

Nyholm:
Thank you

stof:
the github action does not downgrade the major version.

jderusse:
> the github action does not downgrade the major version.

Indeed, hopefully, The GitHub action is listening to tags only (https://github.com/async-aws/sqs/blob/c010d6623da85642410f9f6ac49414ba879e49df/.github/workflows/branch_alias.yml#L5)

PR #1608, opened 2023-11-23
Do not inject InvalidArgument class when not needed

jderusse:
We currently always inject the class, then rely on PhpCsFixer to remove it if unused.

However, when a comment contains the word `InvalidArgument`, then PhpCsFixer keeps it.


GrahamCampbell (approved):
Niiiice

PR #1610, opened 2023-11-24
Add a test to assert branch-alias is well defined

jderusse:


tests/Unit/BranchAliasTest.php
```diff
@@ -0,0 +1,107 @@
...
+ * Rules:
+ * - For component without stable release (ie. 0.2.3):
+ *   - When the ChangeLog file contains a major change => 0.3-dev.
+ *   - When the ChangeLog file contains a minor change => alias absent.
+ *   - When the ChangeLog file contains a patch change or no change => alias absent.
+ * - For component with stable release (ie. 1.2.3):
+ *   - When the ChangeLog file contains a major change => 2.0-dev.
+ *   - When the ChangeLog file contains a minor change => 1.3-dev.
+ *   - When the ChangeLog file contains a patch change or no change => alias absent.
+ */
```
jderusse, line 21: is it correct?

GrahamCampbell: No. `When the ChangeLog file contains a patch change or no change => alias absent.` should be that the alias matches the major.minor from the last release.

GrahamCampbell:
Why are we deleting the aliases?

jderusse:
> Why are we deleting the aliases?

because of https://github.com/async-aws/aws/pull/1610#discussion_r1404171097

PR #1621, opened 2023-12-08
Work around where empty lists are missing from responses

### LINKED ISSUE #1620: SQS Batch Publish Broken
On v2 using the AWS JSON-1.0 protocol: `Undefined array key "Failed"`.

GrahamCampbell:
This is a bug on the AWS server side, but I don't know anyone on the SQS service team to report this to, and basic support ain't gonna allow me to do that. We must thus mitigate on the client side. Fixes #1620.


GrahamCampbell:
I don't believe any of these changes need a change log entry. This is a fix affecting only SQS, which is unreleased, and the other service changes are just a nature of the fact that the generator affects them all, and we probs don't want to hard code sqs into the generator?

PR #1625, opened 2023-12-14
Removed incorrect entry from the SNS change log

GrahamCampbell:


GrahamCampbell:
Ok, so this is odd. I don't see the change in https://github.com/async-aws/sns/compare/1.5.0...1.6.0, but it exists in https://github.com/async-aws/aws/commit/7cfe483d6c118f0719789986ee9c101b9e11aa03.

jderusse:
It has been re-added in https://github.com/async-aws/aws/pull/1598/files#diff-cc29c0f379b4547ed63199f56b8e084efe4d9487a2772ce6011beeaabd8c32f8R419

jderusse (approved):

PR #1629, opened 2023-12-15
gitignore `.idea` folder

jmariller:
This is a folder which is automatically created by PhpStorm/Webstorm and can be safely ignored for Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ composer.lock
 /website/package-lock.json
 /website/template/assets/*
 !/website/template/assets/.gitignore
+.idea


jderusse:
Thanks for your help in improving this package.

However, this is not the correct way to ignore IDE-specific folders, because it requires to:
- add the .gitignore rule in all projects
- add a rule for all IDEs (.idea is Jetbrains IDEs specific)

Instead, it would be best if you used a [global ignore rule on their computer](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer). In that way every contributors can configure their own IDE / preference, and they have to do it only once.

jmariller:
Ah interesting, I did not know about the global ignore rule, thanks for this! Then please do ignore this PR :slightly_smiling_face: I will do the adjustment on my end :+1:

jmariller:
Closing since the folder can be exluded using global git rules as explained by @jderusse (tested it and works like a charm)

PR #1638, opened 2024-01-08
feat: add support for monolog 3

soufianshiawa:


src/Integration/Monolog/CloudWatch/composer.json
```diff
@@ -11,9 +11,9 @@
         "Logs"
     ],
     "require": {
-        "php": "^7.2.5 || ^8.0",
+        "php": ">=8.0",
```
GrahamCampbell, line 14: ^8.0

GrahamCampbell (changes requested):

src/Integration/Monolog/CloudWatch/composer.json
```diff
@@ -11,9 +11,9 @@
         "Logs"
     ],
     "require": {
-        "php": "^7.2.5 || ^8.0",
+        "php": "^8.0",
```
jderusse, line 14: please revert.

Instead of using UnionType, you can either:

a. Remove the typehint from the method, and perform the check at runtime
```
/**
 * @param LogRecord|array $record
 */
protected function write($record): void
  if($record instanceof LogRecord){
    $formatted = $record->formatted;
    $record = array_merge($record->toArray(), ['formatted'=>$formatted]);
  }
  if (!is_array($record)) {
    throw new \TypeError(sprintf('Argument 1 passed to &s must be of the type array or LogRecord, %s given', __METHOD__, get_debug_type($res)));
  }

  // ...
}
```

b. or Use dynamic trait to provide both implementations

similar to https://github.com/symfony/symfony/blob/6.1/src/Symfony/Bridge/Monolog/Handler/CompatibilityHandler.php

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
```diff
@@ -252,4 +262,39 @@ private function send(array $entries): void
 
         $this->client->putLogEvents($data);
     }
+
+    public function handle(LogRecord|array $record): bool
```
jderusse, line 266: Is it needed?

soufianshiawa: @jderusse the idea was to handle both simple array and LogRecord support both simple

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
```diff
@@ -252,4 +271,39 @@ private function send(array $entries): void
...
+    {
+        if(is_array($record)){
+            $message = $record['message']??null;
+            $channel = $record['channel']??null;;
+            $level= $record['level']? Level::fromValue($record['level']) : null;;
+            $context= $record['context']??null;;
+            $extra= $record['extra']??null;;
+            $datetime= $record['datetime'] ? \DateTimeImmutable::createFromMutable($record['datetime']) : null;;
+            $formatted= $record['formatted']??null;;
+            $record = new LogRecord($datetime, $channel, $level, $message, $context, $extra, $formatted);
```
jderusse, line 288: This class does not exist in monolog 2

src/Integration/Monolog/CloudWatch/tests/CloudWatchLogsHandlerTest.php
```diff
@@ -44,7 +44,7 @@ public function testConfig()
             'stream' => $this->streamName,
         ], Logger::CRITICAL, false);
 
-        self::assertEquals(Logger::CRITICAL, $handler->getLevel());
+        self::assertEquals(Logger::CRITICAL, $handler->getLevel()->value);
```
jderusse, line 47: This won't work for tests runing on lower version of PHP.
You should add something like 
```php
if (enum_exists(Level)) {
  self::assertEquals(Logger::CRITICAL, $handler->getLevel()->value);
} else {
  self::assertEquals(Logger::CRITICAL, $handler->getLevel());
}
```

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
jderusse: If I'm not wrong, this method is already implemented in `AbstractProcessingHandler`

Which already handles `array` when Monolog 2 is installed and `LogRecord` when version 3 is installed.

I don't understand why we have to provide an implementation. Please tell me a case (little reproducer/gadget) to show me the issue.

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
```diff
@@ -126,10 +126,20 @@ protected function getDefaultFormatter(): FormatterInterface
...
 
     /**
-     * {@inheritdoc}
+     * @param \Monolog\LogRecord|array $record
      */
-    protected function write(array $record): void
+    protected function write($record): void
     {
+        if (version_compare(PHP_VERSION, '8', '>=')) {
+            if($record instanceof \Monolog\LogRecord){
```
jderusse, line 134: plese use `use Monolog\LogRecord`; instead of FQDN

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
```diff
@@ -126,10 +126,20 @@ protected function getDefaultFormatter(): FormatterInterface
     }
 
     /**
-     * {@inheritdoc}
+     * @param \Monolog\LogRecord|array $record
      */
-    protected function write(array $record): void
+    protected function write($record): void
     {
+        if (version_compare(PHP_VERSION, '8', '>=')) {
```
jderusse, line 133: Why do we need to check the PHP version?

soufianshiawa: tests fails when i switch down to older version of php needs to handle the cases separately

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
soufianshiawa: handling different version  of php lead to this if statement  did use FQDN to avoid importing unrecognizable dependencies for older version.

jderusse: in PHP it's totally fine to `use` a class/interface that does not exist.

You can even use the `TheClass::class` in your code.

see https://3v4l.org/QclWd

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
jderusse: I believe there is something wrong elsewhere.

`if ($arg instanceof Bar) {` works since at least PHP 7.0

https://3v4l.org/Ua4sF#v7.0.33

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
soufianshiawa: trying to call it only when we use monolog 3 by the if statement check

src/Integration/Monolog/CloudWatch/tests/CloudWatchLogsHandlerTest.php
soufianshiawa: rolled back

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
soufianshiawa: fixed

src/Integration/Monolog/CloudWatch/src/CloudWatchLogsHandler.php
soufianshiawa: fixed

jderusse:
thank you @soufianshiawa

PR #1646, opened 2024-01-11
Update event-bridge.md

Eviltoastey:
Change eventbus name to ARN. EventBusName should have the arn of the eventbus else it will not work. In the source code of async-aws it also says it needs an ARN not a eventbus name


jderusse:
Thank you @Eviltoastey

PR #1655, opened 2024-01-22
Event bridge - Optimization of Connection Persistence in HTTP Client

lukasojd:
This merge request addresses a significant performance issue in the HTTP client, specifically relating to the management of keep-alive connections. Previously, when only the status code of the HTTP response was retrieved without accessing the content, it inadvertently led to the termination of keep-alive connections on the client side (particularly noticeable with curl). This behavior resulted in the need for re-establishing connections and performing a new handshake for subsequent requests. Such a process introduced a considerable delay ranging between 30-80 ms for each request, substantially impacting the overall efficiency.

The proposed changes in this merge request aim to optimize this behavior. By ensuring that the content of the HTTP response is accessed ($this->httpResponse->getContent(false)) even when a timeout is not specified (null === $timeout), we maintain the persistence of the keep-alive connection. This approach effectively avoids the overhead of re-establishing connections and performing handshakes for subsequent requests, thereby enhancing the performance of the HTTP client.

These changes are crucial for scenarios where high-frequency requests are made, as it significantly reduces the cumulative latency. This optimization is particularly beneficial for applications requiring fast and efficient communication with AWS services.

Included is the updated code snippet for the resolve function, which now ensures that the content of the HTTP response is accessed irrespective of the timeout condition, thereby maintaining the keep-alive connection state effectively.

### DIFF
--- a/src/Core/src/Response.php
+++ b/src/Core/src/Response.php
@@ -145,13 +145,11 @@ final class Response
         try {
             if (null === $timeout) {
                 $this->httpResponse->getStatusCode();
+                $this->httpResponse->getContent(false);
             } else {
                 foreach ($this->httpClient->stream($this->httpResponse, $timeout) as $chunk) {
                     if ($chunk->isTimeout()) {
                         return false;
-                    }
-                    if ($chunk->isFirst()) {
-                        break;
                     }
                 }
             }
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## NOT RELEASED
 
+## 1.20.2
+
+### Fixed
+
+- Optimization of Connection Persistence in HTTP Client
+
 ## 1.20.1
 
 ### Changed
--- a/src/Core/tests/Unit/Credentials/DateFromResultTest.php
+++ b/src/Core/tests/Unit/Credentials/DateFromResultTest.php
@@ -5,24 +5,17 @@ namespace AsyncAws\Core\Tests\Unit\Credentials;
 use AsyncAws\Core\Credentials\DateFromResult;
 use AsyncAws\Core\Response;
 use AsyncAws\Core\Result;
+use AsyncAws\Core\Test\Http\SimpleMockedResponse;
 use PHPUnit\Framework\TestCase;
 use Psr\Log\NullLogger;
 use Symfony\Component\HttpClient\MockHttpClient;
-use Symfony\Component\HttpClient\Response\MockResponse;
 
 class DateFromResultTest extends TestCase
 {
     public function testWithValidDate()
     {
-        $httpResponse = $this->getMockBuilder(MockResponse::class)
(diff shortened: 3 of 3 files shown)


GrahamCampbell:
Interesting. Is this still happening if you set the max connections to 1 on the curl http client, without using this change?

src/Core/src/Response.php
```diff
@@ -145,14 +145,12 @@ public function resolve(?float $timeout = null): bool
         try {
             if (null === $timeout) {
                 $this->httpResponse->getStatusCode();
```
GrahamCampbell, line 147: This line is no longer needed, due to the line you added after.

GrahamCampbell (changes requested):

lukasojd:
Fix please test cc @lukasojd

lukasojd:
> Interesting. Is this still happening if you set the max connections to 1 on the curl http client, without using this change?

I have conducted tests using the default Symfony client (Symfony\Component\HttpClient\CurlHttpClient), specifically setting the maximum number of host connections to 1. Here's the relevant code snippet:

```
$client = new CurlHttpClient(maxHostConnections: 1);
$this->client = new EventBridgeClient(
    configuration: [
        'accessKeyId' => $this->accessKeyId,
        'accessKeySecret' => $this->accessKeySecret,
        'region' => $this->region,
    ],
    httpClient: $client,
    logger: $this->logger,
);
```

Despite this adjustment, the issue persists. The configuration was set up as per the example above, ensuring that only one connection per host is maintained at any time. However, it seems that this configuration change alone does not resolve the problem.

Additionally, I have also monitored the network traffic and confirmed that the issue indeed involves the closing of connections when the response is not fully retrieved. This results in repeated packets associated with connection re-establishment and the SSL handshake process. It's evident from the network analysis that each new request initiates a fresh connection cycle, including the SSL handshake, which aligns with the observed performance issue.

jderusse:
@nicolas-grekas does it rings a bell?

We don't consume the entire response to save few IO. But it looks like the side effect is,  the response is not free for curl, and it don't reuse it.

How can we tell SymfonyClient to stop streaming the response, but keep the connection alive?

lukasojd:
@GrahamCampbell @jderusse  @nicolas-grekas I've continued troubleshooting the issue and discovered that the problem is related to specific versions of curl. It appears that versions prior to 7.87.0 are affected, while the issue has been resolved starting from curl 7.87.0. Considering this, I believe it may not be necessary to insist on this merge request. The resolution might simply involve updating to a newer version of curl. I leave the decision up to you on how to proceed further.

nicolas-grekas:
Thanks for digging!
Given [7.87.0 is one year old](https://curl.se/docs/vuln-7.87.0.html), it might be fine doing nothing indeed.
Unless we're sure that fetching all the body won't generate needless network traffic (because responses are small) in which case why not.

GrahamCampbell:
Using the amphp http client may be a suitable work-around for you?

PR #1660, opened 2024-02-02
Remove discoverEndpoints when not used anymore

jderusse:
The current logic of the `CodeGenerator` registers the operation when it is flagged `EndpointOperation` (the generator adds a new `discoverEndpoints` method that overrides the method from AbstractApi) 

The issue is, AWS removes the flag `EndpointOperation` on some service but our logic was not able to detect the removal of a flag.

This removes the `discoverEndpoints` method when there isn't operations flagged anymore

This should remove deprecations in https://github.com/async-aws/aws/pull/1657

I had to remove your custom patch @stof . Could you check that endpoint discovery is not needed anymore?

### DIFF
--- a/src/CodeGenerator/src/Definition/ServiceDefinition.php
+++ b/src/CodeGenerator/src/Definition/ServiceDefinition.php
@@ -65,6 +65,11 @@ class ServiceDefinition
      */
     private $apiReferenceUrl;
 
+    /**
+     * @var false|Operation|null
+     */
+    private $cachedEndpointOperation = false;
+
     public function __construct(string $name, array $endpoints, array $definition, array $documentation, array $pagination, array $waiter, array $example, array $hooks, string $apiReferenceUrl)
     {
         $this->name = $name;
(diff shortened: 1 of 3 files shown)


stof:
the `discoverEndpoints` in MediaConvert is not related to a removed flag in EndpointOperation.

It is a manually-written implementation of the method, to provide automatic discovery of the account-specific endpoint instead of making the MediaConvertClient as painful to use than the official SDK. See https://github.com/async-aws/aws/issues/1414

stof:
Oh, actually, it seems like they removed the need to use account-specific endpoints, according to the deprecation in the PHP SDK. But their online doc at https://docs.aws.amazon.com/mediaconvert/latest/apireference/important-notes.html still mentions it.

But instead of changing the code generator to forbid defining a custom `discoverEndpoints` implementation and turning on `endpointdiscovery` for operations when there is no EndpointOperation, I would rather suggest removing manually the custom implementation of the method and removing the custom patches in the manifest that were about making other operations use the endpoint discovery flags to call this method (even though AWS does not actually use the endpoint discovery mechanism for this service). I'll send a PR for that.

PR #1661, opened 2024-02-02
Use the regional endpoint for MediaConvert

stof:
This removes the discovery layer used to discover automatically the account-specific endpoint that was required previously.


stof:
I confirmed that it indeed works fine in my project.

jderusse (approved):

PR #1666, opened 2024-02-16
Add endpoint URL env variable to fallback options

matteokov:
Currently, Async AWS supports almost all environment variables from the official SDK (list is published [here](https://docs.aws.amazon.com/sdkref/latest/guide/settings-reference.html#EVarSettings)). `AWS_ENDPOINT_URL` is currently not supported but can be very useful for local development with LocalStack or a similar service by just providing it to the `.env` file. Here I added it to the list of fallback options. 

Future improvements can include `AWS_ENDPOINT_URL_<SERVICE>` as well.

### DIFF
--- a/src/Core/tests/Unit/ConfigurationTest.php
+++ b/src/Core/tests/Unit/ConfigurationTest.php
@@ -70,6 +70,7 @@ class ConfigurationTest extends TestCase
 
         yield 'default when env missing' => [[], [], ['profile' => 'default']];
         yield 'fallback env' => [[], ['AWS_PROFILE' => 'foo'], ['profile' => 'foo']];
+        yield 'fallback endpoint env' => [[], ['AWS_ENDPOINT_URL' => 'http://localhost:4566'], ['endpoint' => 'http://localhost:4566']];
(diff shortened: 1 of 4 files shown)


jderusse (commented):
Thank you.
Could you add a test in `src/Core/tests/Unit/ConfigurationTest.php` please

src/Core/tests/Unit/ConfigurationTest.php
```diff
@@ -93,4 +93,17 @@ public function testRegionFromIniFiles()
         self::assertFalse($config->isDefault('region'));
         self::assertEquals('eu-central-1', $config->get('region'));
     }
+
+    public function testEndpointFromEnv()
+    {
```
jderusse, line 98: couldn't you use the existing provider at line 78 instead?

matteokov: Removed dedicated test and replaced with additional case in the provider. This case tests the same functionality as `fallback env` but for the newly added env. Do you want me to replace the env inside `fallback env`  with `AWS_ENDPOINT_URL` or is it okay to have both?

jderusse (approved):

jderusse:
Thank you @matteokov

PR #1672, opened 2024-02-22
[SymfonyBundle] feat: Add support for MaxResults on SSMVault

tyx:
To reduce the number of requests sent to SSM

### DIFF
--- a/src/Integration/Symfony/Bundle/tests/Unit/DependencyInjection/ConfigurationTest.php
+++ b/src/Integration/Symfony/Bundle/tests/Unit/DependencyInjection/ConfigurationTest.php
@@ -39,6 +39,7 @@ class ConfigurationTest extends TestCase
                     'pool' => 'cache.system',
                     'ttl' => 600,
                 ],
+                'max_results' => null,
             ],
         ]);
     }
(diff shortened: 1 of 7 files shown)


src/Integration/Symfony/Bundle/tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -39,6 +39,7 @@ public function testDefaultValues(): void
                     'pool' => 'cache.system',
                     'ttl' => 600,
                 ],
+                'max_results' => null,
```
jderusse, line 42: Are you sure that `null`, won't be sent in the request (and AWS won't reject the payload)?

jderusse (approved):
Just a minor question, otherwise LGTM

src/Integration/Symfony/Bundle/tests/Unit/DependencyInjection/ConfigurationTest.php
tyx: According to the `GetParametersByPathRequest` it should not be a problem. `MaxResults` is null by default.

https://github.com/async-aws/aws/blob/125a4b95d49fff611f24ce7f6150ba4528727bd8/src/Service/Ssm/src/Input/GetParametersByPathRequest.php#L88

So I guess SSMClient already manage it correctly no ?

jderusse:
Thank you @tyx

PR #1676, opened 2024-03-03
Add possibility to use locking with DynamoDB sessions

valtzu:
DynamoDB session handler in this library not supporting session locking makes it really difficult to migrate applications using the official AWS SDK with locking (or default file handler with locking) to `async-aws/dynamo-db-session`.

Unless there are some valid arguments not to support locking, I'd propose let's add it.

### DIFF
--- a/src/Integration/Aws/DynamoDbSession/src/SessionHandler.php
+++ b/src/Integration/Aws/DynamoDbSession/src/SessionHandler.php
@@ -7,6 +7,8 @@ use AsyncAws\DynamoDb\DynamoDbClient;
 use AsyncAws\DynamoDb\Enum\BillingMode;
 use AsyncAws\DynamoDb\Enum\KeyType;
 use AsyncAws\DynamoDb\Enum\ScalarAttributeType;
+use AsyncAws\DynamoDb\Exception\ConditionalCheckFailedException;
+use AsyncAws\DynamoDb\ValueObject\AttributeValue;
 
 class SessionHandler implements \SessionHandlerInterface
 {
@@ -17,13 +19,17 @@ class SessionHandler implements \SessionHandlerInterface
 
     /**
      * @var array{
-     *   consistent_read?: bool,
+     *   consistent_read: bool,
      *   data_attribute: string,
      *   hash_key: string,
      *   session_lifetime?: int,
      *   session_lifetime_attribute: string,
      *   table_name: string,
-     *   id_separator: string
+     *   id_separator: string,
+     *   locking: bool,
+     *   max_lock_wait_time: float,
+     *   min_lock_retry_microtime: int<0, max>,
+     *   max_lock_retry_microtime: int<0, max>,
      * }
      */
     private $options;
@@ -56,7 +62,11 @@ class SessionHandler implements \SessionHandlerInterface
      *   session_lifetime?: int,
      *   session_lifetime_attribute?: string,
      *   table_name: string,
-     *   id_separator?: string
+     *   id_separator?: string,
+     *   locking?: bool,
+     *   max_lock_wait_time?: int|float,
+     *   min_lock_retry_microtime?: int<0, max>,
+     *   max_lock_retry_microtime?: int<0, max>,
      * } $options
      */
     public function __construct(DynamoDbClient $client, array $options)
@@ -66,6 +76,11 @@ class SessionHandler implements \SessionHandlerInterface
         $options['hash_key'] = $options['hash_key'] ?? 'id';
         $options['session_lifetime_attribute'] = $options['session_lifetime_attribute'] ?? 'expires';
         $options['id_separator'] = $options['id_separator'] ?? '_';
+        $options['consistent_read'] = $options['consistent_read'] ?? true;
+        $options['locking'] = (bool) ($options['locking'] ?? false);
+        $options['max_lock_wait_time'] = (float) ($options['max_lock_wait_time'] ?? 10.0);
+        $options['min_lock_retry_microtime'] = (int) ($options['min_lock_retry_microtime'] ?? 10000);
+        $options['max_lock_retry_microtime'] = (int) ($options['max_lock_retry_microtime'] ?? 50000);
         $this->options = $options;
     }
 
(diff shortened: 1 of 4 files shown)


src/Integration/Aws/DynamoDbSession/src/SessionHandler.php
```diff
@@ -66,6 +76,11 @@ public function __construct(DynamoDbClient $client, array $options)
         $options['hash_key'] = $options['hash_key'] ?? 'id';
         $options['session_lifetime_attribute'] = $options['session_lifetime_attribute'] ?? 'expires';
         $options['id_separator'] = $options['id_separator'] ?? '_';
+        $options['consistent_read'] = $options['consistent_read'] ?? true;
+        $options['locking'] = (bool) ($options['locking'] ?? false);
```
jderusse, line 80: I think the cast to bool (same for float/int below) is not necessary

The type is already defined in the doc bloc and should be respected.

src/Integration/Aws/DynamoDbSession/src/SessionHandler.php
```diff
@@ -238,4 +254,45 @@ private function formatKey(string $key): array
...
+                    'Key' => $key,
+                    'ConsistentRead' => $this->options['consistent_read'],
+                    'Expected' => ['lock' => ['Exists' => false]],
+                    'AttributeUpdates' => ['lock' => ['Value' => ['N' => '1']]],
+                    'ReturnValues' => 'ALL_NEW',
+                ])->getAttributes();
+            } catch (ConditionalCheckFailedException $e) {
+                // If we were to exceed the timeout after sleep, let's give up immediately.
+                $sleep = rand($this->options['min_lock_retry_microtime'], $this->options['max_lock_retry_microtime']);
+                if (microtime(true) + $sleep * 1e-6 > $timeout) {
```
jderusse, line 290: This could be improved. to sleep at max "remaining time to timeout"
and throw an exception only of timeout is reached (aka when sleep < 0)

src/Integration/Aws/DynamoDbSession/src/SessionHandler.php
```diff
@@ -238,4 +254,45 @@ private function formatKey(string $key): array
...
+                    'TableName' => $this->options['table_name'],
+                    'Key' => $key,
+                    'ConsistentRead' => $this->options['consistent_read'],
+                    'Expected' => ['lock' => ['Exists' => false]],
+                    'AttributeUpdates' => ['lock' => ['Value' => ['N' => '1']]],
+                    'ReturnValues' => 'ALL_NEW',
+                ])->getAttributes();
+            } catch (ConditionalCheckFailedException $e) {
+                // If we were to exceed the timeout after sleep, let's give up immediately.
+                $sleep = rand($this->options['min_lock_retry_microtime'], $this->options['max_lock_retry_microtime']);
```
jderusse, line 289: use random_int instead please

src/Integration/Aws/DynamoDbSession/src/SessionHandler.php
```diff
@@ -66,6 +76,11 @@ public function __construct(DynamoDbClient $client, array $options)
         $options['hash_key'] = $options['hash_key'] ?? 'id';
         $options['session_lifetime_attribute'] = $options['session_lifetime_attribute'] ?? 'expires';
         $options['id_separator'] = $options['id_separator'] ?? '_';
+        $options['consistent_read'] = $options['consistent_read'] ?? true;
+        $options['locking'] = (bool) ($options['locking'] ?? false);
+        $options['max_lock_wait_time'] = (float) ($options['max_lock_wait_time'] ?? 10.0);
+        $options['min_lock_retry_microtime'] = (int) ($options['min_lock_retry_microtime'] ?? 10000);
```
jderusse, line 82: I'm not sure such thing deserve parameters.

jderusse (commented):
If I understand correctly,
Lock is acquired when reading the session
and released when writing the session.

This sound really fragile to me.
- What if session is written several time?
- What if the session is not written (just read and close)
- What if the app crashes before writing the session? Session will stay locked forever?

The last question make me think, this is not the right way, and we should use symfony/lock in a decorator instead

valtzu:
Thanks for the quick review @jderusse!

> Lock is acquired when reading the session and released when writing the session.
> This sound really fragile to me.

The logic is mostly copied from the [official aws-sdk](https://github.com/aws/aws-sdk-php/blob/master/src/DynamoDb/LockingSessionConnection.php), so if you think it's fragile, then the official sdk is fragile?

---

> What if the session is not written (just read and close)

On `close` the the `expires` is updated if a write didn't occur, so we're already doing a write even when there are no changes.

https://github.com/async-aws/aws/blob/bc92ccef1214cefac04d6fcb862f2b6df6242862/src/Integration/Aws/DynamoDbSession/src/SessionHandler.php#L115-L116

---

> What if session is written several time?

1. I belive the official aws-sdk has the same "issue" (maybe it's a non-issue because of number 2)
2. Is it even possible? As I see php only has [`session_write_close`](https://www.php.net/manual/en/function.session-write-close.php), there is no method to trigger write without close? Of course if you don't use the native interface, then yeah

---

> What if the app crashes before writing the session? Session will stay locked forever?

Right now, correct, the same as with the official aws-sdk. But of course we could store a timestamp in `lock` instead of boolean to have some timeout after which the entry would be no longer considered locked.

---

> we should use symfony/lock in a decorator instead

While in general I like the idea, I'm not sure how great that would be in terms of performance & reliability (having separate storage/requests/connections for locks etc) + again it would defeat the purpose of this PR which is to make it easier to migrate from `aws-sdk` or native file-based PHP session handling to this library.

If we were to use `symfony/lock` instead, would such decorator have a place in this project or you meant it as a suggestion to just use `symfony/lock` and write a decorator in my project?

src/Integration/Aws/DynamoDbSession/src/SessionHandler.php
valtzu: It's only respected if the project where you're using this is using static analyzer. But ok :heavy_check_mark:

src/Integration/Aws/DynamoDbSession/src/SessionHandler.php
valtzu: As I understand DynamoDB usually is billed per-request, so giving a possibility to fine-tune these according to your use case may be useful in terms of balancing the functionality & cost-savings. For example, if your sessions usually are locking for several seconds, you likely don't want to be polling every 10-50ms.

Also, these settings are just copied from aws-sdk to make a migration to async-aws as small change as possible.

PR #1689, opened 2024-03-19
Fix deprecation reporting

jderusse:


jderusse:
I don't get why PHP Unit fails in 8.4. Looks like related to indirect depreciation.
Any clue @nicolas-grekas

PR #1717, opened 2024-05-27
[Symfony Bundle] Fix the usage of a deprecated class

javiereguiluz:
Fixes this issue spotted while updating some app to Symfony 7.1:

```
The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered
internal since Symfony 7.1, to be deprecated in 8.1;
use Symfony\Component\DependencyInjection\Extension\Extension instead.
It may change without further notice. You should not use it from 
"AsyncAws\Symfony\Bundle\DependencyInjection\AsyncAwsExtension".
```

### DIFF
--- a/src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
+++ b/src/Integration/Symfony/Bundle/src/DependencyInjection/AsyncAwsExtension.php
@@ -15,8 +15,8 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Reference;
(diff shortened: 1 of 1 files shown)


jderusse:
Hello, tests are failing. Could you please add an entry in the Symfony/Bundle changelog file?

javiereguiluz:
Do we really need an entry in the changelog? The new Symfony class is available since Symfony 5.4, so this change should be completely transparent to end users 🤔

src/Integration/Symfony/Bundle/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## NOT RELEASED
 
+- The extension class now extends Symfony\Component\DependencyInjection\Extension\Extension
```
jderusse, line 5:
```suggestion
### Changed

- Replace deprecated Extension by  Symfony\Component\DependencyInjection\Extension\Extension
```

jderusse:
> Do we really need an entry in the changelog? The new Symfony class is available since Symfony 5.4, so this change should be completely transparent to end users 🤔 In any case, I added the entry. Thanks.

It's required, to automate the releasing of components. I have a script that use the changelog to define the next version (patch / minor)

jderusse:
Thank you @javiereguiluz

PR #1724, opened 2024-06-11
Fix exception not created from HEAD requests

### LINKED ISSUE #1723: async-aws/s3 does not properly throw a NoSuchKey error for invalid keys in headObject
Do a `S3Client::headObject` call with a `Key` that does not exist in the bucket.

As a `<head>` request does not have a body, the error parsing does not find the details in the response body. And so it creates a generic ClientException with no AwsError in it.

The official SDK has a condition to trigger the usage of the `NoSuchKey` error shape based on the response status code, which is what allows it to support this case properly:
https://github.com/aws/aws-sdk-php/blob/53b233072f707d2abfd62b0df4bc2dc27caf3274/src/data/s3/2006-03-01/api-2.json#L5973-L5979

jderusse:
fix #1723

IIUC the manifest, all errors are "linked" to a `httpStatusCode`, but we don't check it.

We might need a deep refactor that:
- check for the list of exceptions by HTTP status
- if the response doesn't have a body, and we have a single exception available for the httpStatus code, then we can use it.

any opinion?

### DIFF
--- a/src/CodeGenerator/src/Generator/OperationGenerator.php
+++ b/src/CodeGenerator/src/Generator/OperationGenerator.php
@@ -198,6 +198,10 @@ class OperationGenerator
             $classBuilder->addUse($errorClass->getFqdn());
 
             $mapping[] = sprintf('%s => %s::class,', var_export($error->getCode() ?? $error->getName(), true), $errorClass->getName());
+            if ('HEAD' === $operation->getHttpMethod() && ($statusCode = $error->getStatusCode()) >= 400) {
+                $this->requirementsRegistry->addRequirement('async-aws/core', '^1.22');
+                $mapping[] = sprintf('%s => %s::class,', var_export('http_status_code_' . $statusCode, true), $errorClass->getName());
+            }
         }
         if ($mapping) {
             $extra .= ", 'exceptionMapping' => [\n" . implode("\n", $mapping) . "\n]";
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Added
+
+- Added support for exception based on response http status code only.
+
 ## 1.21.0
 
 ### Added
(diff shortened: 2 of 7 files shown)


src/CodeGenerator/src/Generator/OperationGenerator.php
```diff
@@ -198,6 +198,9 @@ private function setMethodBody(Method $method, Operation $operation, ClassName $
             $classBuilder->addUse($errorClass->getFqdn());
 
             $mapping[] = sprintf('%s => %s::class,', var_export($error->getCode() ?? $error->getName(), true), $errorClass->getName());
+            if ($operation->getHttpMethod() === 'HEAD' && ($statusCode = $error->getStatusCode()) >= 400) {
```
stof, line 201: this should declare a dependency on `async-aws/core` at version `^1.22` inside this `if`, as this will be the version of `async-aws/core` that contains the logic making it work (otherwise, the bug in async-aws/s3 might not be fixed if it installs an older version of the core).

We have the infrastructure to handle that properly:
```suggestion
            if ($operation->getHttpMethod() === 'HEAD' && ($statusCode = $error->getStatusCode()) >= 400) {
                $this->requirementsRegistry->addRequirement('async-aws/core', '^1.22');
```

stof (changes requested):

src/CodeGenerator/src/Generator/OperationGenerator.php
jderusse: good catch, thanks for the hint

stof:
The CI failure on PHP 8.4 is because of deprecations triggered by PHP in the symfony/polyfill code. This is solved in https://github.com/symfony/polyfill/pull/467 but not released yet

stof:
And the root tests were broken by https://github.com/async-aws/aws/pull/1722 which has been merged with a failing CI

jderusse:
> For the failure in root tests, I opened #1726 to fix it.

tests are green except for 8.4 because of deprecation in polyfill

stof (approved):

PR #1726, opened 2024-06-12
Fix the branch alias for the location-service package

stof:
#1722 broke the root tests by having a branch alias that does not match the changelog (probably due to the second commit doing manual changes to the changelog) and was merged with the failing CI jobs. this is meant to fix those jobs (this PR does not fix the PHP 8.4 job, which is failing for a different reason)

### DIFF
--- a/src/Service/LocationService/composer.json
+++ b/src/Service/LocationService/composer.json
@@ -28,7 +28,7 @@
     },
     "extra": {
         "branch-alias": {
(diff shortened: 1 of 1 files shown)


stof:
The fix for the PHP 8.4 job will be having a new release of symfony/polyfill, which should happen in a few days hopefully.

stof:
Duplicate of https://github.com/async-aws/aws/pull/1725

Issue #1746, opened 2024-08-07
async-aws/sqs: Message deletion not working with ElasticMQ


akuma91:
Hey everyone

When I try to consume a message from ElasticMQ using Symfony messenger, I receive the following error upon trying to delete the message from the queue:

``` 
In Response.php line 432:
  HTTP 400 returned for "http://elasticmq:9324/".

  Code:    InvalidAction
  Message: Invalid request: UnacceptedResponseContentTypeRejection(Set(ContentType(none/none)))
  Type:
  Detail:
```

This seems to happen due to the `Accept` header `application/json` added in async-aws/sqs version 2.1.1 in the `src/Input/DeleteMessageRequest.php`.
Downgrading to async-aws/sqs version 2.1.0 (or removing the Accept header manually) fixes the issue.

Is there anything that can be done about this?

Setup:
- PHP 8.3.9
- ElasticMQ version 1.6.5
- Symfony application using `symfony/messenger` v6.4.9 and `symfony/amazon-sqs-messenger` v6.4.9
- async-aws/core v1.22.0
- async-aws/sqs v2.1.1

jderusse:
Looks like it's a bug in ElasticMQ (https://github.com/softwaremill/elasticmq/issues/1027).

Could you try to update your docker image?

akuma91:
Thanks for the advice, that fixed it!

PR #1747, opened 2024-08-07
Remove accept header for endpoint without response body

jderusse:
Fixe #1746


GrahamCampbell:
Kinda odd. The accept header is still useful to direct that we want JSON responses in error bodies.

GrahamCampbell:
Won't omitting this, risk breaking that?

jderusse:
> Kinda odd. The accept header is still useful to direct that we want JSON responses in error bodies.

That's true. 

IMHO that's an issue in the server. Not the client.
And it looks like async-aws is not alone : https://github.com/softwaremill/elasticmq/issues/1027#issuecomment-2250043025

Issue #1186, opened 2022-02-24
`Result::wait` & exceptions


GrahamCampbell:
It would be cool to be able to use the result waiter in such a way where it batches up exceptions too, rather than crashing everything out. I have a use case where a particular error from AWS is a "success".

jderusse:
Hmmm that should be already handled... For instance, the S3 ObjectNotExistsWaiter wait until the operation returns a 404 error.

Do you have a concrete example ?

GrahamCampbell:
Starting an execution on a step function. I want to send different tasks with the same task name, which result in a task already exists exception.

GrahamCampbell:
I wanna do like 100 parallel requests, and wait for them all to complete, then see which failed.

jderusse:
ohhh... I mixed Waiters, and `Result::wait` (to iterate over several requests).

But this should already be handled... The `Result::wait` does not throw an exception, it yields results. 

But consuming the result might throw an exception (beware that calling `__destruct` consumes the result and might throw an exception)


Here is a POC that work as expected: When consuming failed responses, It displays the 404 message, but the loops did not crashes.

```
$requests = [];
$requests[] = $client->getObject(['Bucket' => 'x', 'Key' => 'file_exists']);
$requests[] = $client->getObject(['Bucket' => 'x', 'Key' => 'does-not-exists']);
$requests[] = $client->getObject(['Bucket' => 'x', 'Key' => 'does-not-exists']);
$requests[] = $client->getObject(['Bucket' => 'x', 'Key' => 'file_exists']);
$requests[] = $client->getObject(['Bucket' => 'x', 'Key' => 'does-not-exists']);
$requests[] = $client->getObject(['Bucket' => 'x', 'Key' => 'does-not-exists']);


foreach (Result::wait($requests) as $response) {
    try {
        dump($response->getContentLength());
    } catch (\Exception $e) {
        dump($e->getMessage());
    }
}
```

GrahamCampbell:
Right, so I just have to be careful to store the result of the call to wait, and then unset it, otherwise destruct will be called immediately for each response?

GrahamCampbell:
Am I right in thinking this code would crash before completing necessarily all of the requests?

```
$requests = [];
$requests[] = $client->getObject(['Bucket' => 'x', 'Key' => 'file_exists']);
$requests[] = $client->getObject(['Bucket' => 'x', 'Key' => 'does-not-exists']);
[...]

PR #1755, opened 2024-08-26
Clarify how to batch requests when the results are not needed

mnapoli:
It took me a while and https://github.com/async-aws/aws/issues/1186#issuecomment-1054241642 to understand this 😄


jderusse:
Thank you @mnapoli

PR #1773, opened 2024-10-01
Add SQS AddPermission operation

stevenbrookes:
New SQS AddPermission operation, required to allow 3rd parties the permission to send messages to queues created using the createQueue endpoint.  

I didn't seem to need to do anything after using the generate script - very cool.  Let me know if I've missed anything.

As it looks like you are using a monorepo how would I go about using my fork of that in my code while the PR is reviewed and hopefully accepted.

Thanks

Steve

### DIFF
--- a/src/Service/Sqs/CHANGELOG.md
+++ b/src/Service/Sqs/CHANGELOG.md
@@ -2,13 +2,13 @@
 
 ## NOT RELEASED
 
-### Changed
-
-- Enable compiler optimization for the `sprintf` function.
-
 ### Added
 
 - Add AddPermission endpoint
+
+### Changed
+
+- Enable compiler optimization for the `sprintf` function.
 
 ## 2.1.1
 
--- a/src/Service/Sqs/CHANGELOG.md
+++ b/src/Service/Sqs/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 
 - Enable compiler optimization for the `sprintf` function.
+- Add AddPermission endpoint
 
 ## 2.1.1
 
--- a/src/Service/Sqs/CHANGELOG.md
+++ b/src/Service/Sqs/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Changed
 
 - Enable compiler optimization for the `sprintf` function.
+
+### Added
+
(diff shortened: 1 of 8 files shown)


GrahamCampbell:
> I didn't seem to need to do anything after using the generate script - very cool. Let me know if I've missed anything.

You need to write the tests.

stevenbrookes:
Test added. I seem to be failing the change log test.  I added an entry, but perhaps not the right format?  Any advice?

src/Core/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Enable compiler optimization for the `sprintf` function.
 - Avoid calls to spl_object_ methods when computing cache key.
+- Add Sqs AddPermission operation
```
jderusse, line 9: You changed the wrong `CHANGELOG` file. This one is fore changes in the `Core` sub package.
You should edit the `./src/Service/Sqs/CHANGELOG`

stevenbrookes: Ah got it.  Right that seems to be passing now apart from the dev check you mentioned.  I've also tested this now on a real account and works as expected.

src/Service/Sqs/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changed
 
 - Enable compiler optimization for the `sprintf` function.
+- Add AddPermission endpoint
```
jderusse, line 8: should be moved in a new `### Added` section

(This is used to guess if next release should be a patch version (bug fix) or minor version (new features)

jderusse:
The last failing test is about bumping the minor version.

Because this PR adds a new feature, the next version of the package will change.
You need to edit the `src/Service/Sqs/composer.json` file and set `2.2-dev` in the `branch-alias` section

stevenbrookes:
Yay finally passed the tests.  May I suggest adding the additional things that need to be done to the contributing documentation so aid future contributors?  I look forward to this PR being accepted.

jderusse:
thank you @stevenbrookes 

Please yes, fill free to improve the documentation :pray:

PR #1781, opened 2024-10-16
Add support for EKS Pod Identity

kevinrudde:
Added support for EKS Pod Identity, see https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html

It is similar to the `ContainerProvider`, but uses the injected environment variables from the Pod Identity Agent.


src/Core/src/Credentials/PodIdentityProvider.php
```diff
@@ -0,0 +1,89 @@
...
+use Symfony\Contracts\HttpClient\Exception\HttpExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * Provides Credentials for running Pod Identity inside an EKS Cluster.
+ *
+ * @see https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html
+ */
+final class PodIdentityProvider implements CredentialProvider
```
jderusse, line 21: What's about reusing the existing `ContainerProvider` instead?

It looks like this is what the official AWS SDK does: https://github.com/aws/aws-sdk-php/blob/master/src/Credentials/EcsCredentialProvider.php

In that way, we could leverage the AuthToken in both full / relative URI 


Endpoint is one of:
- `http://169.254.170.2` . `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`
- pr `AWS_CONTAINER_CREDENTIALS_FULL_URI`

Auth Token is one of:
- `AWS_CONTAINER_AUTHORIZATION_TOKEN`
- content of `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`
- nothing

src/Core/src/Credentials/TokenFileLoader.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\Core\Credentials;
+
+use AsyncAws\Core\Exception\RuntimeException;
+
+final class TokenFileLoader
```
jderusse, line 9: I think this class could be a trait.

src/Core/src/Credentials/ChainProvider.php
```diff
@@ -78,6 +78,7 @@ public static function createDefaultChain(?HttpClientInterface $httpClient = nul
             new WebIdentityProvider($logger, null, $httpClient),
             new IniFileProvider($logger, null, $httpClient),
             new ContainerProvider($httpClient, $logger),
+            new PodIdentityProvider($httpClient, $logger),
             new InstanceProvider($httpClient, $logger),
```
jderusse, line 82: not related to your PR, but I discovered that AWS uses either `EscProvider` or `InstanceProvider` in the default chain:

```
        if (self::shouldUseEcs()) {
            $defaultChain['ecs'] = self::ecsCredentials($config);
        } else {
            $defaultChain['instance'] = self::instanceProfile($config);
        } 
```

I think we could safely do the same :thinking:

src/Core/src/Credentials/PodIdentityProvider.php
```diff
@@ -0,0 +1,89 @@
...
+            $tokenFileContent = $this->tokenFileLoader->getTokenFileContent($tokenFile);
+        } catch (\Exception $e) {
+            $this->logger->warning('"Error reading PodIdentityTokenFile "{tokenFile}.', ['tokenFile' => $tokenFile, 'exception' => $e]);
+
+            return null;
+        }
+
+        // fetch credentials from pod identity agent endpoint
+        try {
+            $response = $this->httpClient->request('GET', $fullUri, ['headers' => ['Authorization' => $tokenFileContent]]);
```
jderusse, line 66: the official AWS SDK checks if the URL is valid, I believe we should do the same:

https://github.com/aws/aws-sdk-php/blob/7412a44da62fd607efbaac4084e69d6621f29de1/src/Credentials/EcsCredentialProvider.php#L231-L249

jderusse (commented):
Thanks for this shiny documented PR :heart_eyes: 

I've just some concerns about introducing a new Provider / Versus reusing the ContainerProvider like the AWS-SDK does :thinking:

src/Core/src/Credentials/PodIdentityProvider.php
kevinrudde: I changed my implementation to reuse the `ContainerProvider` like the AWS SDK does 😄

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -46,15 +51,33 @@ public function __construct(?HttpClientInterface $httpClient = null, ?LoggerInte
...
+        $fullUri = $this->getFullUri($configuration);
+
         // introduces an early exit if the env variable is not set.
-        if (empty($relativeUri)) {
+        if (empty($fullUri)) {
+            return null;
+        }
+
+        $tokenFile = $configuration->get(Configuration::OPTION_POD_IDENTITY_AUTHORIZATION_TOKEN_FILE);
+        if (false === empty($tokenFile)) {
```
jderusse, line 62:
```suggestion
        if (!empty($tokenFile)) {
```

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -46,15 +51,33 @@ public function __construct(?HttpClientInterface $httpClient = null, ?LoggerInte
...
+            try {
+                $tokenFileContent = $this->getTokenFileContent($tokenFile);
+            } catch (\Exception $e) {
+                $this->logger->warning('"Error reading PodIdentityTokenFile "{tokenFile}.', ['tokenFile' => $tokenFile, 'exception' => $e]);
+
+                return null;
+            }
+        }
+
+        if (!$this->isUriValid($fullUri)) {
```
jderusse, line 72: can be move earlier (at line 60 for instance)

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -77,4 +100,78 @@ public function getCredentials(Configuration $configuration): ?Credentials
...
+        }
+
+        // Unknown IP format
+        return false;
+    }
+
+    private function getFullUri(Configuration $configuration): ?string
+    {
+        $relativeUri = $configuration->get(Configuration::OPTION_CONTAINER_CREDENTIALS_RELATIVE_URI);
+        $fullUri = $configuration->get(Configuration::OPTION_POD_IDENTITY_CREDENTIALS_FULL_URI);
```
jderusse, line 138: can be moved at line 144 (no need to instanciate the `fullUri` variable when the `relativeUri` is not null

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -77,4 +100,78 @@ public function getCredentials(Configuration $configuration): ?Credentials
...
     }
+
+    /**
+     * Checks if the provided IP address is a loopback address.
+     *
+     * @param string $host the host address to check
+     *
+     * @return bool true if the IP is a loopback address, false otherwise
+     */
+    public static function isLoopBackAddress(string $host)
```
jderusse, line 111: no need to be static, nor to be public. 
lets avoid exposing ( = maintaining) new methods

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -77,4 +100,78 @@ public function getCredentials(Configuration $configuration): ?Credentials
...
+        }
+
+        if (!isset($parsedUri['scheme'])) {
+            return false;
+        }
+
+        if ('https' !== $parsedUri['scheme']) {
+            $host = trim($parsedUri['host'] ?? '', '[]');
+
+            $ecsHost = parse_url(self::ECS_ENDPOINT)['host'] ?? '';
```
jderusse, line 166: no need to parse the entier url, when you are looking only for the host

```suggestion
            $ecsHost = parse_url(self::ECS_ENDPOINT, \PHP_URL_HOST) ?: '';
```

src/Core/src/Credentials/ContainerProvider.php
```diff
@@ -77,4 +100,78 @@ public function getCredentials(Configuration $configuration): ?Credentials
...
+            $eksHost = self::EKS_HOST_IPV4;
+
+            // check if the host is a valid ecs or eks host or the loopback address
+            if ($host !== $ecsHost && $host !== $eksHost && self::EKS_HOST_IPV6 !== $host && !self::isLoopBackAddress($host)) {
```
jderusse, line 170: feel inconsistent to store ipv4 in a variable, but not for ipv6 :thinking: 

```suggestion
            $eksHost = self::EKS_HOST_IPV4;

            // check if the host is a valid ecs or eks host or the loopback address
            if ($host !== $ecsHost && $host !== self::EKS_HOST_IPV4 && self::EKS_HOST_IPV6 !== $host && !self::isLoopBackAddress($host)) {
```


also, for performance reason, we can re-order logic to check easy computation first

```php
if ('https' !== $parsedUri['scheme']) {
  $host = trim($parsedUri['host'] ?? '', '[]');
  if (self::EKS_HOST_IPV4 === $host || self::EKS_HOST_IPV6 === $host) {
    return true;
  }

  if ($host === parse_url(self::ECS_ENDPOINT, \PHP_URL_HOST)) {
    return true;
  }

  return self::isLoopBackAddress($host);
}
```


also, if we store `ECS_HOST = 169.254.170.2` (instead of endpoint) we could save the `parse_url` thing without compromising too much the readability

jderusse (commented):
looks good, just few minor comment for my side

jderusse (approved):

kevinrudde:
Hey @jderusse  👋 
is there an ETA for merging this? 😄

jderusse:
thank you @kevinrudde

PR #1791, opened 2024-11-09
Upgrade php-cs-fixer to version 3.64

jderusse:


GrahamCampbell (changes requested):
We should not be adding trailing commas in function definitions, unless we want to raise the minimum PHP version.

PR #1793, opened 2024-11-11
Upgrade phpstan

jderusse:


jderusse:
unrelated failure (because of missing entry in the changelog, but the change in `core` is a CS fixing already present in the changelog)

PR #1816, opened 2024-12-12
Remove incorrect change log entry

GrahamCampbell:
This change affects only actions that have not been generated (yet).


jderusse:
good catch. I merged the PR too fast

PR #1830, opened 2025-01-12
Prevent aliasing self class in use statements

jderusse:


src/Service/DynamoDb/src/ValueObject/AttributeValue.php
```diff
@@ -133,8 +131,8 @@ public function __construct(array $input)
         $this->ss = $input['SS'] ?? null;
         $this->ns = $input['NS'] ?? null;
         $this->bs = $input['BS'] ?? null;
-        $this->m = isset($input['M']) ? array_map([AttributeValue1::class, 'create'], $input['M']) : null;
-        $this->l = isset($input['L']) ? array_map([AttributeValue1::class, 'create'], $input['L']) : null;
+        $this->m = isset($input['M']) ? array_map([AttributeValue::class, 'create'], $input['M']) : null;
```
GrahamCampbell, line 134: I'm surprised PHP-CS-Fixer doesn't convert this to `self::`. Does it convert it back, if we used `self::` here?

jderusse: I don't think it converts back.

PR #1831, opened 2025-01-13
Add support for SSO authentication

jderusse:
Fix #1581

PR #1833, opened 2025-01-15
fix minor bug in logic

Nyholm:
I do believe this was missed in the review for #1831

### DIFF
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## NOT RELEASED
 
+## 1.24.1
+
+### Fixed
+
+- Better detection and error messages for when SSO is used but required packages are not installed.
+
 ## 1.24.0
 
 ### Added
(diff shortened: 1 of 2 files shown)


Nyholm:
Wait, there are more problems here. 

I see we have `getCredentialsFromSsoSession()` and `getCredentialsFromLegacySso()` both are using `async-aws/sso`. But `getCredentialsFromSsoSession()` is also using `async-aws/sso-oidc`. Is that correct?

jderusse:
> Wait, there are more problems here.
> 
> I see we have `getCredentialsFromSsoSession()` and `getCredentialsFromLegacySso()` both are using `async-aws/sso`. But `getCredentialsFromSsoSession()` is also using `async-aws/sso-oidc`. Is that correct?

yes, `sso_session` use both `async-aws/sso` and `async-aws/sso-oidc` could you amend your PR please?

Nyholm:
Thank you for merging

PR #1834, opened 2025-01-15
Add SESv2 suppressed destination deletion and retrieval methods

Seldaek:


src/Service/Ses/tests/Unit/SesClientTest.php
```diff
@@ -15,13 +18,25 @@
...
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
 class SesClientTest extends TestCase
 {
+    public function testGetSuppressedDestination(): void
+    {
+        $client = new SesClient([], new NullProvider(), new MockHttpClient());
+
+        $input = new GetSuppressedDestinationRequest([
+            'EmailAddress' => 'change me',
```
Nyholm, line 32: It is a silly example, but we would prefer "real" values in case people read the tests as an example how to use it. 

```suggestion
            'EmailAddress' => 'test@example.org',
```

Nyholm (approved):
Thank you!

Nyholm (approved):

src/Service/Ses/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Added
+
+- AWS api-change: Added support for `getSuppressedDestination` and `deleteSuppressedDestination` to manage SESv2 suppression list entries.
```
Seldaek, line 7: This doesn't seem to please the CI but I have no idea why not

jderusse, line 7:
```suggestion
- Added support for `getSuppressedDestination` and `deleteSuppressedDestination` to manage SESv2 suppression list entries.
```

the `AWS api-change: ` is used by the bot when it updated automatically the API with the latest AWS specs

jderusse (approved):
Last failure is about changing the version in the `src/Service/Ses/composer.json` file (because it's a new feature that will be released in the next minor version, we need to bump

Seldaek:
Looks all good now

PR #1849, opened 2025-02-28
Create BedrockRuntime client with InvokeModel operation

victor-upmeet:
This PR implements a Bedrock Runtime client with InvokeModel operation. Work is still in progress as tests are not implemented yet. This PR is just to check that I'm going in the right direction, and that the base codes for client, operation and other configurations are correct.


victor-upmeet:
I just finished writing tests and fixed what needed to be fixed, however I have a failing test on a generated code (src/Service/BedrockRuntime/src/Result/InvokeModelResponse.php), because the generated code assigns a null value on a property that should not be null. Because I am not allowed to modify generated code, I don't know what to do to fix this test. Any help would be appreciated. Thanks!

src/Service/BedrockRuntime/src/Result/InvokeModelResponse.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->initialize();
+
+        return $this->performanceConfigLatency;
+    }
+
+    protected function populateResult(Response $response): void
+    {
+        $headers = $response->getHeaders();
+
+        $this->contentType = $headers['content-type'][0] ?? null;
```
jderusse, line 63: this is weird. probably a bug in the generator.
I'll have a look

src/Service/BedrockRuntime/tests/Integration/BedrockRuntimeClientTest.php
```diff
@@ -0,0 +1,43 @@
...
+        $result->resolve();
+
+        // self::assertTODO(expected, $result->getBody());
+        self::assertSame('changeIt', $result->getContentType());
+        self::assertSame('changeIt', $result->getPerformanceConfigLatency());
+    }
+
+    private function getClient(): BedrockRuntimeClient
+    {
+        self::markTestSkipped('There is no docker image available for CodeDeploy.');
```
jderusse, line 37: this is not CodeDeploy

manifest.json
```diff
@@ -64,6 +64,19 @@
                 ]
             }
         },
+        "BedrockRuntime": {
+            "source": "https://raw.githubusercontent.com/aws/aws-sdk-php/${LATEST}/src/data/bedrock-runtime/2023-09-30/api-2.json",
+            "documentation": "https://raw.githubusercontent.com/aws/aws-sdk-php/${LATEST}/src/data/bedrock-runtime/2023-09-30/docs-2.json",
+            "pagination": "https://raw.githubusercontent.com/aws/aws-sdk-php/${LATEST}/src/data/bedrock-runtime/2023-09-30/paginators-1.json",
+            "endpoint-rule-set": "https://raw.githubusercontent.com/aws/aws-sdk-php/${LATEST}/src/data/bedrock-runtime/2023-09-30/endpoint-rule-set-1.json",
```
jderusse, line 71: Should be removed, this is not managed by the generator.
Same for `smoke` and `endpoint-tests`

src/Service/BedrockRuntime/src/Result/InvokeModelResponse.php
jderusse: this should be fixed. could you please rebase an run the `generate` commend?

.idea/.gitignore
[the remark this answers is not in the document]
victor-upmeet: Indeed, this is fixed

docs/clients/bedrock-runtime.md
```diff
@@ -0,0 +1,43 @@
...
+    'temperature' => 1
+];
+
+$result = $bedrockRuntime->invokeModel(new InvokeModelRequest([
+    'modelId' => 'us.anthropic.claude-3-7-sonnet-20250219-v1:0',
+    'contentType' => 'application/json',
+    'body' => json_encode($body, JSON_THROW_ON_ERROR)
+]));
+
+$response = json_decode($result->getBody(), true, 512, JSON_THROW_ON_ERROR);
```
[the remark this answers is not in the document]
victor-upmeet: This API invokes some AI models, and each model has different input specifications. Therefore this API is indeed quite generic, and is just an interface to send back and forth the inputs to send to the models, and the output sent back by the LLM.

src/Service/BedrockRuntime/src/Result/InvokeModelResponse.php
victor-upmeet: It is indeed fixed, thanks!

src/Service/BedrockRuntime/tests/Integration/BedrockRuntimeClientTest.php
victor-upmeet: Oops, fixed

manifest.json
victor-upmeet: I am quite confused here as to what should be removed. [In the documentation](https://async-aws.com/contribute/generate.html), it states:

> Register the service in the services section of the ./manifest.json file.
> Look at the official SDK for resources in ./src/data. ([DynamoDB example](https://github.com/aws/aws-sdk-php/tree/3.133.23/src/data/dynamodb/2012-08-10))
> Add a link to source, and, if the files exists, links to documentation, example, pagination, waiter etc.
> Leave the methods key empty. ("methods": [])
> Provide the base URL to the AWS API reference (use [AWS](https://docs.aws.amazon.com/) or [Google](https://www.google.com/search?q=dynamodb+api+reference))

So, I added everything that exist. What attributes of the `BedrockRuntime` of the manifest should be kept, and what should be removed?

jderusse: keep all but `smoke` `endpoint-`

I need to clarify the documentation...

victor-upmeet: Ok done, kept all but `smoke` `endpoint-`

victor-upmeet:
@jderusse Do you need me to do anything else on the PR for it to be merged?

PR #1861, opened 2025-03-20
Adding Converse method to BedrockRuntime client

victor-upmeet:
Hi there

This PR implements the Converse method to the BedrockRuntime client. However, I am facing some issues with the generated code that makes the tests to fail.

In the static analysis, tests fail because constructors are not implemented on some classes, but these same classes are instanciated with arguments. There are also some failures due to wrong documentation types.

In the CI tests, I have a "Undefined index: role" error originating inside the ConverseResponse class in the generated code.

If I can do anything to help let met know, but as far as I understand, the issue seems to be cause by the code generator.


stof:
Looks like the CodeGenerator has an inconsistency in its handling of objects without properties.
However, having such object in the response does not make sense to me. Either the API definition of AWS is wrong, or we don't handle it properly.

stof:
I discovered that `DocumentSource` uses `union: true` as well: https://github.com/aws/aws-sdk-php/blob/33f6d393834053b11a5fc10573a211d92e61c2db/src/data/bedrock-runtime/2023-09-30/api-2.json#L1013-L1021

This also seems something new in those API declarations.

stof:
I opened https://github.com/async-aws/aws/issues/1882 to track the support for document shapes in the code generator

src/Service/BedrockRuntime/tests/Unit/Result/ConverseResponseTest.php
```diff
@@ -0,0 +1,310 @@
...
+        // see https://docs.aws.amazon.com/bedrock/latest/APIReference/API_Converse.html
+        $response = new SimpleMockedResponse('{
+   "metrics":{
+      "latencyMs":100
+   },
+   "output":{
+      "message":{
+         "content":[
+            {
+               "role":"user",
```
stof, line 25: this `"role": "user"` should be a key in the Message object, not in the ContentBlock object (this is why the tests are failing)

src/Service/BedrockRuntime/tests/Unit/Result/ConverseResponseTest.php
```diff
@@ -0,0 +1,310 @@
...
+               "content":{
```
stof, line 26: This looks suspicious to me. https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Message.html says that `content` is an array of content blocks. Those content blocks don't have an additional `content` key

stof (changes requested):

PR #1879, opened 2025-04-17
Fix the lowest bound for the symfony/polyfill-uuid requirement

stof:
The 1.13.0 release (the first release of that package) is unusable due to a bug triggering an error when loading the autoload file.
Defining 1.13.1 as our lowest bound for the requirement (the first working release) makes it easier for projects testing with `--prefer-lowest`.


stof:
See https://github.com/symfony/symfony/pull/60138#discussion_r2037419871 for the original discussion reporting the issue

jderusse:
thank you @stof

Issue #1921, opened 2025-07-22
Better handling of unknown enum values in responses


stof:
Currently, our generator code says that our getter returns `FooEnum::*`, but the code parsing results assigns the string value returned by the backend without checking whether it is a known value. This means that the getter can actually return any string in case AWS adds new values in those endpoints.
Psalm actually detects this mismatch (that's 90% of the content of our baseline). phpstan does not report it because it is a rule checked at level 7 while we run it only at level 6 currently (when trying to enable level 7, it also detects it).

The AWS guidelines for their SDK require that SDK don't fail at runtime when receiving such values. Most official SDKs (maybe even all) implement this by generating an extra case in the enum (`sdkUnknown` in Kotlin, `$unknown` in JS, `UNKNOWN_TO_SDK` in Java, etc...) that is used by parsers when populating results if the case is unknown.
This approach forces projects to decide how to handle that case when they do a `match` on the enum (maybe they decide to fail at runtime with an exception asking to upgrade the SDK, maybe they have a graceful fallback, depending on what is appropriate).

jderusse:
IMHO, for PHP returning `UNKNOWN` does not provide any value than returning the original string.
At the end, users have to check the value.

I get that for strictly typed programming languages, they need to convert the original value to a well-known fallback, but it is not the case for PHP.

Adding a check on the value, to provide a fallback, would consume CPU for nothing in 99.99% of case.

I think we should do what javascript sdk does and adding a `|string`. on the related DocBlocs

https://github.com/aws/aws-sdk-js/blob/657d6feb00447c8be1d65158a0ecc0585b70ed60/clients/s3.d.ts#L1133

This is what I'm doing in #1923

Nyholm:
Very interesting question. 

It seams like the promise AWS is doing is to return a value for X, Y and Z. But they reserve the ability to add more values in the future. I dont think we should make a promise that is more strict nor should we let everything else fallback to a hard coded "unknown". 

I think https://github.com/async-aws/aws/pull/1923 is the best option. `RuntimeName::*|string`. 

Yes, it is not great for type safety but that is what the API promise. Doing a type hint of `RuntimeName::*|string` would also tell the user handle any string input which is exactly what we would like.

stof:
Indeed, their JS SDK is inconsistent:

[...]

PR #1923, opened 2025-07-24
Fix typehint on enums to reflect that AWS might add new values

jderusse:


GrahamCampbell:
Hmmm, but the type `RuntimeName::*|string` is identical to `string`, so we lose the value of having the specific string values.

jderusse:
> Hmmm, but the type `RuntimeName::*|string` is identical to `string`, so we lose the value of having the specific string values.

Yeah, that's true. See the reasoning here https://github.com/async-aws/aws/issues/1921#issuecomment-3112861068

stof:
Based on the analysis of the official JS SDK (see https://github.com/async-aws/aws/issues/1921#issuecomment-3200158314), I'm not sure we should do this change.
Our current approach (returning the string as is but documenting the type as being one of the known values) is exactly what AWS does in their own SDK for JS and Typescript users.

PR #1931, opened 2025-08-10
do not call setAccessible() on PHP 8.1+

xabbuh:
This operation has no effect since PHP 8.1 and the method is deprecated in PHP 8.5.


.github/workflows/ci.yml
```diff
@@ -58,7 +58,7 @@ jobs:
             sed -i -re 's/"require": \{/"repositories": [{"type": "path","url": "..\/..\/Core", "options": {"versions": {"async-aws\/core": "dev-master"}}, "canonical": ${{matrix.strategy != 'lowest' && 'true' || 'false'}} }],"require": \{/' composer.json
             composer config minimum-stability dev
             composer config prefer-stable true
-            composer require --dev --no-update symfony/phpunit-bridge
+            composer require --dev --no-update symfony/phpunit-bridge:^7.3.2
```
xabbuh, line 61: we need symfony/symfony#61304

.github/workflows/ci.yml
```diff
@@ -93,7 +93,7 @@ jobs:
             cd "$CURRENT_DIR/$COMPONENT"
 
             localExit=0
-            composer require symfony/phpunit-bridge --no-update --no-install
+            composer require symfony/phpunit-bridge:^7.3.2@dev --dev --no-update --no-install
```
jderusse, line 96: can we remove the `@dev` here ?

xabbuh: only once 7.3.2 has been tagged, the minimum stability setting to `dev` further above is not applied to all `composer.json` files (see https://github.com/xabbuh/aws/actions/runs/16861959545/job/47763456579#step:8:32)

PR #1938, opened 2025-09-02
fix query params sorting to build correct canonical query string

DmytroVyshnevsky1:
Hello!

I found a bug while working with Async AWS , AWS sdk have the same one.

If one of the query params is an array with 10+ elements it causes an error:
"The request signature we calculated does not match the signature you provided".
The root of problem is sorting of query params for canonical string, cause params should be sorted by name AFTER url encoding.
You can read about this [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html)

quote:
" You must also sort the parameters in the canonical query string alphabetically by key name. The sorting occurs after encoding."
This [ksort](https://github.com/async-aws/aws/blob/cd16cff578cf6ff1fcfe79ef4c1424a175e9774f/src/Core/src/Signer/SignerV4.php#L324) is the reason of problem.

Example URL:
```
'https://example.com/service?transaction_ids[0]=1111&transaction_ids[1]=1111&transaction_ids[10]=1111&transaction_ids[2]=1111&transaction_ids[3]=1111&transaction_ids[4]=1111&transaction_ids[5]=1111&transaction_ids[6]=1111&transaction_ids[7]=1111&transaction_ids[8]=1111&transaction_ids[9]=1111'

```

The example of sorting result on current version:
```
array(11) {
  ["transaction_ids[0]"]=>
  string(4) "1111"
  ["transaction_ids[10]"]=>
  string(4) "1111"
  ["transaction_ids[1]"]=>
  string(4) "1111"
  ["transaction_ids[2]"]=>
  string(4) "1111"
  ["transaction_ids[3]"]=>
  string(4) "1111"
  ["transaction_ids[4]"]=>
  string(4) "1111"
  ["transaction_ids[5]"]=>
  string(4) "1111"
  ["transaction_ids[6]"]=>
  string(4) "1111"
  ["transaction_ids[7]"]=>
  string(4) "1111"
  ["transaction_ids[8]"]=>
  string(4) "1111"
  ["transaction_ids[9]"]=>
  string(4) "1111"
}
```

and the correct one after this fix:
```
array(11) {
  ["transaction_ids[0]"]=>
  string(7) "1111"
  ["transaction_ids[1]"]=>
  string(7) "1111"
  ["transaction_ids[10]"]=>
  string(7) "1111"
  ["transaction_ids[2]"]=>
  string(7) "1111"
  ["transaction_ids[3]"]=>
  string(7) "1111"
  ["transaction_ids[4]"]=>
  string(7) "1111"
  ["transaction_ids[5]"]=>
  string(7) "1111"
  ["transaction_ids[6]"]=>
  string(7) "1111"
  ["transaction_ids[7]"]=>
  string(7) "1111"
  ["transaction_ids[8]"]=>
  string(7) "1111"
  ["transaction_ids[9]"]=>
  string(7) "1111"
}
```

As you can see current sorting put 10 and 1 in wrong places. This not allowed signer to build correct canonical query string.
Also need to clarify that Query parameters are added to the request via:
```
AsyncAws\Core\Request::setQueryAttribute
```

### DIFF
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Fixed
+- SignerV4: sort canonical query parameters **after** URI-encoding (per AWS SigV4). Fixes incorrect ordering for array-style keys (e.g., `transaction_ids[10]` vs `transaction_ids[1]`), which could cause `The request signature we calculated does not match the signature you provided`. See: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
+
+
 ## 1.27.0
 
 ### Added
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## NOT RELEASED
 
 ### Fixed
-- SignerV4: sort canonical query parameters **after** URI-encoding (per AWS SigV4). Fixes incorrect ordering for array-style keys (e.g., `transaction_ids[10]` vs `transaction_ids[1]`), which could cause `The request signature we calculated does not match the signature you provided`. See: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html
 
+- SignerV4: fix sort of query parameters to build correct canoncal query string 
 
 ## 1.27.0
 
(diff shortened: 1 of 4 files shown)


jderusse (commented):
thanks for the investigation end the psycho. 
could you please provide a test for this ?

jderusse:
> @jderusse Thanks for the quick review — I can add tests. Also, could you clarify what you meant by “end the psycho”? Was that a typo?

uh.. my keyboard replaced pull reaquest by psycho 😅

DmytroVyshnevsky1:
@jderusse 
I added the tests, 
I think they also will help if someone will try to add proper parsing for query string with arrays

jderusse:
thank you @DmytroVyshnevsky1

PR #1941, opened 2025-09-08
bump checkout action to v5

jderusse:


derrabus (approved):
You could configure dependabot to do that for you. 🙂

PR #1957, opened 2025-10-10
Support S3::PutPublicAccessBlock

shadowhand:


stof:
how are you running PHPUnit ? Have you installed latest dependencies in your local setup ? And is this based on an uptodate state of the master branch ?

shadowhand:
@stof `make test`. Yes, I have run `composer install`. Yes, this is from `master` branch.

stof:
In case you have your local setup since quite some times, run `composer update` instead of `composer install` (`composer install` would reinstall what you have in your local composer.lock file from the last time you updated vendors)

stof:
@shadowhand the issue with `make test` is now fixed in the latest master branch.

src/Service/S3/tests/Unit/Input/PutPublicAccessBlockRequestTest.php
```diff
@@ -0,0 +1,38 @@
...
+class PutPublicAccessBlockRequestTest extends TestCase
+{
+    public function testRequest(): void
+    {
+        self::fail('Not implemented');
+
+        $input = new PutPublicAccessBlockRequest([
+            'Bucket' => 'change me',
+            'ContentMD5' => 'change me',
+            'ChecksumAlgorithm' => 'change me',
```
stof: all those `change me` values should be updated to be valid values for those field in the test (as hinted by the string content in the generated test skeleton)

src/Service/S3/tests/Unit/Input/PutPublicAccessBlockRequestTest.php
```diff
@@ -0,0 +1,38 @@
...
+            ]),
+            'ExpectedBucketOwner' => 'change me',
+        ]);
+
+        // see https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutPublicAccessBlock.html
+        $expected = '
+            PUT / HTTP/1.0
+            Content-Type: application/xml
+
+            <change>it</change>
```
stof: this `<change>it</change>` should also be replaced by the actual expected content.

src/Service/S3/tests/Unit/Input/PutPublicAccessBlockRequestTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+use AsyncAws\Core\Test\TestCase;
+use AsyncAws\S3\Input\PutPublicAccessBlockRequest;
+use AsyncAws\S3\ValueObject\PublicAccessBlockConfiguration;
+
+class PutPublicAccessBlockRequestTest extends TestCase
+{
+    public function testRequest(): void
+    {
+        self::fail('Not implemented');
```
stof: once the test is properly implemented, this line should be removed.

src/Service/S3/tests/Integration/S3ClientTest.php
```diff
@@ -918,6 +920,27 @@ public function testPutObjectTagging(): void
...
     }
 
+    public function testPutPublicAccessBlock(): void
+    {
+        $client = $this->getClient();
+
+        $input = new PutPublicAccessBlockRequest([
+            'Bucket' => 'change me',
+            'ContentMD5' => 'change me',
+            'ChecksumAlgorithm' => 'change me',
```
stof: all those `change me` values should be updated to be valid values for those field in the test (as hinted by the string content in the generated test skeleton)

src/Service/S3/composer.json
```diff
@@ -16,7 +16,7 @@
         "ext-filter": "*",
         "ext-hash": "*",
         "ext-simplexml": "*",
-        "async-aws/core": "^1.22"
+        "async-aws/core": "^1.9"
```
stof: this change is wrong. This requirement is managed by the code generator. Please make sure that you ran the latest version of the code generator (rather than running an old version and then rebasing the generated code)

src/Service/S3/src/Input/PutPublicAccessBlockRequest.php
```diff
@@ -0,0 +1,224 @@
...
+            throw new InvalidArgument(\sprintf('Missing parameter "PublicAccessBlockConfiguration" for "%s". The value cannot be null.', __CLASS__));
+        }
+
+        $node->appendChild($child = $document->createElement('PublicAccessBlockConfiguration'));
+        $child->setAttribute('xmlns', 'http://s3.amazonaws.com/doc/2006-03-01/');
+        $v->requestBody($child, $document);
+
+        if (null === $this->contentMd5) {
+            $this->contentMd5 = \hash('md5', $document->saveXML($child));
+        }
```
shadowhand: @stof I assume that this functionality should be included, to mimic the official SDK behavior?

shadowhand: It looks like the one failing check is because of this change. It seems that similar parameters are _not_ automatically resolved by async-aws, so I will remove it.

src/Service/S3/composer.json
shadowhand: FYI, the code generator is what changed this. I have not modified the generator in any way.

stof: I think this might happen when generating a single operation for a service instead of regenerating all operations (when the need for a higher version is because of another operation)

src/Service/S3/src/Input/PutPublicAccessBlockRequest.php
stof: Resolving that automatically would have to be implemented in the code generator;
If you want such behavior, I suggest opening a dedicated issue to discuss it, allowing to figure out how this should be managed.

PR #1959, opened 2025-10-17
Fix the local test setup

stof:
When installing dependencies in the root project, the bundle tests are installing phpunit as a transitive dependency. This causes issue when the installed version is incompatible with the version installed by the simple-phpunit script.
CI was not impacted because it injects such conflict rule in component-level files to prevent this issue already.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,7 @@
         "symfony/polyfill-uuid": "^1.13.1"
     },
     "conflict": {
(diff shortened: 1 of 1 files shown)


stof:
I'll work separately on migrating away from `simple-phpunit`. Letting composer manage the PHPUnit installation as a dev dependency might make things a lot simpler.
This smaller step is meant to improve contributor experience (see #1957)

PR #1965, opened 2025-10-25
Fix the version update for BC breaking changes

stof:
The refresh script was computing the level variable based on the kind of changes in the changelog of the unreleased version, and was later reinitializing the variable, losing that info.

This fixes the issue detected in https://github.com/async-aws/aws/pull/1964

### DIFF
--- a/.github/workflows/bin/refresh
+++ b/.github/workflows/bin/refresh
@@ -265,7 +265,6 @@ foreach ($services as $service => $hasChange) {
             exit(1);
         }
         $parts = explode('.', $lastPackageVersion);
-        $level = 1;
         if (0 === (int) $parts[0]) {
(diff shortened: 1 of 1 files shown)


.github/workflows/bin/refresh
```diff
@@ -265,7 +265,6 @@ foreach ($services as $service => $hasChange) {
             exit(1);
         }
         $parts = explode('.', $lastPackageVersion);
-        $level = 1;
```
stof, line 268: this variable is initialized as `1` on line 242. Then, it is updated on line 246 to a value of `0` in case we have changes requiring a major version. But this was lost due to this line.

jderusse (approved):

PR #1967, opened 2025-10-25
Drop support for PHP lower than 8.2

### LINKED ISSUE #1961: [RFC] Bump the minimum version of PHP we support
We currently support PHP 7.2.5+. This support range forces us to keep support for unmaintained dependencies:

- Symfony 6.4 requires PHP 8.1+ (and Symfony 7 requires PHP 8.2+). Using older PHP versions require using Symfony 5.4 which does not receive bug fixes anymore (it will still receive security fixes for a few years thanks to the extended LTS that has been sponsored by Ibexa instead of reaching EOL at the end of november)
- PHPUnit 9.6 has to be used to run tests instead of being able to migrate to PHPUnit 11 or 12 (we probably won't migrate to PHPUnit 10 at all, as we want deprecation reporting. the `symfony/phpunit-bridge` implements it for PHPUnit 9.x and PHPUnit 11 provides an equivalent feature set natively, while PHPUnit 10 has a less powerful feature). PHPunit 9.x is on life support, where it only receives patches to solve deprecations reported in new PHP versions (until the day where it becomes too complex to implement such patches)

stof:
Dropping support for old PHP versions simplifies our maintenance work, by unlocking a bunch of improvements (which will be done in separate PRs):

- updating Psalm to a maintained versions
- updating to PHPUnit 11
- investigating the usage of newer PHP features (typed properties, readonly properties, etc...) when doable without BC breaks.

Note that this also fixes the correctness of the implementation of #1882, which relies on union type that are not compatible with PHP 7 (not detected by the CI as we don't yet generate code needing such shape, but there is an in-progress PR that will use it once updated).

Closes #1961

### DIFF
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -1,47 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.11.0@c9b192ab8400fdaf04b2b13d110575adc879aa90">
+<files psalm-version="5.14.1@b9d355e0829c397b9b3b47d0c0ed042a8a70284d">
   <file src="src/Core/src/AbstractApi.php">
     <UndefinedFunction>
       <code><![CDATA[trigger_deprecation('async-aws/core', '1.2', 'Extending "%s"" without overriding "%s" is deprecated. This method will be abstract in version 2.0.', __CLASS__, __FUNCTION__)]]></code>
       <code><![CDATA[trigger_deprecation('async-aws/core', '1.2', 'providing an endpoint with placeholder is deprecated and will be ignored in version 2.0. Provide full endpoint instead.')]]></code>
     </UndefinedFunction>
-  </file>
-  <file src="src/Core/src/Credentials/PsrCacheProvider.php">
-    <InvalidCatch>
-      <code><![CDATA[try {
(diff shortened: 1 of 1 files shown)


stof:
Remaining TODO; figuring out the proper way to document this in the changelog files so that we release it as a semver-minor, not as a semver-patch.

jderusse (approved):

stof:
The changes in the psalm baseline correspond to different kinds:
- updating the psalm version on the root node (apparently, it's been a while since the last time we did an automated generation of the baseline as it was with an older version of Psalm)
- reordering files in the baseline (manual edition of the baseline did not preserve alphabetical sorting)
- removing some errors in `Core` that are no longer reported thanks to more precise type declarations in `psr/cache` and `symfony/cache`, thanks to using newer versions now that Psalm does not run on PHP 7 anymore.
- new errors reported in the `Monolog` integration due to running Psalm using Monolog 3.x instead of 2.x (fixing them will requiring dropping support for Monolog 2.x, to migrate to modern Monolog APIs)

stof:
@jderusse @Nyholm could one of you remove all PHP versions lower than 8.2 from the list of required CI checks ? I don't have admin access on the repo to do it myself.

PR #1973, opened 2025-10-26
Upgrade Psalm to 6.13

stof:
Now that we don't need to support PHP 7 anymore, we can upgrade Psalm to version 6.

I disabled some rules in config, especially given Psalm is now stricter about the types of SimpleXMLElement (which returns `null` for properties when accessing them on an element that does not exist in its parent)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "nikic/php-parser": "^4.0",
         "nyholm/symfony-bundle-test": "^3.0",
         "phpstan/phpstan": "~2.0.0",
-        "psalm/phar": "~5.14.1",
+        "psalm/phar": "~6.13.1",
         "psr/cache": "^1.0 || ^2.0 || ^3.0",
         "swaggest/json-diff": "^3.7",
         "symfony/cache": "^4.4 || ^5.0 || ^6.0 || ^7.0 || ^8.0",
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -1,14 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.14.1@b9d355e0829c397b9b3b47d0c0ed042a8a70284d">
-  <file src="src/Core/src/AbstractApi.php">
-    <UndefinedFunction>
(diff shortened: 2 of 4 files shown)


src/Core/src/Credentials/IniFileLoader.php
```diff
@@ -62,6 +62,7 @@ public function loadProfiles(array $filepaths): array
 
             foreach ($this->parseIniFile($filepath) as $name => $profile) {
                 $name = preg_replace('/^profile /', '', (string) $name);
+                \assert($name !== null);
```
jderusse: how is it possible to get null here ?

src/Service/S3/src/S3Client.php
```diff
@@ -3313,7 +3313,10 @@ protected function getEndpoint(string $uri, array $query, ?string $region): stri
             return parent::getEndpoint($uri, $query, $region);
         }
 
-        return preg_replace('|https?://|', '${0}' . $bucket . '.', parent::getEndpoint('/' . $uriWithOutBucket, $query, $region));
+        $endpoint = preg_replace('|https?://|', '${0}' . $bucket . '.', parent::getEndpoint('/' . $uriWithOutBucket, $query, $region));
+        \assert($endpoint !== null); // We assume that the regex does not fail as we know it is valid.
```
jderusse: same

src/Core/src/Credentials/IniFileLoader.php
stof: `preg_replace` returns `null` on failure (when the regex is invalid, or when the PCRE JIT reaches the backtracking limit or when using a `u` modifier in the regex with an input that is not UTF-8). This is because the `preg_*` functions use warnings instead of exceptions for error reporting (and so they still return something on error)
As this regex is known to be valid, does not use the `u` modifier and does not require any backtracking, I can know for sure it won't fail in a way meaningful way, so I used such assert to handle that case (which will be a no-op when using a prod config of PHP)

stof: Note that phpstan would also complain about those at level 7 (for cases returning `false`) or 8 (for cases using `null`). But our current config runs at level 6 only.

PR #1977, opened 2025-10-28
Migrate the testsuite to PHPUnit 11

stof:
This migrates the testsuite to use PHPUnit 11.

I'm now installing phpunit through composer rather than using `simple-phpunit` as it is much simpler: `matthiasnoback/symfony-config-test` already brings us a transitive dependency anyway (for which we needed to add a conflict rule to prevent issues).

I'm still registering `Symfony\Bridge\PhpUnit\SymfonyExtension` from the symfony phpunit bridge in our config. The feature I wanted is the registration of the DebugClassLoader of `symfony/error-handler` to report deprecations detecting by the DebugClassLoader (which was previously done for root tests where error-handler is installed by FrameworkBundle but not for component tests where it was not installed). It might also be useful if we want to use the ClockMock or DnsMock features in the future.

### DIFF
--- a/src/Integration/Laravel/Queue/phpunit.xml.dist
+++ b/src/Integration/Laravel/Queue/phpunit.xml.dist
@@ -1,17 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/11.5/phpunit.xsd"
          backupGlobals="false"
          colors="true"
          bootstrap="vendor/autoload.php"
          failOnRisky="true"
          failOnWarning="true"
+         failOnDeprecation="true"
+         failOnNotice="true"
 >
-  <coverage>
+  <source ignoreSuppressionOfDeprecations="true" ignoreIndirectDeprecations="true">
     <include>
       <directory>./src</directory>
     </include>
-  </coverage>
+    <deprecationTrigger>
+      <function>trigger_deprecation</function>
+    </deprecationTrigger>
+  </source>
   <php>
     <ini name="error_reporting" value="-1"/>
   </php>
@@ -20,4 +25,7 @@
       <directory>./tests/</directory>
     </testsuite>
   </testsuites>
+  <extensions>
+    <bootstrap class="Symfony\Bridge\PhpUnit\SymfonyExtension"/>
+  </extensions>
 </phpunit>
--- a/src/Integration/Symfony/Bundle/composer.json
+++ b/src/Integration/Symfony/Bundle/composer.json
@@ -23,9 +23,14 @@
         "async-aws/ses": "^1.0",
         "async-aws/sqs": "^1.0 || ^2.0",
         "async-aws/ssm": "^1.0 || ^2.0",
-        "matthiasnoback/symfony-config-test": "^4.1 || ^5.0",
+        "matthiasnoback/symfony-config-test": "^6.1",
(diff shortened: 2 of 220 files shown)


src/Integration/Symfony/Bundle/composer.json
```diff
@@ -23,9 +23,14 @@
...
         "async-aws/sqs": "^1.0 || ^2.0",
         "async-aws/ssm": "^1.0 || ^2.0",
-        "matthiasnoback/symfony-config-test": "^4.1 || ^5.0",
+        "matthiasnoback/symfony-config-test": "^6.1",
         "nyholm/symfony-bundle-test": "^3.0",
-        "symfony/cache": "^4.4 || ^5.0 || ^6.0 || ^7.0 || ^8.0"
+        "phpunit/phpunit": "^11.5.42",
+        "symfony/cache": "^4.4 || ^5.0 || ^6.0 || ^7.0 || ^8.0",
+        "symfony/error-handler": "^7.3.2 || ^8.0",
+        "symfony/framework-bundle": "^5.4.45 || ^6.4.13 || ^7.1.6",
```
stof, line 31: This requirement on FrameworkBundle is meant to fix our `--prefer-lowest` CI job by restrict the lowest bound (for the dependency coming from `nyholm/symfony-bundle-test`) so that we have a version including the change in https://github.com/symfony/framework-bundle/commit/1c11e27c68c3dd576edf93dcb7a36087b80a50df

this change combined with the fact that I install `symfony/runtime` solves the issue that booting the bundle would attempt at replacing the error handler, causing issues with PHPUnit 11 (this is why I don't let the composer plugin run for the runtime component: I won't care about it).

src/Integration/Laravel/Queue/phpunit.xml.dist
```diff
@@ -1,17 +1,22 @@
...
          backupGlobals="false"
          colors="true"
          bootstrap="vendor/autoload.php"
          failOnRisky="true"
          failOnWarning="true"
+         failOnDeprecation="true"
+         failOnNotice="true"
 >
-  <coverage>
+  <source ignoreSuppressionOfDeprecations="true" ignoreIndirectDeprecations="true">
```
stof, line 12: this package gets indirect deprecations (from `illuminate/container`) when using the DebugClassLoader. I'm ignoring them here as I still want the DebugClassLoader for our own code.

jderusse:
> The remaining failures for root tests are because our changelog tests currently don't allow changing the composer.json without changelog entries. But here, I'm only changing dev requirements. @jderusse should we add the `composer.json` to the exclusion list ?

I'm mixed filling... the changelog is used by the `release` script to guess the next version.

So components with only change in composer will not be released. This is fine now because the changelog has a pending entry for php 8.2.
But I wonder if we could have case where we bump a dependency (and only update composer.json) that would require a new release 🤔 (ie. add a conflict for an incompatible version of a dependency, or replace an unmaintained dep by its forks, ...)
Those are really edge cases, but the test is here to avoid forgetting this.

I'm fine with merging this PR with red test, but I think we should not change the test.

stof:
> I'm fine with merging this PR with red test, but I think we should not change the test.

This would require disabling the branch protection rule though (unless you have permission to bypass it, which I don't).
But this is indeed an option.

jderusse:
> > I'm fine with merging this PR with red test, but I think we should not change the test.
> 
> This would require disabling the branch protection rule though (unless you have permission to bypass it, which I don't). But this is indeed an option.

I have permission to do it. 
Shall we wait for https://github.com/symfony/symfony/pull/62196 to be merged

stof:
> Shall we wait for [symfony/symfony#62196](https://github.com/symfony/symfony/pull/62196) to be merged

it would make sense, yes. Otherwise the job will fail for all PRs. But I expect it to be merged soon.

jderusse:
thank you @stof

PR #1983, opened 2025-11-07
Skip changelog's test when the dataprovider returns nothing

jderusse:
Fix failling test when the DataProviders returns an empty list.

https://github.com/async-aws/aws/actions/runs/19167154579/job/54790443661#step:8:52

### DIFF
--- a/tests/Unit/ChangelogTest.php
+++ b/tests/Unit/ChangelogTest.php
@@ -131,8 +131,12 @@ class ChangelogTest extends TestCase
     }
 
     #[DataProvider('provideChangedServicesWithoutChangelog')]
-    public function testChangelogEntryForService(string $service, string $base, $isCommentOnly)
+    public function testChangelogEntryForService(string $service, string $base, bool $isCommentOnly)
     {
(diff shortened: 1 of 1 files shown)


tests/Unit/ChangelogTest.php
```diff
@@ -186,7 +190,7 @@ static function ($branch) {
         );
 
         if (0 === \count($branches)) {
-            self::markTestSkipped('Cannot find the master branch');
+            yield [null, null, null];
```
stof, line 193: why yielding null which will trigger a type error ?

Also, `markTestSkipped` was throwing, which was stopping the data provider.
Now, such case would trigger a notice when accessing `$branches[0]`

jderusse: yielding null is a mistake and should be fixed by #1985 

Throwing an exception in the dataprovider breaks the test suite => https://github.com/async-aws/aws/actions/runs/19174431727/job/54815354423#step:8:11

yielding a dummy value is the single way I found to prevent PhpUnit from throwing an error

PR #1984, opened 2025-11-07
Fix badges in the readme of subtree packages

stof:
when I fixed the URL of the GHA badges in the main readme in https://github.com/async-aws/aws/pull/1970, I missed the fact that subtree repos also have such badges in their readme.


jderusse:
thank you

PR #1985, opened 2025-11-07
Fix provider when branch is not master

jderusse:


tests/Unit/ChangelogTest.php
```diff
@@ -188,9 +188,9 @@ static function ($branch) {
                 return 'master' === $branch || '/master' === substr($branch, -7);
             }
         );
-
+        self::markTestSkipped('Cannot find the master branch');
```
jderusse: temporary test. to remove

tests/Unit/ChangelogTest.php
```diff
@@ -190,7 +190,7 @@ static function ($branch) {
         );
 
         if (0 === \count($branches)) {
-            yield [null, null, null];
+            yield ['', '', false];
```
stof, line 193: this also needs a `return;` to end the data provider in such case

jderusse: of course... :facepalm:

PR #1998, opened 2025-11-27
Add UNKNOWN_TO_SDK value to enums

### LINKED ISSUE #1921: Better handling of unknown enum values in responses

jderusse:
This PR asserts that enums returned by the API are wellknown values. Otherwise it fallback to the `UNKNOWN_TO_SDK` generic value.

In that way, we respect the contract when saying methods returns a `FooEnum::*`.

Side note: We still have to keep the psalm exceptions, because psalm can not infer the type of `Enum::exists($value) ? $value : Enum:FALLBACK` to `Enum::*`.

see https://psalm.dev/r/518a0924fe

We would need an equivalent of typescript's `is` keyword https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

fix #1921


stof:
The static analysis can actually be fixed by using `@psalm-assert-if-true`: https://psalm.dev/r/b90672fbda

jderusse:
Diff is huge.. Here is a summary:

1. added `public const UNKNOWN_TO_SDK = 'UNKNOWN_TO_SDK';` in all Enums
2. added `@psalm-assert-if-true self::* $value` on all `Enum`'s `static function exists` methods
3. in resutls: replaced `$value` by  `!TheEnum::exists($value) ? TheEnum::UNKNOWN_TO_SDK : $value`.
4. in input: added `/** @psalm-suppress NoValue */` after `if (!TheEnum::exists($v)) { throw ... }` because I want to keep the check even if the docbloc says `$v is a TheEnum::*` and I'm not sure we should relax the docblock to `TheEnum::*|string`

src/Service/RdsDataService/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Added
+
+- Add `UNKNOWN_TO_SDK` value to enums that is used when the API returns an value that is not (or not yet) knwon by the AsyncAws
```
stof: typo in those changelog entry: `knwon` vs `known`

src/Service/AppSync/src/Enum/ConflictDetectionType.php
```diff
@@ -4,9 +4,13 @@
 
 final class ConflictDetectionType
 {
+    public const UNKNOWN_TO_SDK = 'UNKNOWN_TO_SDK';
```
stof: I would suggest to define it as the last constant instead of the first one.

stof (approved):

PR #2000, opened 2025-12-01
Fix camel case format for ignored case

jderusse:
From AWS spec's properties we tried to normalize these name to camelCase.

For some cases, (when the name contains multiple consecutive UpperCases) where the generator is not able to generate the CamelCase value by itself, we maintain a list of (opinionated) exceptions telling that we shoiuld use the form "first char upper, rest lower". ie `ETag` become `Etag`.

We also maintain a list of "ignored" cases where multiple consecutive Upper cases are fine.
ie `XOffset` stays `XOffset`.

The issue, is: By mistake, we also convert the second list to the form "first char upper, rest lower".

This PR fix that.

This might not be a BC break, because PHP is not strict about case on methods.. But I wonder if it worth it...

### DIFF
--- a/src/CodeGenerator/src/Generator/GeneratorHelper.php
+++ b/src/CodeGenerator/src/Generator/GeneratorHelper.php
@@ -63,28 +63,28 @@ class GeneratorHelper
             'SS' => 'Ss',
         ];
         static $ignored = [
-            'GB' => 'Gb',
-            'GiB' => 'Gib',
-            'BFrame' => 'Bframe',
-            'BReference' => 'Breference',
-            'BBox' => 'Bbox',
-            'VCpu' => 'Vcpu',
-            'IInterval' => 'Iinterval',
-            'IFrame' => 'Iframe',
-            'XCoordinate' => 'Xcoordinate',
-            'YCoordinate' => 'Ycoordinate',
-            'XOffset' => 'Xoffset',
-            'YOffset' => 'Yoffset',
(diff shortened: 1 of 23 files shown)


src/CodeGenerator/src/Generator/GeneratorHelper.php
```diff
@@ -63,28 +63,28 @@ public static function normalizeName(string $propertyName): string
...
+            'YPosition',
         ];
 
         $originalPropertyName = $propertyName;
         $propertyName = strtr($propertyName, $replacements);
 
         if (preg_match('/[A-Z]{2,}/', $propertyName)) {
-            $propertyName = strtr($propertyName, $ignored);
-            if (preg_match('/[A-Z]{2,}/', $propertyName)) {
+            $propertyNameWithoutIgnored = strtr($propertyName, array_fill_keys($ignored, ''));
```
stof: This replacement might miss other cases where we would have another capital letter just before or after an ignored group.
The cause of the bug is that this validation overrides `$propertyName` instead of using a different variable name (as done here with `$propertyNameWithoutIgnored`). The fix is to rename the variable without changing the way its content is generated (and keeping the existing structure in `$ignored`)

jderusse: Refactored:
I kept the $ignored as a list to avoid typo and ease maintenance.
But I used `array_fill_keys($ignored, 'Xx')` to avoid collision with neighbors.

PR #2009, opened 2025-12-12
Respect the order of AWS when choosing an alternative protocol

jderusse:


src/CodeGenerator/src/Definition/ServiceDefinition.php
```diff
@@ -158,9 +158,9 @@ public function getProtocol(): string
         }
 
         $protocols = $this->definition['metadata']['protocols'] ?? [];
-        foreach (self::SUPPORTED_PROTOCOLS as $supportedProtocol) {
-            if (\in_array($supportedProtocol, $protocols, true)) {
-                return $supportedProtocol;
+        foreach ($protocols as $protocol) {
```
stof, line 161: this overrides the `$protocol` variable defined at line 155 while it was still used after that loop.

jderusse: true, but given neither `definition['metadata']['protocol']` nor each member of `definition['metadata']['protocols']` is supported, it will fail later anyway.

stof: but it will fail with the *least preferred* alternative protocol instead of the primary protocol, which would make the error less good IMO.

PR #2015, opened 2025-12-24
Add a command adding a changelog entry in all modified packages

stof:
This command is useful when doing updates in the code generator or the php-cs-fixer configuration, where we know that all changes will have the same cause.

For reference, the changelog entries of #2014 have been generated with this command.

### DIFF
--- a/src/CodeGenerator/src/Command/AddChangeLogCommand.php
+++ b/src/CodeGenerator/src/Command/AddChangeLogCommand.php
@@ -1,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\CodeGenerator\Command;
+
+use AsyncAws\CodeGenerator\ChangeLogUpdater;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\ConsoleOutput;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Process\Process;
+
+/**
+ * @internal
+ */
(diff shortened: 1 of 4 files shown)


stof:
The current implementation is detecting git _modified_ files, i.e. changes that have not been staged yet. It means this command is expected to be called before using `git add`. I'm wondering whether we should instead make it detect _staged_ files (it is a matter of switching the flag used in the `git ls-files` command). This would mean that the command would expect to be run _after_ `git add`, but could be used for use cases of partial commits.

src/CodeGenerator/src/Command/AddChangeLogCommand.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        $errorOutput = $output instanceof ConsoleOutput ? $output->getErrorOutput() : $output;
+
+        if (!\in_array($input->getArgument('type'), self::CHANGELOG_LABELS, true)) {
+            $errorOutput->writeln('Invalid type provided. It must be one of: ' . implode(', ', self::CHANGELOG_LABELS) . '.');
+
+            return 1;
+        }
+
+        $changedFiles = explode("\n", (new Process(['git', 'ls-files', '--modified']))->mustRun()->getOutput());
```
jderusse, line 48: What's about reusing part of the code in the `ChangelogTest`?
using `git diff` and to select service that have changes
excluding service touched the `CHANGELOG` file

stof: My goal is to add a changelog entry for new changes, not for all files that are different from the master branch (which might have *other* changelog entries for previous changes)

stof: the logic identifying the package for a modified file is indeed copied from the changelogtest. And `refresh` has a slightly different logic (as it identifies the `Sts` service rather than `Core` as package to find the remote manifest)

Issue #1900, opened 2025-06-07
Add support for union shapes in the code generator


stof:
Newer AWS services like bedrock-runtime are using union shapes in their API. Contrary to older services (like S3 or MediaConvert) that have a `type` property containing an enum, and then additional properties used depending on the type, union shapes expect exactly 1 of their members to be filled.
In the SDK manifest, those union shapes are defined with `union: true`: https://github.com/aws/aws-sdk-php/blob/33f6d393834053b11a5fc10573a211d92e61c2db/src/data/bedrock-runtime/2023-09-30/api-2.json#L1013-L1021

Looking at official AWS SDKs for other languages (the official PHP SDK does not return typed responses and so cannot be used as comparison), here is how they represent it:

- JS defines a subtype for each member (naming it like `<shapeName>.<memberName>Member`) with that member defined with its type and all other members defined as optional properties with type `never` (which forbids having such properties in the object in typescript) and then uses a union type: https://github.com/aws/aws-sdk-js-v3/blob/b09a8f969ad6b2508e4d3de93089fe4498f57408/clients/client-bedrock-runtime/src/models/models_0.ts#L2351-L2382. They also expose a `Visitor` interface with methods for each type, with a `visit` function calling the right method of the visitor: https://github.com/aws/aws-sdk-js-v3/blob/b09a8f969ad6b2508e4d3de93089fe4498f57408/clients/client-bedrock-runtime/src/models/models_0.ts#L2544-L2569 (note that this Visitor interface seems to imply that adding a new member in the union would be a BC break)
- Go defines an interface for the union shape, with implementations for each member (named `<shapeName>Member<memberName>`) that have a `Value` field containing the value of the member: https://github.com/aws/aws-sdk-go-v2/blob/43445011355b4bb4226fa797c5fc141bbc1b793f/service/bedrockruntime/types/types.go#L127-L137 The serializer and deserializer take care of using the right structure for each kind of members.
- the .NET SDK defines a class with a property for each member, which does not seem to enforce in the type system that only one member is set (maybe their generator does not have support for union shapes and generated it as a normal structure shape): https://github.com/aws/aws-sdk-net/blob/09201e894db71b5d536db6cb0975f361b85f1f52/sdk/src/Services/BedrockRuntime/Generated/Model/ContentBlock.cs
[...]

PR #2017, opened 2025-12-25
feat: S3 Vectors pkg

### LINKED ISSUE #2010: Support S3 Vectors Bucket API
https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-s3vectors-2025-07-15.html

aszenz:
Fixes #2010 

Relates to https://github.com/symfony/ai/issues/1251

### DIFF
--- a/src/Service/S3Vectors/tests/Integration/S3VectorsClientTest.php
+++ b/src/Service/S3Vectors/tests/Integration/S3VectorsClientTest.php
@@ -27,6 +27,8 @@ class S3VectorsClientTest extends TestCase
 {
     public function testCreateIndex(): void
     {
+        self::markTestSkipped('localstack doesn\'t support s3 vectors api yet (see https://github.com/localstack/localstack/issues/13498)');
+
         $client = $this->getClient();
 
         $input = new CreateIndexInput([
@@ -54,6 +56,8 @@ class S3VectorsClientTest extends TestCase
 
     public function testCreateVectorBucket(): void
     {
+        self::markTestSkipped('localstack doesn\'t support s3 vectors api yet (see https://github.com/localstack/localstack/issues/13498)');
+
         $client = $this->getClient();
 
         $input = new CreateVectorBucketInput([
@@ -73,6 +77,8 @@ class S3VectorsClientTest extends TestCase
 
     public function testDeleteIndex(): void
     {
+        self::markTestSkipped('localstack doesn\'t support s3 vectors api yet (see https://github.com/localstack/localstack/issues/13498)');
+
         $client = $this->getClient();
 
         $input = new DeleteIndexInput([
@@ -87,6 +93,8 @@ class S3VectorsClientTest extends TestCase
 
     public function testDeleteVectorBucket(): void
     {
+        self::markTestSkipped('localstack doesn\'t support s3 vectors api yet (see https://github.com/localstack/localstack/issues/13498)');
+
         $client = $this->getClient();
 
         $input = new DeleteVectorBucketInput([
@@ -100,6 +108,8 @@ class S3VectorsClientTest extends TestCase
 
     public function testDeleteVectors(): void
     {
+        self::markTestSkipped('localstack doesn\'t support s3 vectors api yet (see https://github.com/localstack/localstack/issues/13498)');
+
         $client = $this->getClient();
 
         $input = new DeleteVectorsInput([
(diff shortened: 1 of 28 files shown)


aszenz:
I get these errors when adding some operations:

GetVectors/ListVectors/PutVectors/QueryVectors -> `Union shapes are not supported yet by the code generator. The shape "VectorData" is a union shape.`

`UntagResource` -> `ListShape in request part "querystring" is not yet implemented`

manifest.json
```diff
@@ -590,6 +590,30 @@
...
+                "DeleteVectorBucketPolicy",
+                "DeleteVectors",
+                "GetIndex",
+                "GetVectorBucket",
+                "GetVectorBucketPolicy",
+                "ListIndexes",
+                "ListTagsForResource",
+                "ListVectorBuckets",
+                "PutVectorBucketPolicy",
+                "TagResource"
```
jderusse: Do you need all these operations?

The philosophy of this project is to add only what is asked/needed by people, not to add everything.

jderusse:
Thanks for this addition, I have few comments

> `UntagResource` -> `ListShape in request part "querystring" is not yet implemented`

This should be fixed by https://github.com/async-aws/aws/pull/2018

> Union shapes are not supported yet

Yes, we need to fix that. https://github.com/async-aws/aws/issues/1900
In the meantime, you can patch the manifest like @stof did in https://github.com/async-aws/aws/pull/2012/changes.

manifest.json
jderusse: Do you need all these operations?

The philosophy of this project is to add only what is asked/needed by people, not to add everything.

aszenz: Removed some of them related to tagging/policies, for the store implementation i would need create bucket/index functionality

aszenz:
Related to tests:

- Integration -> Had to skip, waiting for https://github.com/localstack/localstack/issues/13498
- Unit -> Used AI to generate mocked request inputs/outputs, let me know if they aren't any good

docs/clients/s3-vectors.md
```diff
@@ -0,0 +1,7 @@
+---
+layout: client
+category: clients
+name: S3Vectors
+package: async-aws/s3-vectors
+---
+
```
stof, line 7: Please write the content of the documentation page.

phpstan.neon.dist
```diff
@@ -12,5 +12,6 @@ parameters:
 
     ignoreErrors:
         - identifier: missingType.iterableValue
+        - identifier: return.unusedType
```
stof: what was reported without this ignore rule ?

aszenz:
```
>  ------ ------------------------------------------------------------------------------------------------------------------------------ 
  Line   Service/S3Vectors/src/Input/QueryVectorsInput.php                                                                             
 ------ ------------------------------------------------------------------------------------------------------------------------------ 
  124    Method AsyncAws\S3Vectors\Input\QueryVectorsInput::getFilter() never returns null so it can be removed from the return type.  
         🪪  return.unusedType                                                                                                         
 ------ ------------------------------------------------------------------------------------------------------------------------------
```

So the method never returns null since it uses `?? []`, but it's still in the return type.

This pkg doesn't seem to allow changing generated code, so i suppressed it.

stof: This looks like a case where we need to improve the code generator instead.

docs/clients/s3-vectors.md
aszenz: Done

phpstan.neon.dist
aszenz: > This looks like a case where we need to improve the code generator instead.

should the code generator skip the null type or not do any null coalescing ?

stof: the getter should not have the `?? []`. I guess our code generator has an issue for document shapes (which are quite new, and are not yet used in any merged clients). The `@var` annotation on the property is also suspicious, with duplicate `null` types.

stof: this should all be fixed by #2021

aszenz: thanks, i rebased on your branch, seems to have fixed it

stof: Please rebase again on latest master as my PR is now merged.

aszenz: done

PR #2018, opened 2025-12-26
Add support for Lists in querystring

jderusse:
This PR adds support for List shapes in querystring.

It uses the same logic as the official AWS SDK: 
- build an object with either values (either string or array)
- pass the object to the Guzzle\PSR7 request
- build a querystring without altering the key (https://github.com/guzzle/psr7/blob/21dc724a0583619cd1652f673303492272778051/src/Query.php#L74-L117)

This PR should fix https://github.com/async-aws/aws/pull/2017#issuecomment-3691631490

### DIFF
--- a/src/CodeGenerator/src/Generator/InputGenerator.php
+++ b/src/CodeGenerator/src/Generator/InputGenerator.php
@@ -351,19 +351,28 @@ class InputGenerator
     {
         $memberShape = $member->getShape();
         if ($memberShape instanceof ListShape) {
-            if ('header' !== $requestPart) {
-                throw new \InvalidArgumentException(\sprintf('ListShape in request part "%s" is not yet implemented', $requestPart));
+            if ('header' === $requestPart) {
+                $bodyCode = '
+                APPLY_HOOK
+                $items = [];
+                foreach (INPUT as $value) {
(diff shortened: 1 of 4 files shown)


src/CodeGenerator/src/Generator/InputGenerator.php
```diff
@@ -351,19 +351,27 @@ private function inputClassRequestPartGetters(StructureMember $member, ClassBuil
...
+                OUTPUT = implode(\',\', $items);
+                ';
+            } elseif ('querystring' === $requestPart) {
+                $bodyCode = '
+                APPLY_HOOK
+                foreach (INPUT as $value) {
+                    VALIDATE_ENUM
+                    OUTPUT[] = VALUE;
+                }
+                ';
```
stof, line 371: this case should add a requirement on the core package to enforce a higher lower bound, to ensure we have the version supporting this (we already have the infrastructure to handle this, and we use it elsewhere)

stof (changes requested):

src/CodeGenerator/src/Generator/InputGenerator.php
jderusse: good catch

PR #2019, opened 2025-12-26
Implement the Union Shape

### LINKED ISSUE #1900: Add support for union shapes in the code generator

jderusse:
fix #1900

This PR add support for union shape.

- Use abstract class that return `null` for all membres
- For each member, add a TheParentMemberTheChild class
- Child class extends provide a single not null property (parent provides the rest of the props) 
- Populate methods returns the right child and throws exception when the child is unknown (all members are null)
- `static create` methods return the right child and throws exception if the child is unknown


stof:
> * `static create` methods return the right child and throws exception if the child is unknown

this would mean we throw an exception if the API answers with a new value in the union, which is something that AWS does not treat as a BC break.
The official AWS SDKs instead of a special member in the union `UNKOWN_TO_SDK` (the actual name differs between languages but they have such concept).

> * Use abstract class that return `null` for all membres

I don't think having all nullable properties is a good idea. It makes the SDK types less precise for dubious benefits. See [my comment](https://github.com/async-aws/aws/issues/1900#issuecomment-3695102533) on the issue showing that your "simple" code justifying this choice is actually buggy code (which won't be detectable by static analysis in this solution, while it could be detected in others)

src/CodeGenerator/src/Definition/UnionShape.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\CodeGenerator\Definition;
+
+/**
+ * @internal
+ */
+class UnionShape extends StructureShape
```
stof: should we actually extend StructureShape ? Many places in the code generator rely on `instanceof` to choose their behavior, and they throw for unsupported shapes. Should union shapes be treated like normal structure shapes for most of those places ? If no, it would be better to have a separate type.

src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -88,6 +89,101 @@ public function generate(StructureShape $shape, bool $forEndpoint = false): Clas
...
+            $serializerBuilderResult = $serializer->generateRequestBuilder($shape, false);
+            foreach ($serializerBuilderResult->getUsedClasses() as $classNameFqdn) {
+                $classBuilder->addUse($classNameFqdn);
+            }
+
+            $method = $classBuilder->addMethod('requestBody')
+                ->setReturnType($serializerBuilderResult->getReturnType())
+                ->setAbstract(true)
+                ->setPublic()
+                ->setComment('@internal');
```
stof, line 157: should the abstract class really be `@internal` ? It will be used in public return types and so will likely need to appear in downstream types as well sometimes.

Instead, I think we should use `@psalm-inheritors` to restrict the allowed subclasses in static analysis. phpstan supports this feature as well (through `@phpstan-sealed`). However, while phpstan is able to read the psalm annotation, [psalm does not support reading that phpstan annotation yet](https://github.com/vimeo/psalm/issues/11510) so we should use the psalm annotation to support both tools.

src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -250,6 +352,37 @@ private function namedConstructor(StructureShape $shape, ClassBuilder $classBuil
...
+            $classBuilder->addUse($memberClassName->getFqdn());
+            $body[] = strtr('if (isset($input[NAME])) {
+                return new CLASS([NAME => $input[NAME]]);
+            }', [
+                'NAME' => var_export($name, true),
+                'CLASS' => $memberClassName->getName(),
+            ]);
+        }
+        $classBuilder->addUse(InvalidArgument::class);
+        $body[] = 'throw new InvalidArgument(\'Invalid union input\');';
```
stof: This should rather return an `UnknownToSdk` type for API results (we can omit generating it if used only in input), with the request body throwing an exception for that type if used in input.

jderusse: right. 

Class added, and instanciation added in `populate` methods (The code called when parsing the AWS response)

This line is called only by userland when users call TheClass::create(), I think we can keep the exception in that case. What do you think?

src/CodeGenerator/src/Generator/ObjectGenerator.php
jderusse: The comment is set on the `requestBody` which is internal (like all generated `requestBody` methods)

The class itself is not internal

src/CodeGenerator/src/Definition/UnionShape.php
jderusse: code refactored

src/CodeGenerator/src/Generator/ObjectGenerator.php
stof: sorry, I missed that.

src/CodeGenerator/src/Generator/ObjectGenerator.php
stof: I agree.

src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -112,16 +133,108 @@ public function generate(StructureShape $shape, bool $forEndpoint = false): Clas
...
+                }
+
+                $method = $classBuilder->addMethod('requestBody')->setReturnType($serializer->getRequestBuilderReturnType())->setBody($serializerBuilderResult->getBody())->setPublic()->setComment('@internal');
+                foreach ($serializer->getRequestBuilderExtraArguments() as $arg => $type) {
+                    $method->addParameter($arg)->setType($type);
+                }
+            }
+        }
+
+        // generate fallback for unkown member
```
stof: When the shape is used as input, the unknown member must implement `requestBody` to satisfy the abstract parent method (it should throw an exception as the unknown value is not valid in input)

stof: and btw, this comment has a typo (`unknown` vs `unkown`)

src/CodeGenerator/src/Generator/CodeGenerator/TypeGenerator.php
```diff
@@ -39,30 +41,114 @@ public function __construct(NamespaceRegistry $namespaceRegistry)
...
+        if ($shape instanceof ObjectShape) {
+            $memberClassNames[] = $className = $this->namespaceRegistry->getObject($shape);
+
+            return [$className->getFqdn(), $className->getName(), $memberClassNames];
+        }
+
+        if ($shape instanceof ListShape) {
+            $listMemberShape = $shape->getMember()->getShape();
+            [$type, $doc, $memberClassNames] = $this->getPhpType($listMemberShape);
+            if ('::*' === substr($doc, -3)) {
```
stof: I suggest using `str_ends_with` instead.

jderusse:
after looking deeper, indeed, having `null` is a fortunate edge case for RDS results because `null` is a possible value of the union.

I've removed it from this PR

src/CodeGenerator/src/Definition/ObjectShape.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace AsyncAws\CodeGenerator\Definition;
+
+/**
+ * @internal
+ */
+class ObjectShape extends Shape
```
stof: I would make it abstract, as we should never instantiate it directly. We should always be using a child class.

src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -112,16 +134,93 @@ public function generate(StructureShape $shape, bool $forEndpoint = false): Clas
...
+        $this->generated[$child->getName()] = $childClassName = $this->namespaceRegistry->getObject($child);
+        $classBuilder = $this->classRegistry->register($childClassName->getFqdn());
+        $classBuilder->setExtends($abstractClassName->getFqdn());
+        $constructor = $classBuilder->addMethod('__construct');
+        $constructor->addParameter('input')->setType('array');
+        $constructor->setBody('$this->unknown = $input;');
+        $property = $classBuilder->addProperty('unknown')->setPrivate();
+        $property->addComment('@var array<string, mixed>');
+        $method = $classBuilder->addMethod('getUnknown')
+            ->setReturnType('array');
```
stof: this should also have the comment `@return array<string, mixed>` to specify the array type.

src/Service/S3Vectors/src/ValueObject/VectorDataMemberUnknownToSdk.php
```diff
@@ -0,0 +1,31 @@
...
+use AsyncAws\Core\Exception\LogicException;
+
+class VectorDataMemberUnknownToSdk extends VectorData
+{
+    /**
+     * @var array<string, mixed>
+     */
+    private $unknown;
+
+    public function __construct(array $input)
```
stof, line 19: I suggest marking the constructor of the unknown member as `@internal` as this class is not meant to be instantiated outside our populate methods.

src/Service/RdsDataService/src/ValueObject/ArrayValue.php
```diff
@@ -2,165 +2,52 @@
 
 namespace AsyncAws\RdsDataService\ValueObject;
 
+use AsyncAws\Core\Exception\InvalidArgument;
+
 /**
  * Contains an array.
  */
-final class ArrayValue
+abstract class ArrayValue
```
stof, line 12: I suggest generating the `@psalm-inheritors` annotation on those abstract class, so that static analysis tools enforce that it is not extended outside the package.

We could also add a `@phpstan-sealed` annotation which is the same annnotation for phpstan. However, phpstan reads the psalm annotation while the opposite is not yet true, so the psalm annotation is more interoperable.

jderusse: done

src/CodeGenerator/src/Generator/ObjectGenerator.php
jderusse: good catch

src/CodeGenerator/src/Generator/ResponseParser/RestJsonParser.php
```diff
@@ -225,6 +229,54 @@ private function parseElement(string $input, Shape $shape, bool $required, bool
...
+                    'PROPERTY_NAME' => var_export($name, true),
+                    'PROPERTY_ACCESSOR' => $this->parseElement('$json', $child, true, true),
+                ]);
+                $className = $this->namespaceRegistry->getObject($child);
+                $this->imports[] = $className;
+            }
+
+            // Generate fallback for unknown member
+            $childForUnknown = $shape->getChildForUnknown();
+            $functionNameForUnknwon = 'populateResult' . ucfirst($childForUnknown->getName());
```
stof:
```suggestion
            $functionNameForUnknown = 'populateResult' . ucfirst($childForUnknown->getName());
```

And updating its usages of course

src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -112,16 +134,106 @@ public function generate(StructureShape $shape, bool $forEndpoint = false): Clas
...
+
+        // generate fallback for unknown member
+        $child = $shape->getChildForUnknown();
+        $this->generated[$child->getName()] = $childClassName = $this->namespaceRegistry->getObject($child);
+        $classBuilder = $this->classRegistry->register($childClassName->getFqdn());
+        $classBuilder->setFinal()->setExtends($abstractClassName->getFqdn());
+        $inheritors[] = $childClassName->getName();
+
+        $classBuilder->addMethod('__construct')
+            ->setComment('@internal')
```
stof: should also add the `@param array<string, mixed> $input` in the comment.

src/CodeGenerator/src/Generator/ObjectGenerator.php
```diff
@@ -112,16 +134,106 @@ public function generate(StructureShape $shape, bool $forEndpoint = false): Clas
...
+                }
+
+                $method = $classBuilder->addMethod('requestBody')->setReturnType($serializer->getRequestBuilderReturnType())->setBody($serializerBuilderResult->getBody())->setPublic()->setComment('@internal');
+                foreach ($serializer->getRequestBuilderExtraArguments() as $arg => $type) {
+                    $method->addParameter($arg)->setType($type);
+                }
+            }
+        }
+
+        // generate fallback for unknown member
```
stof: should we generate the fallback member only for shapes used as result ? That's what we did for the `UNKNOWN_TO_SDK` case in enums.

jderusse: good catch

src/CodeGenerator/src/Generator/ShapedUsedInContextTrait.php
```diff
@@ -0,0 +1,122 @@
...
+use AsyncAws\CodeGenerator\Definition\UnionShape;
+
+/**
+ * Provides method to test if a shaped is used in input or output.
+ *
+ * @author Jérémy Derussé <jeremy@derusse.com>
+ *
+ * @internal
+ */
+trait ShapedUsedInContextTrait
```
stof: Instead of a trait, maybe this should actually be a dedicated class injected in the various generators needing it. This would allow the cache to be shared instead of having separate caches in each generator using the trait.

src/CodeGenerator/src/Generator/ShapedUsedInContextTrait.php
```diff
@@ -0,0 +1,122 @@
...
+                if (null === $shape) {
+                    return;
+                }
+                if (isset($this->usedShapedOutput[$shape->getName()])) {
+                    // Node already visited
+                    return;
+                }
+
+                $this->usedShapedOutput[$shape->getName()] = true;
+                if ($shape instanceof StructureShape) {
```
stof: this misses walking union shapes.

stof: this is the reason why `ArrayValue` is missing its unknown member in the generated code (detected by phpstan and psalm) as it is used only inside the `Field` union shape.

jderusse: refactored to use the same "walker" method/logic.

src/CodeGenerator/src/Generator/ShapeUsageHelper.php
```diff
@@ -0,0 +1,113 @@
...
+     * @var array<string, bool>|null
+     */
+    private $usedShapedInput;
+
+    /**
+     * @var array<string, bool>|null
+     */
+    private $usedShapedOutput;
+
+    public function __construct(array $managedMethods)
```
stof: the named `managedMethods` is weird. ServiceGenerator calls that `managedOperations`. Maybe we should use the same.

src/CodeGenerator/src/Generator/ShapeUsageHelper.php
```diff
@@ -0,0 +1,113 @@
...
+                        $this->walkShape($error, $this->usedShapedOutput);
+                    }
+                }
+            }
+        }
+
+        return $this->usedShapedOutput[$shape->getName()] ?? false;
+    }
+
+    private function walkShape(?Shape $shape, array &$marker): void
```
stof, line 90: you should add the phpdoc for `@param array<string, bool> $marker`

PR #2020, opened 2025-12-26
Fix incorrect version for first version of BedrockAgent in CHANGELOG

jderusse:


stof:
Why incorrect ? We could release it as 1.0.0

PR #2044, opened 2026-02-27
Handle Amazon header formatting so pre-signed S3 PUT URLs with metadata result in valid signatures

ryanmab:
## Description

Closes https://github.com/async-aws/aws/issues/2043.

Pre-signed S3 PUT requests (with user-defined metadata) are rejected by S3 because the signature is not valid.

This was caused by [a change in the formatting of the user-defined metadata keys](https://github.com/async-aws/aws/pull/2037/changes#diff-e4d706be2f18732329a2930eeacfde8390a0dc965baedf636b4e8a1b9bcc8d24R247).

This fixes the issue by preserving the casing of S3 user-defined metadata - I've validated on my side that S3 PUT requests are now accepted.

### Can we apply the case change to `X-Amz-Meta-` still, if the signature _also_ includes the case change?

I tried uniformly applying `formatAmazonHeader` to all `x-amz` values in the `buildCanonicalRequest` too (so everything formatted as `X-Amz-<something>` and `X-Amz-Meta-<something>`), thinking it was just because the canonical request was still in lowercase and that effected the signature. However that didn't appear to make a difference - the signatures were still rejected by S3.

I _haven't_ looked into the official AWS SDK, but [my assumption from the docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata) is its an expectation that the user-defined metadata is all lowercase.

### DIFF
--- a/src/Core/src/Signer/SignerV4.php
+++ b/src/Core/src/Signer/SignerV4.php
@@ -244,7 +244,8 @@ class SignerV4 implements Signer
     {
         foreach ($request->getHeaders() as $name => $value) {
             if ('x-amz' === substr($name, 0, 5)) {
-                $attribute = implode('-', array_map(ucfirst(...), explode('-', $name)));
+                $attribute = self::formatAmazonHeader($name);
+
                 $request->setQueryAttribute($attribute, $value);
             }
 
@@ -309,7 +310,10 @@ class SignerV4 implements Signer
             $this->buildCanonicalQuery($request),
             implode("\n", array_values($canonicalHeaders)),
             '', // empty line after headers
-            implode(';', array_keys($canonicalHeaders)),
+            implode(
+                ';',
+                array_keys($canonicalHeaders)
+            ),
             $bodyDigest,
         ]);
     }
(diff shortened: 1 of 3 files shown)


src/Core/src/Signer/SignerV4.php
```diff
@@ -371,4 +375,26 @@ private function buildSignature(string $stringToSign, string $signingKey): strin
     {
         return hash_hmac('sha256', $stringToSign, $signingKey);
     }
+
+    private function formatAmazonHeader(string $key): string
+    {
+        if (
+            'x-amz' === substr($key, 0, 5)
+            && 'x-amz-meta' !== substr($key, 0, 10)
```
ryanmab, line 383: This is the change.

src/Core/src/Signer/SignerV4.php
```diff
@@ -371,4 +375,26 @@ private function buildSignature(string $stringToSign, string $signingKey): strin
     {
         return hash_hmac('sha256', $stringToSign, $signingKey);
     }
+
+    private function formatAmazonHeader(string $key): string
```
GrahamCampbell: This method can be made static?

ryanmab: Yeah no reason why not - have made this change!

jderusse:
superseeded by #2057

PR #2053, opened 2026-03-19
Add ListEventSourceMappings, PutFunctionConcurrency to Lambda

shadowhand:


shadowhand:
When trying to generate the LambaClient::ListEventSourceMappings operation, I get the following error:

```
No camel case property "UUID" is not yet implemented
```

But I'm unsure what this means...?

jderusse:
> When trying to generate the LambaClient::ListEventSourceMappings operation, I get the following error:
> 
> ```
> No camel case property "UUID" is not yet implemented
> ```
> 
> But I'm unsure what this means...?

Add a new entry in the [GeneratorHelper](https://github.com/async-aws/aws/blob/master/src/CodeGenerator/src/Generator/GeneratorHelper.php#L38)
like `'UUID' => 'Uuid'`

We have code that normalize names in camelCase, but fails when the input  contains multiple consecutive upper cases. So we maintain a table of replacement/ignore for those cases.

jderusse:
thank you @shadowhand

shadowhand:
@jderusse thanks for finishing this up!

PR #2055, opened 2026-03-24
Fix: filter not assigned from input in QueryVectorsInput constructor

agavss:
The filter field is declared as a property but never assigned in __construct(), so any filter passed to queryVectors() is silently ignored and never included in the HTTP request body.


agavss:
What can I do to fix the PR?

jderusse:
closed in favor of #2056 

Thank you for the hint @agavss

PR #2057, opened 2026-03-27
Fix PresignedUrl issue with case convertion of header names

jderusse:
an alternative to #2044
that should fix #2043 without breaking #2035

Could you test this patch @root-aza and @ryanmab ?

### DIFF
--- a/src/Core/CHANGELOG.md
+++ b/src/Core/CHANGELOG.md
@@ -8,6 +8,10 @@
 - AwsClientFactory: add `ec2()` method for the new `async-aws/ec2` client
 - XmlAwsErrorFactory: parse EC2-style error envelopes (`<Response><Errors><Error>...`)
 
+### Fixed
+
+- SignerV4: Fix presign request for S3 when request contains user-defined metadata
+
 ## 1.28.1
 
 ### Changed
--- a/src/Core/src/AbstractApi.php
+++ b/src/Core/src/AbstractApi.php
@@ -152,7 +152,7 @@ abstract class AbstractApi
 
         $length = $request->getBody()->length();
         if (null !== $length && !$request->hasHeader('content-length')) {
-            $request->setHeader('content-length', (string) $length);
+            $request->setHeader('Content-Length', (string) $length);
         }
 
         // Some servers (like testing Docker Images) does not support `Transfer-Encoding: chunked` requests.
--- a/src/Core/src/Request.php
+++ b/src/Core/src/Request.php
@@ -15,6 +15,8 @@ use AsyncAws\Core\Stream\RequestStream;
 final class Request
 {
     use BuildHttpQueryTrait;
+    private const UPPER = '_ABCDEFGHIJKLMNOPQRSTUVWXYZ';
+    private const LOWER = '-abcdefghijklmnopqrstuvwxyz';
 
     /**
      * @var string
@@ -30,6 +32,11 @@ final class Request
      * @var array<string, string>
      */
     private $headers;
+
+    /**
+     * @var array<string, string> Map of lowercase header name => original name at registration
+     */
+    private $headerNames = [];
 
     /**
      * @var RequestStream
@@ -70,8 +77,10 @@ final class Request
         $this->method = $method;
         $this->uri = $uri;
         $this->headers = [];
-        foreach ($headers as $key => $value) {
-            $this->headers[strtolower($key)] = (string) $value;
+        foreach ($headers as $name => $value) {
+            $normalized = strtr($name, self::UPPER, self::LOWER);
+            $this->headerNames[$normalized] = $name;
+            $this->headers[$name] = (string) $value;
         }
         $this->body = $body;
         $this->query = $query;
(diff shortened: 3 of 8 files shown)


ryanmab:
Hey @jderusse,

Thanks for the work on this PR!

I've just pulled this version in locally on my side, but no change I'm afraid - the signed PUT requests are still rejected by S3 with the error response of:

```
<Error>
	<Code>SignatureDoesNotMatch</Code>
	<Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
        // .....
</Error>
```

For what it's worth, I wanted to rule out that I'd not tested the patch correctly, so also tried pinning `async-aws/core` to `1.28.0` and then applying the changes in the `core` and `s3` packages manually file by file.

Before applying the changes, the signed PUT requests worked against S3 (i.e. just plain `1.28.0` of `async-aws/core`), but after applying this patch to each file manually, the signed URLs to began failing - so I believe I'm exercising the Signer correctly.

I realise thats not a while lot to go on though, so let me know if theres anything outside of the original issue or my follow-up PR which might help you track it down.

Ryan

jderusse:
Could you provide a reproducer? 

I used the following code, and both have the same signature (only querystring parameter order changes)

```
$s3Async = new AsyncS3([
    'pathStyleEndpoint' => true,
]);
$input = new PutObjectRequest([
    'Bucket' => 'xxxx',
    'Key' => 'a.txt',
    'Metadata' => [
        'BAZ' => 'qux',
        'foo' => 'bar',
    ],
]);
$urlAsync = $s3Async->presign($input, new \DateTimeImmutable('+60 min'));


$client = new SyncS3Client(
    [
        'bucket'                  => 'xxxx',
        'use_path_style_endpoint' => true,
    ]
);
$command = $client->getCommand('PutObject', [
    'Bucket' => 'xxx',
    'Key' => 'a.txt',
    'Metadata' => [
        'BAZ' => 'qux',
        'foo' => 'bar',
    ],
]);
$request = $client->createPresignedRequest($command, new \DateTimeImmutable('+60 min'));
$urlSdk = (string) $request->getUri();


var_dump($urlAsync);
var_dump($urlSdk);
```

ryanmab:
Hey @jderusse,

Ah, so, there's a curveball!

I put together a reproduction using this branch of `async-aws` and the official SDK (below), and indeed you're spot on, your patch _does_ produce the same signature as the SDK. But, I took it one step further, and used a live S3 bucket and cURL (also below) to try both pre-signed requests, and it turns out that _neither_ SDK produced a valid signature...!

Digging a bit deeper, I tried a few variants out, and the reason for the invalid signature from both SDKs is the `BAZ` metadata key (line annotated below) - remove that metadata, or change the key to be lowercase, and both requests begin working.

So, I guess this raises an interesting, and I imagine unexpected (?), realisation - the original implementation of the `async-aws` Signer, which mutated the metadata keys case, appeared to inadvertently handle additional cases which the official SDK does not (keys with uppercase characters).

In terms of next steps, I'm totally happy to follow your lead, but I guess theres two ways forward:

1. We tweak the new Signer implementation to handle S3 metadata keys that are uppercase, by mutating only them to lowercase before signing (similar to the PR I raised), so that AWS will accept the resulting presigned URL.
   - It's closest to the original `<= v1.28.0` behaviour, and I suppose a better DX for developers when it comes to not having to worry about the casing of their metadata.
3. We merge this PR in as is, because it has behavioural parity with the official SDK, and accept that presigned URLs with metadata keys that contain uppercase characters will fail.
   - Perhaps a note [in the docs](https://async-aws.com/features/presign.html) pointing people in the direction of lowercasing all their metadata keys?

Of course, I'd prefer option 1, since its closer to the (I suppose arguably wrong!) `<= v1.28.0` behaviour, but I totally understand why maintaining behavioural parity with the official SDK would be preferable, and can work around that in my use case if thats where you land.

Thanks,
Ryan

PHP:
```
<?php
require("vendor/autoload.php");

$config = \AsyncAws\Core\Configuration::create([
    'region' => 'eu-west-2',
    'pathStyleEndpoint' => true,
    'accessKeyId' => '<key>',
    'accessKeySecret' => '<secret>',
    'sessionToken' => '<token>'
]);
$s3Async = new \AsyncAws\S3\S3Client($config);
$input = new \AsyncAws\S3\Input\PutObjectRequest([
    'Bucket' => '',
    'Key' => 'a.txt',
    'Metadata' => [
        'BAZ' => 'qux', // Remove, or lower case, this metadata key and the request will start working
        'foo' => 'bar',
    ],
]);
$urlAsync = $s3Async->presign($input, new \DateTimeImmutable('+60 min'));

$s3Client = new \Aws\S3\S3Client([
    'region' => 'eu-west-2',
    'use_path_style_endpoint' => true,
    'credentials' => new \Aws\Credentials\Credentials(
        '<key>',
        '<secret>',
        '<token>'
    )
]);

$request = $s3Client->createPresignedRequest(
    $s3Client->getCommand('PutObject', [
        'Bucket' => '',
        'Key' => 'a.txt',
        'Metadata' => [
            'BAZ' => 'qux', // Remove, or lower case, this metadata key and the request will start working
            'foo' => 'bar',
        ],
    ]),
    '+1 hour'
);

var_dump($urlAsync);
var_dump((string) $request->getUri());

```

cURL:
```
curl --request PUT \
        --url '' \
        --header 'Content-Type: text/plain' \
        --data '{}'
```

jderusse:
Thank you for the deep investigation and complet report.

Indeed, the official documentation says that Meta Key have to be lower case https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html

This PR provides a implementation that is similar to the SDK (header case is respected). So I prefer this generic solution, than using a special if branch when key matches `x-amz-meta`.

Thank you a lot for you help and patience.

PR #2061, opened 2026-04-09
RawUrlEncode CopySource on CopyObjectRequest

### LINKED ISSUE #2060: `CopyObjectRequest`: `CopySource` parameter should be `rawurlencode`d
The `CopySource` parameter in `CopyObjectRequest` is currently passed as a raw string to the `x-amz-copy-source` header. However, according to Amazon S3 documentation, the source object name must be URL-encoded if it contains special characters.

In `vendor/async-aws/s3/src/Input/CopyObjectRequest.php`, the `request()` method sets the header directly:

```php
if (null === $v = $this->copySource) {
    throw new InvalidArgument(\sprintf('Missing parameter "CopySource" for "%s". The value cannot be null.', __CLASS__));
}
$headers['x-amz-copy-source'] = $v;
```

It should be encoded similarly to how the `Key` parameter is handled for the URI:

```php
$headers['x-amz-copy-source'] = str_replace('%2F', '/', rawurlencode($v));
```

This ensures that object keys with spaces or other special characters are correctly processed by S3 during the copy operation.

The error :
```
AsyncAws\\Core\\Exception\\Http\\ClientException(code: 400): HTTP 400 returned for \"https://s3.eu-west-3.amazonaws.com/EDITED-OPERATION%20EDITED%2020%25-111122-0.png\".\n\nCode:    InvalidArgument\nMessage: Invalid copy source encoding\nType:    \nDetail:  \n at /var/task/vendor/async-aws/core/src/Response.php:427
```

Cafeine42:
fix #2060

### DIFF
--- a/src/Service/S3/CHANGELOG.md
+++ b/src/Service/S3/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Fixed
+
+- Fix URL encoding of `CopySource` in `CopyObjectRequest` (fix #2060)
+
 ### Added
 
 - Added `S3Client::getBucketLifecycleConfiguration()` and `S3Client::putBucketLifecycleConfiguration()` methods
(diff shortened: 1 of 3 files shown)


jderusse:
Nothing in ther AWS SDK specs says that this specific header have to be encoded.
Also the official SDK does not encoded this header neither.

PR #2064, opened 2026-04-16
Implement EC2 client

GrahamCampbell:


src/CodeGenerator/src/Generator/RequestSerializer/Ec2QuerySerializer.php
```diff
@@ -0,0 +1,59 @@
...
+ *    locationName is ignored, and the flattened/non-flattened distinction is
+ *    irrelevant (every list serializes positionally).
+ *  - Structure-member names prefer `queryName` verbatim, else `ucfirst(locationName)`,
+ *    else the PHP key (already PascalCase in AWS EC2 models).
+ *
+ * Mirrors aws-sdk-php's `Aws\Api\Serializer\Ec2ParamBuilder`.
+ *
+ * @internal
+ */
+class Ec2QuerySerializer extends QuerySerializer
```
jderusse, line 26: It looks like "queryName" attribute is only used by Ec2. 

Maybe we can make this class the default implementation for `QuerySerializer`?

jderusse: hmm... ok, it looks like the official SDK have a special implementation for this...

jderusse:
thank you @GrahamCampbell

PR #2072, opened 2026-04-17
fix: exclude .gitattributes from exports in Core and Ses

Root-Aamir:
"Hello! As suggested in issue #2046, I have added the export-ignore attribute for .gitattributes files in both the Core and Ses packages. This ensures that the file itself is excluded when the library is distributed or archived.

Closes #2046"


src/Core/.gitattributes
```diff
@@ -3,3 +3,4 @@
 /.gitignore export-ignore
 /Makefile export-ignore
 /phpunit.xml.dist export-ignore
+/.gitattributes export-ignore
```
jderusse: any reason for updating only the files in `Core` and `Ses`?
What's about the other sub-splits?

~You'll also need to add an entry in the CHANGELOG of each entry.
IIRC we do have a `add_change` command that'll automate this part~

Root-Aamir:
"I have updated the .gitattributes files in the major sub-splits (Core, S3, Lambda, Kms, Iam) and ensured each file ends with a proper newline to fix the previous warning. I've also manually added the changelog entries for these packages since the automated script had some environment issues. Ready for review now! Thanks."

jderusse:
I added missing entries. 
Thank you @Root-Aamir

Root-Aamir:
"Thank you @jderusse for the help and for merging this! Glad I could contribute to the project."

PR #2073, opened 2026-04-17
[SimpleS3] Use PartSize as threshold for multipart upload

### LINKED ISSUE #2063: SimpleS3: Upload chunks too large for minIO
In our project, we're using minIO as a compatible S3 replacement. I'd like to use Flysystem and AsyncAWS to write to that minIO bucket. However, if I attempt to upload a file between 16 and 64 MiB, the upload fails with a HTTP 400 and minIO reports:

> chunk too big: choose chunk size <= 16MiB

While Amazon's S3 allows rather large files to be sent with a single request, minIO seems to be more strict by design, see minio/minio#21303.

Looking into the implementation of the SimpleS3 client, I found that I can control the chunk size through configuration (`PartSize`), but the threshold up to which a simple PUT is used instead of a multipart upload is hardcoded to 64 MiB.

https://github.com/async-aws/aws/blob/41a14437d4512f1c7f0f11c87aebc80305428971/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php#L91-L93

I would like to work on a PR that fixes the issue for us. The straightforward solution would be to introduce a way to configure the threshold like we can already configure the `PartSize`.

Another idea would be to couple the threshold to the `PartSize`: We only do a single-request PUT to avoid a multipart upload with only one part. As `PartSize` currently defaults to 64 MiB, we would not change anything for downstream code that sticks to the defaults.

What do you think?

derrabus:
Fixes #2063.

Currently, `SimpleS3Client` uses a simple `PUT` instead of a multipart upload if the file size is below 64 MiB. This is fine for Amazon's S3, but as I laid out in #2063, 3rd party implementations of the S3 protocol such as minIO might no accept that.

This PR couples the formerly hardcoded theshold to the `PartSize` setting. The consquence is that `SimpleS3Client` uses a `PUT` in cases where the multipart upload would consist of only a single part. Given that the `PartSize` defaults to 64 MiB, the default behavior remains almost unchanged.

Projects that use minIO should now be able to upload by setting `PartSize` to `16` or a lower value.

### DIFF
--- a/src/Integration/Aws/SimpleS3/CHANGELOG.md
+++ b/src/Integration/Aws/SimpleS3/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 - Remove redundant ext-json requirement
+- Use PartSize as threshold for multipart upload
 
 ## 3.0.0
 
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -89,7 +89,7 @@ class SimpleS3Client extends S3Client
         unset($options['PartSize']);
 
         // If file is less than 64MB, use normal upload
-        if (null !== $contentLength && $contentLength < 64 * $megabyte) {
+        if (null !== $contentLength && $contentLength <= $partSize * $megabyte) {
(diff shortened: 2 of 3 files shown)


jderusse:
thank you @derrabus

PR #2082, opened 2026-05-03
feat: add non-buffered response streaming

GrahamCampbell:
Allow streaming output operations to opt out of response buffering so large downloads can be consumed without temp storage while keeping buffered behavior as the default. My use case is streaming a large CSV file out of S3. Without this change, I have to go around the SDK and create a presigned URL and download it myself using another HTTP client. Without the ability to turn off buffering, the entire file gets written to `/tmp`.

### DIFF
--- a/docs/clients/s3.md
+++ b/docs/clients/s3.md
@@ -73,8 +73,10 @@ request which could have a performance impact.
 ### Download files
 
 When you download a file from S3, AsyncAws gives you a `ResultStream` which
-can be used as a string, as a resource, or iterated over. This allows you to handle
-larger files without having them in memory.
+can be used as a string, as a resource, or iterated over. By default, response
+bodies are buffered into temporary storage so the stream can be replayed and
+safely reused by another request. This avoids loading large objects into PHP
+memory, but large objects can still use local temporary storage.
 
 ```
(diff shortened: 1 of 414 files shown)


GrahamCampbell:
Hmmm, tough to test due to the cyclic nature of needing a newer version of core...

jderusse:
I would avoid creating precedence by adding a new custom `@responseBuffer` parameter.

I didn't yet read the implementation, but some replacement suggestion in mind

1. Is there a reason for buffering the output? Couldn't we alway stream without buffer? (rewind come in my mind, but do we have something else? could we let end-user buffer the output if they want to rewind. And release a new Major version that don't buffer)
2. or buffer when the response is small, and that size is configurable with something like `new S3Client(['doNotBufferWhenSizeExceed' => '3Gib'])` 
3. Or maybe more simple : can't we add a new `response->getContentAsResource(buffer: false)` to achieve the same?

GrahamCampbell:
Yes, we do want to buffer the output in general by default (and by default, small responses buffer entirely in memory without spilling to disk anyway). Without buffering it, we can't read it more than once, and there are many situations where we would want to do that.

PR #2087, opened 2026-06-01
Generate code for SSO and SSO-OIDC

Nyholm:
Added some more endpoints to be able to authenticate an app with SSO and then list accounts and roles.

### DIFF
--- a/src/Integration/Symfony/Bundle/CHANGELOG.md
+++ b/src/Integration/Symfony/Bundle/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Fixed
+
+- Fixed tests on Symfony 8.1
+
 ## 1.17.0
 
 ### Added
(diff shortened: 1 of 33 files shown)


src/Integration/Symfony/Bundle/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## NOT RELEASED
 
+### Fixed
+
+- Fixed tests on Symfony 8.1
```
jderusse, line 7: What's about creating a dedicated PR for this change?

jderusse (approved):

src/Integration/Symfony/Bundle/CHANGELOG.md
Nyholm: I considered it, but it was just to update a dev-dependency. So I decided it was a "fix CI" type of change.

PR #2096, opened 2026-06-08
Fixed deprecation warnings on Symfony 8.1

jderusse:


src/Integration/Symfony/Bundle/src/AsyncAwsBundle.php
```diff
@@ -4,16 +4,46 @@
...
 
+use AsyncAws\Symfony\Bundle\DependencyInjection\AsyncAwsExtension;
 use AsyncAws\Symfony\Bundle\DependencyInjection\Compiler\InjectCasterPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
+use Symfony\Component\DependencyInjection\Kernel\AbstractBundle;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
-class AsyncAwsBundle extends Bundle
+if (class_exists(AbstractBundle::class)) {
```
jderusse, line 14: I'm a bit rusty with BC layer. Is this the right way to do it?

stof:
Which deprecation is this solving ? AFAIK, we have not deprecated the HttpKernel bundle class (it would be far too disruptive for the ecosystem of bundles to do it at the same time than adding the new API)

jderusse:
> Which deprecation is this solving ? AFAIK, we have not deprecated the HttpKernel bundle class (it would be far too disruptive for the ecosystem of bundles to do it at the same time than adding the new API)

the Symfony\Component\HttpKernel\Bundle\BundleInterface interface is triggering the CI
https://github.com/async-aws/aws/actions/runs/27763869537/job/82145149935?pr=2092#step:7:9

stof:
This is an issue in Psalm IMO. AsyncAwsBundle does not *directly* implement the deprecated interface. It inherits it from its non-deprecated parent class (which keeps implementing the interface for BC reasons).

Symfony does *not* trigger a deprecation warning on such usage.

I suggest reporting the issue to Psalm, and putting this in the psalm baseline in the meantime.

PR #2103, opened 2026-07-01
[SimpleS3] Fix changelog

ngrie:
While we are patiently waiting for a release that includes #2073 I realized that the changelog entry of that pull request made it into the wrong section, making it look like there aren't any unreleased changes currently 🙂


jderusse:
good catch, Thanks

PR #2115, opened 2026-07-31
Add conditional multipart uploads to SimpleS3

chrisjenkinson:
`SimpleS3Client::upload()` already forwards `IfMatch` and `IfNoneMatch` for small objects written with `PutObject`, but these conditions are lost when the upload switches to multipart. This can cause a caller expecting a conditional write to overwrite an object unconditionally.

This change keeps conditional options off multipart initiation and part requests, then applies them to `CompleteMultipartUpload`, where S3 evaluates the write condition. It also preserves conditions when an unknown-length stream falls back to `PutObject` and ignores null-valued conditional options.

### DIFF
--- a/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
+++ b/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
@@ -98,11 +98,13 @@ class SimpleS3Client extends S3Client
         }
 
         $completeMultipartUploadOptions = [];
-        foreach (['IfMatch', 'IfNoneMatch'] as $option) {
-            if (\array_key_exists($option, $options)) {
-                $completeMultipartUploadOptions[$option] = $options[$option];
-                unset($options[$option]);
-            }
+        if (\array_key_exists('IfMatch', $options)) {
+            $completeMultipartUploadOptions['IfMatch'] = $options['IfMatch'];
+            unset($options['IfMatch']);
+        }
+        if (\array_key_exists('IfNoneMatch', $options)) {
+            $completeMultipartUploadOptions['IfNoneMatch'] = $options['IfNoneMatch'];
+            unset($options['IfNoneMatch']);
         }
 
         $parts = [];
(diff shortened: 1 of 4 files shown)


src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -147,17 +159,31 @@ public function upload(string $bucket, string $key, $object, array $options = []
...
+        try {
+            $this->completeMultipartUpload(array_merge($completeMultipartUploadOptions, [
+                'Bucket' => $bucket,
+                'Key' => $key,
+                'UploadId' => $uploadId,
+                'MultipartUpload' => new CompletedMultipartUpload(['Parts' => $parts]),
+            ]));
+        } catch (\Throwable $e) {
+            try {
+                $this->abortMultipartUpload([
```
jderusse: This is not related to you PR.
Also I'm not sure it's a good idea. It prevent retrying the final completion part (in case of 500, 503, 429 error)

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
```diff
@@ -95,6 +97,16 @@ public function upload(string $bucket, string $key, $object, array $options = []
             return;
         }
 
+        $completeMultipartUploadOptions = [];
+        if (\array_key_exists('IfMatch', $options)) {
+            $completeMultipartUploadOptions['IfMatch'] = $options['IfMatch'];
+            unset($options['IfMatch']);
+        }
+        if (\array_key_exists('IfNoneMatch', $options)) {
```
jderusse: `isset` instead of `array_key_exists` to handle null cases

jderusse (commented):
Fix llok good, I would just remove the thing about aborting the requests

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
chrisjenkinson: Removed the completion-failure abort. The original completion exception now propagates without destroying the upload ID and parts, preserving retryability for transient errors.

src/Integration/Aws/SimpleS3/src/SimpleS3Client.php
chrisjenkinson: Changed extraction to isset(). Null values are omitted from completion and unconditionally removed before CreateMultipartUpload; coverage now checks both conditional headers.

jderusse:
thank you @chrisjenkinson

PR #2124, opened 2026-08-26
Add BedrockAgentCore client with InvokeAgentRuntime

### LINKED ISSUE #2084: [New Service] BedrockAgentCore
BedrockAgentCore is AWS's managed runtime for invoking AI agents built on Amazon Bedrock.
It allows calling agent runtimes directly via a streaming HTTP response, with session management and optional endpoint qualifiers. 
It is the successor/complement to BedrockAgentRuntime for the new AgentCore product line.

Documentation: [https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_InvokeAgentRuntime.html](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_InvokeAgentRuntime.html)

There is one method to implement:

- [ ] InvokeAgentRuntime

Generic use-cases:
* Invoke an AI agent runtime for conversational applications https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-invoke-agent.html
* Multi-agent orchestration pipelines
* Streaming agent responses with session continuity (runtimeSessionId)

Amoifr:
Closes #2084.

Amazon Bedrock AgentCore is the managed runtime for agents built on Bedrock. `InvokeAgentRuntime` sends a payload to an agent or tool hosted there and returns its response, keeping several invocations in the same conversation through the runtime session id.

Built by the book from `docs/contribute/generate.md`: service copied from `src/Service/.template`, registered in `manifest.json` against the `bedrock-agentcore/2024-02-28` definition, operation generated, then the stubs filled in. Also wired into `AwsClientFactory::bedrockAgentCore()`, `AwsPackagesProvider`, the root `README.md` overview, `docs/clients/bedrock-agent-core.md`, `docs/clients/index.md` and `couscous.yml`.

Only `InvokeAgentRuntime` is generated, as the issue asks. The service has sixty-odd other operations left in `"methods"` for whoever needs them.

## Naming

I went with `BedrockAgentCore` / `async-aws/bedrock-agent-core`, following the `Foobar` → `DynamoDb` / `foobar` → `dynamo-db` rule from the contribute page. AWS itself spells the endpoint prefix `bedrock-agentcore` (no second dash), so say the word if you would rather have `async-aws/bedrock-agentcore` and I will rename.

## Tests

`InvokeAgentRuntimeRequestTest` covers both the plain invocation, with the ARN URL-encoded into the path and the `qualifier` query parameter, and the MCP header set. `InvokeAgentRuntimeResponseTest` covers a full response and one where every optional header is absent, which is the realistic streaming case (`text/event-stream`). Integration tests are skipped like BedrockAgent's, since there is no docker image for the service.

Green locally: the service suite, PHPStan, Psalm and PHP-CS-Fixer. The only failures in `src/Core` and the Symfony bundle are two pre-existing `APCu is not enabled` errors from my machine.

Thanks for async-aws, and for pointing at the generator on the issue. 🙏


jderusse:
thank you @Amoifr

